PR #161, opened 2021-11-19
[MLC-168] added support to add text to the check suite;

yromanyshyn:


matanper:
Decided to not do that

PR #195, opened 2021-11-26
[MLC-200] added ClassPerformanceImbalanceCheck;

yromanyshyn:


deepchecks/checks/performance/class_performance_imbalance.py
```diff
@@ -0,0 +1,236 @@
...
+                raise DeepchecksValueError(
+                    f'Check {check_name} expecting that alternative metrics will return '
+                    f"not empty instance of 'Dict[Hasable, float|int]', but got Dict[Hashable, {value_type}]" #pylint: disable=inconsistent-quotes
+                )
+
+            result[metric_name] = metric_result
+
+        return result
+
+    def add_condition_percentage_difference_not_greater_than(self: CP, threshold: float = 0.3) -> CP:
```
noamzbr, line 185: Think it's better that the default condition be on ratio of change / difference rather than absolute value. Not saying that having this one in addition is bad.

deepchecks/metric_utils.py
```diff
@@ -64,12 +64,11 @@ def task_type_check(model: Union[ClassifierMixin, RegressorMixin], dataset: 'Dat
     model_type_validation(model)
     dataset.validate_label(task_type_check.__name__)
 
-    if getattr(model, 'predict_proba', None):
+    if hasattr(model, 'predict_proba'):
         unique_labels = dataset.label_col().unique()
         if sorted(unique_labels) != list(range(len(unique_labels))):
             raise DeepchecksValueError(f'Classification labels must be a consecutive set from 0 to MAX_LABEL,'
                                        f' found {sorted(unique_labels)}.')
-        model: ClassifierMixin
```
noamzbr, line 72: this used to serve to help my type hint in the IDE

deepchecks/checks/performance/class_performance_imbalance.py
```diff
@@ -0,0 +1,236 @@
...
+
+        if self.alternative_metrics is not None:
+            df = pd.DataFrame.from_dict(self._execute_alternative_metrics(y_true, y_pred))
+
+        else:
+            labels = unique_labels(y_true, y_pred)
+            df = pd.DataFrame.from_dict({
+                name: dict(zip(
+                    labels,
+                    t.cast(t.Iterable, metric_func(y_true, y_pred, labels=labels))
```
noamzbr, line 115: In order to function similarly to TrainTestDifferenceOverfit, better to have the user supplied metric be either scorer strings or sklearn scorers. Look how it's done there (and feel free to offer a different opinion)

noamzbr (commented):
Apart from comments, just reminding to also add a notebook and add the check and it's default condition to the relevant suite (performance)

deepchecks/metric_utils.py
yromanyshyn: hm/
1. how is it helping if `model` var is not used below that annotation?
2. I am using VSCode with the pyright static type checker, and that is what he is saying if that annotation is present - `Parameter declaration "model" is obscured by a declaration of the same name`
3. for cases like this it is better to use `typing.cast` (or `assert`), mypy docs about it - [link](https://mypy.readthedocs.io/en/stable/casts.html), python docs about it - [link](https://docs.python.org/3/library/typing.html#typing.cast)

deepchecks/checks/performance/class_performance_imbalance.py
noamzbr: My bad! It is already percent change...

deepchecks/checks/performance/class_performance_imbalance.py
```diff
@@ -0,0 +1,236 @@
...
+        check_name = type(self).__name__
+        expected_model_types = [ModelType.BINARY, ModelType.MULTICLASS]
+
+        dataset.validate_label(check_name)
+        Dataset.validate_dataset(dataset, check_name)
+        task_type_validation(model, dataset, expected_model_types, check_name)
+
+        features = dataset.features_columns()
+
+        if features is None:
```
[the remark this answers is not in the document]
yromanyshyn: updated

deepchecks/checks/performance/class_performance_imbalance.py
```diff
@@ -0,0 +1,236 @@
...
+        """
+        super().__init__()
+        self.alternative_metrics = metrics
+
+        if self.alternative_metrics is not None:
+
+            if len(self.alternative_metrics) == 0:
+                raise ValueError('Expected to receive not empty dict of callables!')
+
+            incorrect_argument = next(
```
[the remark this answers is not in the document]
yromanyshyn: updated

deepchecks/checks/performance/class_performance_imbalance.py
yromanyshyn: updated

deepchecks/checks/performance/class_performance_imbalance.py
```diff
@@ -0,0 +1,236 @@
...
+
+
+class ClassPerformanceImbalanceCheck(SingleDatasetBaseCheck):
+    """Visualize class imbalance by displaying the difference between class metrics."""
+
+    def __init__(
+        self,
+        metrics: t.Optional[t.Mapping[str, MetricFunc]] = None
+    ):
+        """Initialize ClassPerformanceImbalanceCheck check.
```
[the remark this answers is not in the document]
yromanyshyn: updated

deepchecks/checks/performance/class_performance_imbalance.py
```diff
@@ -0,0 +1,239 @@
...
+from deepchecks.string_utils import format_percent
+from deepchecks.utils import DeepchecksValueError
+
+
+__all__ = ['ClassPerformanceImbalanceCheck']
+
+
+MetricFunc = t.Callable[
+    [object, pd.DataFrame, pd.Series], # model, features, labels
+    t.Dict[t.Hashable, t.Union[float, int]] # score for each label
```
matanper, line 22: doesn't metric function returns a single number?

deepchecks/checks/performance/class_performance_imbalance.py
```diff
@@ -0,0 +1,239 @@
...
+
+    def __init__(
+        self,
+        alternative_metrics: t.Optional[t.Mapping[str, AlternativeMetric]] = None
+    ):
+        super().__init__()
+        self.alternative_metrics: t.Optional[t.Mapping[str, MetricFunc]] = None
+
+        if alternative_metrics is not None and len(alternative_metrics) == 0:
+            raise ValueError('alternative_metrics - expected to receive not empty dict of scorers!')
```
matanper, line 53: raise Deepchecks error

deepchecks/checks/performance/class_performance_imbalance.py
```diff
@@ -0,0 +1,239 @@
...
+        self,
+        dataset: Dataset,
+        model: t.Any # TODO: find more precise type for model
+    ) -> CheckResult:
+        check_name = type(self).__name__
+        expected_model_types = [ModelType.BINARY, ModelType.MULTICLASS]
+
+        dataset.validate_label(check_name)
+        dataset.validate_features(check_name)
+        Dataset.validate_dataset(dataset, check_name)
```
matanper, line 102: Need to be first, if it's not Dataset object, the previous calls will fail on unwanted exception

deepchecks/checks/performance/class_performance_imbalance.py
```diff
@@ -0,0 +1,239 @@
...
+                model, features, labels
+            ))
+        else:
+            y_true = labels
+            y_pred = model.predict(features)
+            unique_labels = get_unique_labels(labels, y_pred)
+            df = pd.DataFrame.from_dict({
+                name: dict(zip(unique_labels, metric_func(y_true, y_pred, labels=unique_labels))) # type: ignore
+                for name, metric_func in self._get_default_metrics().items()
+            })
```
matanper, line 119: why not initalize dict like alternative_metrics and then have single 
```
df = pd.DataFrame.from_dict(self._execute_alternative_metrics(
                model, features, labels
            ))
```

deepchecks/checks/performance/class_performance_imbalance.py
```diff
@@ -0,0 +1,239 @@
...
+    ) -> CheckResult:
+        check_name = type(self).__name__
+        expected_model_types = [ModelType.BINARY, ModelType.MULTICLASS]
+
+        dataset.validate_label(check_name)
+        dataset.validate_features(check_name)
+        Dataset.validate_dataset(dataset, check_name)
+        task_type_validation(model, dataset, expected_model_types, check_name)
+
+        labels = t.cast(pd.Series, dataset.label_col())
```
matanper, line 105: what is this casting for? is it only for the developer to have autocomplete?

deepchecks/checks/performance/class_performance_imbalance.py
```diff
@@ -0,0 +1,239 @@
...
+            )
+
+        return CheckResult(
+            value=df.transpose().to_dict(),
+            header='Class Performance Imbalance',
+            check=type(self),
+            display=display
+        )
+
+    def _get_default_metrics(self) -> t.Dict[str, t.Callable[..., np.ndarray]]:
```
matanper, line 142: we have utils with get default metrics

tests/checks/performance/class_performance_imbalance_test.py
```diff
@@ -0,0 +1,197 @@
...
+        })
+    )
+
+    validate_class_performance_imbalance_check_result(
+        check_result,
+        value=value_matcher
+    )
+
+
+def test_class_performance_imbalance_with_custom_metrics_that_return_values_with_incorrect_dtype(
```
matanper, line 77: split this to different tests

tests/checks/performance/class_performance_imbalance_test.py
```diff
@@ -0,0 +1,197 @@
...
+)
+
+from deepchecks import Dataset, CheckResult, ConditionCategory
+from deepchecks.checks import ClassPerformanceImbalanceCheck
+from deepchecks.utils import DeepchecksValueError
+
+from tests.checks.utils import equal_condition_result
+
+
+def test_class_performance_imbalance(
```
matanper, line 17: Add annotations for each test like in other tests:
```
# Arrange
# Act
# Assert
```
right now I can't quickly understand what is being tested in each test

deepchecks/checks/performance/class_performance_imbalance.py
```diff
@@ -0,0 +1,239 @@
...
+
+            df.transpose().plot.bar(
+                title=title,
+                backend='matplotlib',
+                xlabel='Metrics',
+                ylabel='Values'
+            )
+
+        return CheckResult(
+            value=df.transpose().to_dict(),
```
matanper, line 136: why not instead of dict -> df -> dict, just get the dict and only in the display function convert it to dataframe?

deepchecks/checks/performance/class_performance_imbalance.py
```diff
@@ -0,0 +1,239 @@
...
+        }
+
+    def _execute_alternative_metrics(
+        self,
+        model: t.Any,
+        features: pd.DataFrame,
+        labels: pd.Series,
+    ) -> t.Dict[str, t.Dict[t.Hashable, t.Union[int, float]]]:
+        check_name = type(self).__name__
+        metrics = t.cast(t.Mapping[str, MetricFunc], self.alternative_metrics)
```
matanper, line 156: why do you need the cast?

deepchecks/checks/performance/class_performance_imbalance.py
```diff
@@ -0,0 +1,239 @@
...
+        result: t.Dict[str, t.Dict[t.Hashable, t.Union[int, float]]] = {}
+
+        for metric_name, metric_func in metrics.items():
+            metric_result = metric_func(model, features, labels)
```
matanper, line 160: I still don't understand how metric function returns a dict

matanper (changes requested):

deepchecks/checks/performance/class_performance_imbalance.py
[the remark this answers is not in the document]
matanper: Aren't we create features automatically in dataset init?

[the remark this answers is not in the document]
yromanyshyn: yes, but return type of the `dataset.features_columns()` is `Optional[DataFrame]`. It returns `None` if `dataset._features` list is empty or is `None`

deepchecks/checks/performance/class_performance_imbalance.py
yromanyshyn: type of `self.alternative_metrics` is `Optional[Dict]`, but we know that this method is called only when it is not `None`. But static type checker is not so "smart" to infer this, so we need to help him.

Mypy doc about `cast` - [link](https://mypy.readthedocs.io/en/stable/casts.html)

deepchecks/checks/performance/class_performance_imbalance.py
yromanyshyn: no, it is for the static type checker, to help him to infer of type the object is.

Mypy doc about cast - [link](https://mypy.readthedocs.io/en/stable/casts.html)

deepchecks/checks/performance/class_performance_imbalance.py
yromanyshyn: ok, but, I think that this error type is too overloaded with too many meanings and uses. In the python world, it is more common (appropriate) to use built-in `ValueError` to show that an incorrect value was passed to the function/

deepchecks/checks/performance/class_performance_imbalance.py
yromanyshyn: 1. some scorers will return dict of values per label if to pass `average=None`
2. do not we want to calc metric per label (class)?

matanper: yes, but if the user passes custom scorer I guess it will not return a dict.
@noamzbr if we filter only specific labels and run metrics on them it will work? or it will need also the false positive for this class?

deepchecks/checks/performance/class_performance_imbalance.py
matanper: We have a single exception type, but the idea is to indicate the exception is thrown specifically by our package and not by some dependency

deepchecks/checks/performance/class_performance_imbalance.py
yromanyshyn: from the `precision_score` doc string:

```
average : {'micro', 'macro', 'samples', 'weighted', 'binary'} or None, default='binary'
        This parameter is required for multiclass/multilabel targets.
        If ``None``, the scores for each class are returned. Otherwise, this
        determines the type of averaging performed on the data:
```

deepchecks/checks/performance/class_performance_imbalance.py
matanper: and do you think it's worth the extra code lines?

yromanyshyn: PEP about type annotations and `cast` - [link](https://www.python.org/dev/peps/pep-0484/#casts)

deepchecks/checks/performance/class_performance_imbalance.py
yromanyshyn: PEP about type annotations and `cast` - [link](https://www.python.org/dev/peps/pep-0484/#casts)

deepchecks/checks/performance/class_performance_imbalance.py
matanper: I feel like this makes the code a bit harder to read, but easier to write. I try to understand if you agree

also, is the next different?
```
labels: pd.Series = dataset.label_col()
```

yromanyshyn: > bit harder to read

maybe/

> but easier to write.

it is not about autocompletion, it is about keeping your code correct. Static type checker helps to find bugs (like None value bugs), errors, unappropriate use of classes/functions before running your program, before shipping it into production, so from my point of view it is definitely worth it

> also, is the next different?

yes, you are saying that `labels` var is of type `pd.Series` but then you are trying to assign a value of type `Union[pd.Series, None]` to it.

Whereas with help of `cast` you are saying the type checker that we are confident that the type of `labels` is `pd.Series`, and that from now he also must consider it of type `pd.Series`

deepchecks/checks/performance/class_performance_imbalance.py
yromanyshyn: yup, if the user passes a scorer function without applying `avarage=None`, then check will fail within the `_execute_alternative_metrics` method, with the next message error:
```
Check {check_name} expecting that alternative metrics will return
not empty instance of 'Mapping[Hashable, float|int]', but got {type(value)}
```

what I mean is that it is the user's responsibility to follow the check API contract.

deepchecks/checks/performance/class_performance_imbalance.py
yromanyshyn: we want the value to be in the format `{'label_name': {'metric_name': <value>}}`, without `DataFrame` we would need to make some extra work to achieve this.

What I mean is that with help of `DataFrame.transpose` implementation looks cleaner and more readable.

deepchecks/checks/performance/class_performance_imbalance.py
yromanyshyn: from another reply:

> we want the value to be in the format {'label_name': {'metric_name': value}}, without DataFrame we would need to 
> make some extra work to achieve this.
>
> What I mean is that with help of DataFrame.transpose implementation looks cleaner and more readable.

deepchecks/checks/performance/class_performance_imbalance.py
matanper: I'd edit it to something more readable, `metrics should return a dictionary of score per label class`

tests/checks/performance/class_performance_imbalance_test.py
yromanyshyn: updated

tests/checks/performance/class_performance_imbalance_test.py
yromanyshyn: updated

deepchecks/checks/performance/class_performance_imbalance.py
yromanyshyn: updated

deepchecks/checks/performance/class_performance_imbalance.py
matanper: I didn't understand your comment. I'm using dataframe in the example I wrote. I'm saying to reuse the `_execute_alternative_metrics` also in the else case.

yromanyshyn: oh, sorry I did not get you.

User-provided and our default metrics are different.
UserProvided metrics must have signature `Callable[[model, features, labels], Dict[str, int|float]]`
whereas our default metrics have signature `Callable[[y_true, y_pred], np.ndarray]`

therefore we process thems differently

deepchecks/checks/performance/class_performance_imbalance.py
noamzbr: it's not metrics it's scorers (difference is scorers receive (model, features, labels) and not (pred, labels))

deepchecks/checks/performance/class_performance_imbalance.py
yromanyshyn: hm/ it will not work

sorry, I answered one of the Matan questions previously incorrectly,
there is no scorer that returns dict if you pass `average=None`
they return only numbers or np.array

`classification_report` can return dict but not scorer functions

so, as a result:

1. 
there is actually no sense in passing the names of the sklearn scorers to the check, because we will not be able to use them`s to calculate metrics per class

2. 
in order to build a dict of values per class we need to zip list of unique labels with an array returned by the scorer, in the same way how it is done in the 'classification_report' function

from 'classification_report' (lines 2107-2116):

```
labels = ... # list of labels
target_names = ... # stringified list of labels
p, r, f1, s = precision_recall_fscore_support(
    y_true,
    y_pred,
    labels=labels,
    average=None,
    sample_weight=sample_weight,
    zero_division=zero_division,
)
rows = zip(target_names, p, r, f1, s)
```

matanper: so instead of using scorer that returns dict, maybe it better to use regular scorer and for each class transform it to 0 and 1:
```
for label in labels:
  df['target'].map(lambda x: 1 if x == label else 0)
  # calc score for label...
```

yromanyshyn: so, did I get you right that to calculate a metric for some specific class it is enough to have data (labels/predictions) only with that class? I mean that I do not need an entire dataset to calc metric for some specific class? and that it is true for any existing metric?

Mehdish1414 (dismissed):

deepchecks/checks/performance/class_performance_imbalance.py
matanper: I think so, but I don't see a reason to filter out rows with other target/predictions, it doesn't make it easier.
@noamzbr is that right?

noamzbr: Look at this [thread]( https://stats.stackexchange.com/questions/51296/how-do-you-calculate-precision-and-recall-for-multiclass-classification-using-co) to better understand.
Short version - it's calculated per class using the appropriate row in the confusion matrix, which is in turn calculated on *all* samples. You can see an example in [precision_recall_fscore_support](https://github.com/scikit-learn/scikit-learn/blob/0d378913b/sklearn/metrics/_classification.py#L1370). It's true that samples that were both not predicted as the inspected class AND their label is not the inspected class don't contribute to the calculation of per class precision/recall.

I'm for the approach @yromanyshyn suggested - zipping the labels with the np.array

deepchecks/checks/performance/class_performance_imbalance.py
noamzbr: You're right, we should rename it globally and make clear we're using scorers rather than metrics.

Unless you or @matanper think the standard of using scorers is wrong

matanper: I have no opinion in this matter, so let's go for scores

deepchecks/checks/performance/class_performance_imbalance.py
matanper: the zip sounds like technical stuff, the problem is implementing a function like  `precision_recall_fscore_support` which use scorer to calculate each class separatly

matanper: Another option is to use `utils.metrics.get_metrics_list` and add parameter to it "average". than of course we will have to change the rest of the implementation to use `make_scorer` with this param

yromanyshyn: >he zip sounds like technical stuff, the problem is implementing a function like precision_recall_fscore_support which use scorer to calculate each class separatly

I did not get you, You mean that it will be uneasy for the users to impl custom scorers for the check?

> Another option is to use utils.metrics.get_metrics_list and add parameter to it "average". than of course we will have to change the rest of the implementation to use make_scorer with this param

1. after we wrap the scorer function into `make_scorer` we cannot call it anymore with the `average` parameter, so we will not be able to use scorers returned by the `utils.metrics.get_metrics_list` whereas they are initialized with default `average` or with `average='macro'`
2. I do not think that we need `average=None` everywhere/

deepchecks/checks/performance/class_performance_imbalance.py
yromanyshyn: renamed, but only in this check. For others I think it is better to create separate PR (also created an issue for this - [MLC-301](https://deepchecks.atlassian.net/browse/MLC-301))

deepchecks/checks/performance/class_performance_imbalance.py
matanper: 1. that's why I said "we will have to change the rest of the implementation to use make_scorer with this param". we will create it on the fly

yromanyshyn: I updated implementation

deepchecks/checks/performance/class_performance_imbalance.py
```diff
@@ -0,0 +1,240 @@
...
+    np.ndarray # scores
+]
+
+AlternativeScorer = t.Union[str, ScorerFunc]
+
+
+CP = t.TypeVar('CP', bound='ClassPerformanceImbalanceCheck')
+
+
+class ClassPerformanceImbalanceCheck(SingleDatasetBaseCheck):
```
matanper, line 32: In other checks we don't use the word `check` in the class name, so better to have the same

deepchecks/checks/performance/class_performance_imbalance.py
```diff
@@ -0,0 +1,240 @@
...
+    ) -> t.Iterator[t.Tuple[str, np.ndarray]]:
+        # We need to be sure that user-provided scorers returned value with correct
+        # datatype, otherwise we need to raise an exception with an informative message
+        expected_types = t.cast(
+            str,
+            np.typecodes['AllInteger'] + np.typecodes['AllFloat'] # type: ignore
+        )
+        message = (
+            f"Check '{type(self).__name__}' expecting that scorer "
+            "'{scorer_name}' will return an instance of numpy array with "
```
matanper, line 167: missing f-string?

deepchecks/checks/performance/class_performance_imbalance.py
```diff
@@ -0,0 +1,240 @@
...
+                    additional=f"But got array of '{str(score.dtype)}'"
+                ))
+            if len(score) != number_of_classes:
+                raise DeepchecksValueError(message.format(
+                    scorer_name=scorer_name,
+                    additional=f"But got array with shape ({len(score)},)"
+                ))
+            yield scorer_name, score
+
+    def add_condition_ratio_difference_not_greater_than(self: CP, threshold: float = 0.3) -> CP:
```
matanper, line 189: do you think it makes sense to have single condition for many metrics? maybe we should limit the check to single metric

deepchecks/checks/performance/class_performance_imbalance.py
yromanyshyn: nope, it will be added below

deepchecks/checks/performance/class_performance_imbalance.py
yromanyshyn: do you suggest adding another parameter to the method, to narrow the condition to some specific metric?
I do not know/ @noamzbr what do you think?

deepchecks/checks/performance/class_performance_imbalance.py
yromanyshyn: oh, you are right, other check types do not use 'Check' noun in the name, renamed

deepchecks/checks/performance/class_performance_imbalance.py
noamzbr: Sound reasonable to limit to a single metric

matanper: I mean to limit the check itself, not the condition

yromanyshyn: oh, then, did I get you right?
- instead of calculating multiple scores by default you suggest calculating only one? (which one?)
- instead of accepting dict of alternative scorers you suggest accepting only one scorer?

@noamzbr your thoughts?

noamzbr: For this check we must calculate many scores. To understand imbalance I usually look at both recall and precision for e.g. In that light, I think it makes sense to limit the condition, and have the default one be on f1 (do we have f1 in the default scorers?)

yromanyshyn: > do we have f1 in the default scorers?

nope, but I added, also updated condition method

noamzbr (approved):
Approved. Only fix needed is removing accuracy from the default metric list (it's equivalent to recall in this case) and fixing the merge and notebook issues

deepchecks/checks/performance/class_performance_imbalance.py
Mehdish1414: Mehdish14

noamzbr (approved):

matanper (approved):

PR #198, opened 2021-11-28
custom check/condition/suite example

JKL98ISR:


deepchecks/base/check.py
```diff
@@ -148,9 +148,10 @@ def _ipython_display_(self):
             display_html(f'<h4>{self.header}</h4>', raw=True)
         if self.check and '__doc__' in dir(self.check):
             docs = self.check.__doc__
-            # Take first non-whitespace line.
-            summary = next((s for s in docs.split('\n') if not re.match('^\\s*$', s)), '')
-            display_html(f'<p>{summary}</p>', raw=True)
+            if docs:
```
matanper, line 151: a. why?
b. move the condition to the upper if instead of additional if

matanper (commented):
IMO the example is too complex, too much code. need to be simpler

shir22:
Hi, I didn't write any specific review yet because there were several comments that were discussed during the demo on Sunday, so I'm waiting for the updated version to review.
On a general note, following Matan's comment - I agree that we want the condition itself to be a relatively simple demonstration so it won't be too long (the logic itself), so that the user can quickly understand and focus on "how do they do this", "what do they need to know" (what to inherit from, how to make the check plot a graph ,etc.")

deepchecks/base/check.py
JKL98ISR: a. the original didn't worked. the __docs__ was None
b. wanted to not make duplicate code. doesn't really matter

matanper: I think that's matter, it's uglier to have `if` inside `if`
and it's not saving any work, since `self.check.__doc__` access a variable not a function call

JKL98ISR:
> Hi, I didn't write any specific review yet because there were several comments that were discussed during the demo on Sunday, so I'm waiting for the updated version to review. On a general note, following Matan's comment - I agree that we want the condition itself to be a relatively simple demonstration so it won't be too long (the logic itself), so that the user can quickly understand and focus on "how do they do this", "what do they need to know" (what to inherit from, how to make the check plot a graph ,etc.")

Just now noticing your comment. I updated the documentation btw.
About the check: I did the check you suggested, this check can't be done much less complex.

notebooks/examples/basic/custom_check_example.ipynb
```diff
@@ -0,0 +1,689 @@
+{
```
ItayGabbay, line 1: <pre class="ql-syntax" spellcheck="false">Line #49.                fig.suptitle(f'Disterbution of {self.selected_feature} [{min_val:.2}, {max_val:.2}]', fontsize=16)
</pre><p>"Distribution"</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/deepchecks/deepchecks/pull/198/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='14'/>

shir22:
Hey, I had another look, my main high-level comments are:

1. There is still more work in making this a "user facing" notebook. Note that this is going to be added as a "tutorial". So it should be much more clear, aesthetic and step-by-step explainable.

2. About the check and its output -
a. The current check is too complex, making the user probably lose us because of the amounts of code. We can consider adding it at the bottom as an "additional example for implementing a check"
b. We should think about a simpler custom check. The focus and target of this notebook are "how do I implement a custom check" and not "this is how a custom check with complex logic looks like". Feel free to suggest ideas and get feedback for them.  In addition I think it's better to demonstrate something that compares between two datasets, possibly using also the model and predictions (so that the user can derive from that also for the simpler use cases).
One example - add a custom metric (e.g. a metric that we don't have yet implemented for a regression / classification use case. such as Precision Recall curve, if we don't have it).
3. We discussed adding a demonstration about - "how do I add a condition with a lambda function to an existing check", this is missing as well

matanper:
Small and stupid check suggestion:  
name: Dataset Size Comparison
return value: { 'train': size, 'test': size}
display: graph with 2 bars
condition: ratio is between `x` to `y`

noamzbr:
> Small and stupid check suggestion: name: Dataset Size Comparison return value: { 'train': size, 'test': size} display: graph with 2 bars condition: ratio is between `x` to `y`

Yuri is adding a very similar check RN. I can suggest something similar to what Shir suggested, but with a single metric, such as [log-loss](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.log_loss.html#sklearn.metrics.log_loss), and then condition on ratio of log-loss between train and test

Different suggestion - naïve "bias" - e.g. in adult dataset check that performance on "Female" >= performance on "Male"

noamzbr:
> Small and stupid check suggestion: name: Dataset Size Comparison return value: { 'train': size, 'test': size} display: graph with 2 bars condition: ratio is between `x` to `y`

Nvm let's go with this

matanper:
> > Small and stupid check suggestion: name: Dataset Size Comparison return value: { 'train': size, 'test': size} display: graph with 2 bars condition: ratio is between `x` to `y`
> 
> Nvm let's go with this

Boom code ninja in your face

matanper:
Will do the example in another PR

PR #204, opened 2021-11-29
[MLC-242] added check decorators;

yromanyshyn:
We need to provide a less boilerplate way for our users to define a check and first what comes to mind - decorators. Example:

```python
from deepchecks import (
    single_dataset_check, 
    two_datasets_check, 
    train_test_datasets_check, 
    model_check, 
    CheckResult, 
    CheckSuite
)


@single_dataset_check
def custom_single_dataset_check(dataset, model=None) -> CheckResult:
    ....

@two_datasets_check
def custom_two_datasets_check(dataset, baseline_dataset, model=None) -> CheckResult:
    ....



suite = CheckSuite(
   'Custom Suite',
   custom_single_dataset_check,
   custom_two_datasets_check
)



```

### DIFF
--- a/deepchecks/base/check.py
+++ b/deepchecks/base/check.py
@@ -275,6 +275,9 @@ class BaseCheck(metaclass=abc.ABCMeta):
         return split_camel_case(cls.__name__)
 
 
+CheckFunc = Callable[..., CheckResult] # TODO: add more precise signature
+
+
 class SingleDatasetBaseCheck(BaseCheck):
     """Parent class for checks that only use one dataset."""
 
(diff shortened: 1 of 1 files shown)


matanper:
I don't understand why this is needed

yromanyshyn:
@matanper 

It could be useful for cases when the users want to define their own custom checks,
especially when those checks are simple and implementing whole class for it looks like something redundant

matanper:
I'm against the simplification (not specifically this, but any), so leaving this to you @noamzbr @ItayGabbay

yromanyshyn:
I am closing this PR whereas currently this functionality is not needed
MLC-242 will be reconsidered

PR #207, opened 2021-11-30
[MLC-236] added 'from_numpy' and 'from_dict' constructors to the Dataset class

yromanyshyn:


deepchecks/base/dataset.py
```diff
@@ -126,6 +144,168 @@ def __init__(self, df: pd.DataFrame,
...
+                    f'{features_array.shape[1]} features were provided '
+                    f'but only {len(feature_names)} name(s) for them`s.'
+                )
+            elif feature_names is None:
+                feature_names = [f'X{index}'for index in range(1, features_array.shape[1] + 1)]
+
+            return features_array, t.cast(t.List[str], feature_names)
+
+        if len(args) == 1:
+            features_array, features_names = validate_features_array(args[0], features)
```
matanper, line 236: maybe instead of creating inner func, can just move this logic before this `if`

deepchecks/base/dataset.py
```diff
@@ -126,6 +144,168 @@ def __init__(self, df: pd.DataFrame,
...
+
+            return cls(
+                df=pd.DataFrame(data=data, columns=columns),
+                features=features_names, # type: ignore TODO
+                label=label, # type: ignore TODO
+                **kwargs
+            )
+
+    @classmethod
+    def from_dict(
```
matanper, line 279: In which use cases from_dict is needed?

deepchecks/base/dataset.py
```diff
@@ -126,6 +144,168 @@ def __init__(self, df: pd.DataFrame,
...
+
+        >>> dataset = Dataset.from_numpy(
+        ...    features, labels,
+        ...    features=['sensor-1', 'sensor-2', 'sensor-3',],
+        ...    label='labels'
+        ... )
+
+        """
+        if len(args) == 0 or len(args) > 2:
+            raise ValueError(
```
matanper, line 207: Replace with deepchecks error

deepchecks/base/dataset.py
yromanyshyn: I thought it could be useful, but if it is not needed I will remove it

deepchecks/base/dataset.py
yromanyshyn: updated

deepchecks/base/dataset.py
yromanyshyn: updated

deepchecks/base/dataset.py
ItayGabbay: I think it could be useful

ItayGabbay (approved):

PR #208, opened 2021-11-30
updated readme

JKL98ISR:


README.md
```diff
@@ -51,20 +51,13 @@ pip install deepchecks #--user
...
 ```
-and then either:
+For stable tag:
 ```bash
-pip install .
-```
-or
-```bash
-python setup.py install
+pip install git+https://github.com/deepchecks/deepchecks.git@<tag_name>
```
ItayGabbay, line 60: Maybe write the latest tag and let's change every release?

README.md
```diff
@@ -51,20 +51,13 @@ pip install deepchecks #--user
...
 
 ### From source
-First clone the repository and then install the package from inside the repository's directory:
+For latest stable version:
 ```bash
-git clone https://github.com/deepchecks/deepchecks.git
-cd deepchecks
-# for installing stable tag version and not the latest commit to main
-git checkout tags/<name_of_latest_tag>
+pip install git+https://github.com/deepchecks/deepchecks.git@0.0.10
```
matanper, line 56: do we want to have the version in the README? we'll have to update on each release

JKL98ISR: I think it's usually like that, I also changed that according to gabbay's comment

shir22: I don't think it gives value to add which version
Lets leave it as is

JKL98ISR: ok

shir22 (approved):

PR #214, opened 2021-12-01
[MLC-153] new dataset column type

yromanyshyn:
There are two things that still need to be done (TODO):

1.
We need to verify all checks, that there is no function/method that tries to perform string operation over a column name without first checking that it is actually a string. Probably it is a good idea to add a test case to those checks for which this is actual. 

2.

A lot of routines in our code-base use the next dtype for the columns list argument - `Union[Hashable, Iterable[Hashable]]`. For example `LabelAmbiguity` check.

It raises a question - _**What if the provided value is hashable but also iterable?**_ like tuple for example
I think it is better to change those routines signature just to `Iterable[Hashable]`

@noamzbr @matanper @ItayGabbay what do you think?


matanper:
1. "perform string operation over a column name" - do you have an example? I can't think of a place it is done
2. I'm slightly against it. I think it's preferable to allow the users to pass a single column without list. 
also not sure how allowing only `Iterable[Hashable]` will work? `str` is also fit this so we can take each character in string as a column by mistake

yromanyshyn:
@matanper 

> do you have an example?

no, I have not seen, but it worth to check

> str is also fit this so we can take each character in the string as a column by mistake

we can add if-statement for strings, for example:

```python
def func(columns: Iterable[Hashable]):
    if isinstance(columns, AnyStr):
        columns = [columns]
    ...
```

> I'm slightly against it. 

oh, actually, the is the better solution, to always use unhashable sequence types like List or Set. Example

```python
def func(columns: t.Union[t.Hashable, t.List[t.Hashable]]):
    ...


isinstance([], t.Hashable) # false
isinstance(set([]), t.Hashable) # false
```

yromanyshyn:
@matanper 

> "perform string operation over a column name" - do you have an example? I can't think of a place it is done

actually, there was one problem. Often next line could be seen in the code:
```python
columns = [...]  # list of dataset columns (features)
.. ",".join(columns) ...
```

'join' method accepts only iterable of strings, if you try to pass something else it will raise `TypeError`

```python
>> l = [1,2,3]
>> "".join(l)
TypeError: sequence item 0: expected str instance, int found
```

matanper:
so how did you solve it at the end? what kind of objects do you allow to pass?

noamzbr:
> @matanper
> 
> > "perform string operation over a column name" - do you have an example? I can't think of a place it is done
> 
> actually, there was one problem. Often next line could be seen in the code:
> 
> ```python
> columns = [...]  # list of dataset columns (features)
> .. ",".join(columns) ...
> ```
> 
> 'join' method accepts only iterable of strings, if you try to pass something else it will raise `TypeError`
> 
> ```python
> >> l = [1,2,3]
> >> "".join(l)
> TypeError: sequence item 0: expected str instance, int found
> ```

Maybe we can solve it with ",".join(apply(str, columns)) ?

noamzbr (commented):
Looks good. About the points

1. Test - can you think on a way to test it efficiently in all checks? (without writing such a test for each check)
2. having hashable / list[hashable] seems good to me

yromanyshyn:
@noamzbr 

> Test - can you think on a way to test it efficiently in all checks?

currently, I do not know how to done it without writing a lot of tests/

@matanper 

> so how did you solve it at the end? what kind of objects do you allow to pass?

Do you mean how I deal with `'',join(columns)`?

I transformed it to `''.join(map(str, columns))`

noamzbr:
> @noamzbr
> 
> > Test - can you think on a way to test it efficiently in all checks?
> 
> currently, I do not know how to done it without writing a lot of tests/
> 
> @matanper
> 
> > so how did you solve it at the end? what kind of objects do you allow to pass?
> 
> Do you mean how I deal with `'',join(columns)`?
> 
> I transformed it to `''.join(map(str, columns))`

Think I have an idea - define a single test running the Overall suite with a data frames containing many different types of hashable column labels. It's not waterproof (not all checks are in suite). Need to check if you can call the display function in this check in order to test the display callables as well.

yromanyshyn:
@noamzbr sounds good, will implement

matanper:
> @noamzbr
> 
> > Test - can you think on a way to test it efficiently in all checks?
> 
> currently, I do not know how to done it without writing a lot of tests/
> 
> @matanper
> 
> > so how did you solve it at the end? what kind of objects do you allow to pass?
> 
> Do you mean how I deal with `'',join(columns)`?
> 
> I transformed it to `''.join(map(str, columns))`

I mean what is possible to pass as columns. Can I pass single string? Can I pass list? Can I pass set?

yromanyshyn:
@matanper 

> mean what is possible to pass as columns. Can I pass single string? Can I pass list? Can I pass set?

yes, yes, yes.

mutable sequence will be considered as a list of column names, 
unmutable hashable sequence will be considered as a single column name

matanper (dismissed):

yromanyshyn:
@noamzbr I added test, in general, I did not notice similar problems anymore

matanper (approved):

matanper (approved):

PR #217, opened 2021-12-02
[MLC-237] refactored model validation function

yromanyshyn:
description of the task from the ticket:

> The current model validation checks that the model can infer on a row, and raises a general error if it can’t. It would be better to have more informative - e.g. couldn’t infer because feature numbers don’t match, or order doesn’t match, or names don’t match, or types don’t match etc. I think sklearn raises these errors internally, maybe just catching them and showing to the user is enough. 

@noamzbr 
It looks like, if something is wrong sklearn will always raise an `ValueError`, so no need in catching different types of errors, let`s just append that exception message into our exception instance.


deepchecks/utils/validation.py
```diff
@@ -23,3 +24,48 @@ def model_type_validation(model: t.Any):
...
+            'But function received empty dataset!'
+        ))
+
+    try:
+        model_features = set(model_features) # type: ignore
+        if model_features != features_names:
+            raise errors.DeepchecksValueError(error_message.format(
+                'But function received dataset with a different set of features!'
+            ))
+    except (TypeError, ValueError):
```
matanper, line 62: why exception is needed instead of some `if model_features is not None`?

deepchecks/utils/validation.py
```diff
@@ -23,3 +24,48 @@ def model_type_validation(model: t.Any):
...
+        # in case if 'model.feature_names_in_' was None or not iterable
+        pass
+
+    try:
+        model.predict(features.head(1))
+    except Exception as exc:
+        raise errors.DeepchecksValueError(
```
matanper, line 69:
```suggestion
raise errors.DeepchecksValueError('error when trying to predict with model on dataset') from exc
```

deepchecks/utils/validation.py
yromanyshyn: but what if `model_features` is not None but also not an iterable?

deepchecks/utils/validation.py
yromanyshyn: we need to include caught exception message into our to be  more informative

deepchecks/utils/validation.py
```diff
@@ -23,3 +24,48 @@ def model_type_validation(model: t.Any):
...
+def validate_model(dataset: 'base.Dataset', model: t.Any):
+    """Check model is able to predict on the dataset.
+
+    Raise:
+        DeepchecksValueError: if dataset does not match model
+    """
+    model_type_validation(model)
+
+    error_message = (
+        'In order to evaluate model correctness we need not empty dataset '
```
noamzbr, line 38: drop the "!" in all places IMO

deepchecks/utils/validation.py
noamzbr: I think what @matanper proposed does exactly that

deepchecks/utils/validation.py
```diff
@@ -23,3 +24,48 @@ def model_type_validation(model: t.Any):
...
+            raise errors.DeepchecksValueError(error_message.format(
+                'But function received dataset with a different set of features!'
+            ))
+    except (TypeError, ValueError):
+        # in case if 'model.feature_names_in_' was None or not iterable
+        pass
+
+    try:
+        model.predict(features.head(1))
+    except Exception as exc:
```
noamzbr, line 68: Open question - should we be catching ValueError?

tests/base/feature_importance_utils_test.py
```diff
@@ -42,8 +42,27 @@ def test_calculate_importance(iris_labeled_dataset):
...
+        r'(In order to evaluate model correctness we need not empty dataset with the '
+        r'same set of features that was used to fit the model! But function received '
+        r'dataset with a different set of features!)'
+        # FIXME: 
+        # There is a problem, when I run this test locally, the 'validate_model' 
+        # function from the utils package as predicted raises DeepchecksValueError 
+        # with the message displayed above, but when it is run in the cloud (CI/CD) 
+        # the same function fails with different exception message (message below). 
+        # The reason for this is that when it is run on the cloud somewhy model instance 
+        # passed to the `calculate_feature_importance` does not have 'feature_names_in_' attribute
```
noamzbr, line 55: Sounds like an issue with scikitlearn version. Can check if have different scikitlearn versions locally and in cloud and if random forest has changed this attribute

deepchecks/utils/validation.py
yromanyshyn: hm/ not exactly, in general, it will concatenate tracebacks but will not include caught exception message into our, the user will see only our message

deepchecks/utils/validation.py
yromanyshyn: I saw that sklearn always raises ValueError, but I think there is nothing bad in catching more general error type/ just in case if in some rare situation it will raise something different

noamzbr: I can go with that

deepchecks/utils/validation.py
yromanyshyn: updated

tests/base/feature_importance_utils_test.py
yromanyshyn: Indeed, locally I use version 1.0.1, at the cloud (CI/CD) version 0.24 is used and that attribute was added at version 1.0

I removed my virtual env and reinstalled it with help of the `make env` and still got version 1.0.1/
I think that the python version which you use to install it could affect which package version will be used/

so, I suggest leaving this test as it is (but I updated the comment)

noamzbr: I'm ok with not testing for the specific message. Is that what the test does? Because there are two strings there

yromanyshyn:
@matanper @noamzbr do you have any additional comments/thoughts?

noamzbr (approved):
From my side good to go, just resolve the error-raising with @matanper

matanper:
> From my side good to go, just resolve the error-raising with @matanper

Doesn't matter much to me, so no more comments

matanper (approved):

PR #220, opened 2021-12-05
regression error distribution check

JKL98ISR:


deepchecks/checks/performance/regression_error_distribution.py
```diff
@@ -0,0 +1,89 @@
...
+        """Run check.
+
+        Arguments:
+            dataset (Dataset): A dataset object.
+            model (BaseEstimator): A scikit-learn-compatible fitted estimator instance
+
+        Returns:
+           CheckResult:
+                - value is a dict with mse and kurtosis values.
+                - display is histogram of error distirbution and the extreme prediction errors.
```
nirhutnik, line 36: again, would say "top ... errors" and not "extreme"

deepchecks/checks/performance/regression_error_distribution.py
```diff
@@ -0,0 +1,89 @@
...
+    def _regression_error_distribution(self, dataset: Dataset, model: BaseEstimator):
+        check_name = self.__class__.__name__
+        Dataset.validate_dataset(dataset, check_name)
+        dataset.validate_label(check_name)
+        task_type_validation(model, dataset, [ModelType.REGRESSION], check_name)
+
+        y_test = dataset.label_col()
+        y_pred = model.predict(dataset.features_columns())
+
+        mse = mean_squared_error(dataset.label_col(), y_pred, squared=False)
```
nirhutnik, line 52: So it is MSE and not RMSE? Why, just out of curiousity?

deepchecks/checks/performance/regression_error_distribution.py
```diff
@@ -0,0 +1,89 @@
...
+        """
+        def max_kurtosis_condition(result: float) -> ConditionResult:
+            kurtosis_value = result['kurtosis']
+            if abs(kurtosis_value) > max_kurtosis:
+                return ConditionResult(False, f'kurtosis: {format_number(kurtosis_value, 5)}')
+            else:
+                return ConditionResult(True)
+
+        return self.add_condition(f'Absolute kurtosis value is not greater than {format_number(max_kurtosis, 5)}',
+                                  max_kurtosis_condition)
```
nirhutnik, line 89: Why no condition on MSE?

nirhutnik (changes requested):

deepchecks/checks/performance/regression_error_distribution.py
JKL98ISR: it is rmse

deepchecks/checks/performance/regression_error_distribution.py
JKL98ISR: it's a bit hard to write a condition for it as the value too dependent on the data

JKL98ISR: maybe we don't really need to calculate it here even, we have a better usage of it in regression bias and we actually use it in a condition. what do you think? I think I should remove it

deepchecks/checks/performance/regression_error_distribution.py
JKL98ISR: sure

deepchecks/checks/performance/regression_error_distribution.py
```diff
@@ -0,0 +1,89 @@
...
+from deepchecks.utils.strings import format_number
+
+__all__ = ['RegressionErrorDistribution']
+
+
+class RegressionErrorDistribution(SingleDatasetBaseCheck):
+    """Calculate MSE and kurtosis, display an error histogram and most extreme prediction errors.
+
+    Args:
+        n_extreme (int): amount of samples to show which are of Largest under / over estimation errors.
```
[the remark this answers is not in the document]
JKL98ISR: sure

deepchecks/checks/performance/regression_error_distribution.py
```diff
@@ -0,0 +1,89 @@
...
+        n_smallest_diff = diff.nsmallest(self.n_extreme)
+        n_smallest_diff.name= n_smallest_diff.name + '_pred_diff'
+        n_smallest = pd.concat([dataset.data.loc[n_smallest_diff.index], n_smallest_diff], axis=1)
+
+        def display_hist():
+            diff = y_test - y_pred
+            diff.hist(bins = 40)
+            plt.title('Histogram of prediction errors')
+            plt.xlabel(f'{dataset.label_name()} prediction error')
+            plt.ylabel('Frequency')
```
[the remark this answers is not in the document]
JKL98ISR: ok

deepchecks/checks/performance/regression_error_distribution.py
```diff
@@ -0,0 +1,89 @@
...
+
+from deepchecks import CheckResult, Dataset, SingleDatasetBaseCheck, ConditionResult
+from deepchecks.utils.metrics import ModelType, task_type_validation
+from deepchecks.utils.strings import format_number
+
+__all__ = ['RegressionErrorDistribution']
+
+
+class RegressionErrorDistribution(SingleDatasetBaseCheck):
+    """Calculate MSE and kurtosis, display an error histogram and most extreme prediction errors.
```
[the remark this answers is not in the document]
JKL98ISR: ok

deepchecks/checks/performance/regression_error_distribution.py
matanper: why is it too dependent on the data? it's an averaged value

JKL98ISR: an error of 3000 can be big for some data and tiny for other data

nirhutnik: I agree its dependent on the data, and that's exactly why it should be a condition. Conditions are there for exactly this purpose, in my opinion. 
Also, would do this as a condition on average RMSE and not the most extreme sample.

JKL98ISR: average RMSE? 
and still, wouldn't it be better to just remove it. it sounds more like the bias check than distribution check

nirhutnik: Meant RMSE and not average RMSE. Meant not to check the largest error but the average one.
@noamzbr WDYT?

deepchecks/checks/performance/regression_error_distribution.py
```diff
@@ -0,0 +1,89 @@
...
+        check_name = self.__class__.__name__
+        Dataset.validate_dataset(dataset, check_name)
+        dataset.validate_label(check_name)
+        task_type_validation(model, dataset, [ModelType.REGRESSION], check_name)
+
+        y_test = dataset.label_col()
+        y_pred = model.predict(dataset.features_columns())
+
+        mse = mean_squared_error(dataset.label_col(), y_pred, squared=False)
+        diff = y_test - y_pred
```
matanper, line 53: don't you need to absolute value the diff?

JKL98ISR: no

JKL98ISR: if it was absolute the distribution will be wrong

matanper: Why? This is error, the direction doesn't matter

noamzbr: It does, it may be instructive to see if deviations from normal distribution are when under estimating or over estimating

deepchecks/checks/performance/regression_error_distribution.py
noamzbr: We already have that functionality (condition on metric) in PerformanceReport, so seems kinda redundant to have it here

notebooks/checks/performance/regression_error_distribution.ipynb
```diff
@@ -0,0 +1,339 @@
+{
```
noamzbr, line 1: Think that it would be nice to see the OK and Not OK cases in the notebook.

JKL98ISR:
switched to using p value calculated from kurtosistest in the condition

deepchecks/checks/performance/regression_error_distribution.py
```diff
@@ -0,0 +1,113 @@
...
+        kurtosis_value = kurtosis(diff)
+        # alternative doesn't work on pytho3.6
+        try:
+            kurtosis_pvalue = kurtosistest(diff, alternative = self.alternative).pvalue
+        except TypeError:
+            kurtosis_pvalue = kurtosistest(diff).pvalue
+
+        n_largest_diff = diff.nlargest(self.n_top_samples)
+        n_largest_diff.name= n_largest_diff.name + '_pred_diff'
```
noamzbr, line 80: Also show prediction. If showing a diff column, it needs to have a nice name such as "Target Prediction Difference"

deepchecks/checks/performance/regression_error_distribution.py
```diff
@@ -0,0 +1,113 @@
...
+        task_type_validation(model, dataset, [ModelType.REGRESSION], check_name)
+
+        y_test = dataset.label_col
+        y_pred = model.predict(dataset.features_columns)
+
+        diff = y_test - y_pred
+        kurtosis_value = kurtosis(diff)
+        # alternative doesn't work on pytho3.6
+        try:
+            kurtosis_pvalue = kurtosistest(diff, alternative = self.alternative).pvalue
```
noamzbr, line 75: Why do the switch to p-value? Because it does not depend on the target units?
One problem I can think about concerning p-values is that they do not distinguish between the "fat tail" case and the "narrow distributing" case. We would like to have the default condition fail only on the fat tail case.

notebooks/checks/performance/regression_error_distribution.ipynb
```diff
@@ -0,0 +1,618 @@
...
+ "cells": [
```
noamzbr, line 2: Does the condition pass for 1 and fail for 2?

deepchecks/checks/performance/regression_error_distribution.py
JKL98ISR: I went with the logics that if scipy thinks you should diffultivly check on both caes we should too

notebooks/checks/performance/regression_error_distribution.ipynb
JKL98ISR: yes

deepchecks/checks/performance/regression_error_distribution.py
JKL98ISR: ok

deepchecks/checks/performance/regression_error_distribution.py
noamzbr: Scipy is checking if the distribution is normal, we're checking if it has fat tails, so it makes sense we'll want different things.
I have no problem with having both the value and p value, and consider the p value only if the value is negative (or whatever side means fat tails)

JKL98ISR: it's problematic because we can't use it in python 3.6

JKL98ISR: I think that because both cases should probably be looked into it's ok, even though maybe one is more important

nirhutnik (approved):

PR #224, opened 2021-12-06
License add

ItayGabbay:


.github/workflows/build.yml
```diff
@@ -1,3 +1,19 @@
...
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
```
shir22, line 16:
```suggestion
# ----------------------------------------------------------------------------
# Copyright (C) 2021 Deepchecks (https://www.deepchecks.com)
#
# This file is part of Deepchecks.
#
# Deepchecks is free software: you can redistribute it and/or modify
# it under the terms of the GNU Affero General Public License as published by
# the Free Software Foundation, either version 3 of the License, or
# (at your option) any later version.
#
# Deepchecks is distributed in the hope that it will be useful,
# but WITHOUT ANY WARRANTY; without even the implied warranty of
# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
# GNU Affero General Public License for more details.
#
# You should have received a copy of the GNU Affero General Public License
# along with Deepchecks.  If not, see <http://www.gnu.org/licenses/>.
# ----------------------------------------------------------------------------
```

shir22: Suggestion above is following the second part in "The license notices" section here:
https://www.gnu.org/licenses/gpl-howto.html

If we want an alternative (shorter version), inspired by - https://github.com/pyinstaller/pyinstaller/blob/develop/PyInstaller/compat.py
we can go for:

shir22:
```suggestion
# ----------------------------------------------------------------------------
# Copyright (C) 2021 Deepchecks (https://www.deepchecks.com)
#
# This file is part of Deepchecks.
# Deepchecks is distributed under the terms of the GNU Affero General
# Public License (version 3 or later).
# You should have received a copy of the GNU Affero General Public License
# along with Deepchecks.  If not, see <http://www.gnu.org/licenses/>.
# ----------------------------------------------------------------------------
```

shir22: Personally I prefer the second one since it's much shorter (and it does follow the lines of "Why license notices" here:
https://www.gnu.org/licenses/gpl-howto.html)
and it seems that the header wording isn't strict (and I assumer that the linter can be configured).

shir22 (changes requested):
Left my comments on the first *.yml
(The first file with python-style comment string in the review)
lets finalize there before replacing for all

shir22 (approved):

PR #226, opened 2021-12-06
[MLC-241] dynamic suite output order;

yromanyshyn:


deepchecks/base/display_suite.py
```diff
@@ -46,11 +48,18 @@ def get_display_exists_icon(exists: bool):
     return '<div style="text-align: center">No</div>'
 
 
-def display_suite_result(suite_name: str, results: List[Union[CheckResult, CheckFailure]]):
+def display_suite_result(
+    suite_name: str,
+    results: List[Union[CheckResult, CheckFailure]],
+    output_order: Optional['suite.SuiteOutputOrder'] = None
```
matanper, line 54: right now we have a single option order, so I don't think this is needed

matanper, line 54: right now we have a single option order, so I don't think this is needed

deepchecks/base/display_suite.py
```diff
@@ -68,6 +77,22 @@ def display_suite_result(suite_name: str, results: List[Union[CheckResult, Check
             name = result.check.name()
             others_table.append([name, msg, 1])
 
+    if output_order == suite.SuiteOutputOrder.FAILED_FIRST:
+
+        def key(item: CheckResult) -> int:
```
matanper, line 82: there is already `CheckResult.get_conditions_sort_value()`

matanper (changes requested):

deepchecks/base/display_suite.py
yromanyshyn: not sure if I understood you
in the ticket, you described two options?

how exactly it should be, `FAILED_FIRST` by default?

yromanyshyn:
do not know how, but I think I accidentally merged this PR branch with some another branch, I will recreate it

PR #227, opened 2021-12-07
License add

ItayGabbay:


.github/workflows/build.yml
```diff
@@ -51,4 +51,35 @@ jobs:
...
+      uses: pilosus/action-pip-license-checker@v0.6.1
+      with:
+        requirements: 'requirements-all.txt'
+        fail: 'Copyleft,Other,Error'
+        exclude: '(visions.*|pyzmq.*|debugpy.*|certifi.*|tqdm.*)'
+        # visions is BSD-4
+        # pyzmq is Revised BSD
+        # debugpy is EPL-2.0
+        # certifi is MPL-2.0
+        # tqdm is MPL-2.0
```
shir22, line 82: 1. Add the various versions explicitly so we'll make sure to keep this up to date
2. Add a link to license file for each package, for convenient verification
3. Optional - because I'm not sure if possible, but if it is please do so - is there a way to verify that the excluded files are really in use? (where found) and if not it should fail, just to make sure that items here won't be stale over time.

.github/workflows/build.yml
```diff
@@ -51,4 +51,35 @@ jobs:
...
+    - name: Get explicit and transitive dependencies
+      run: |
+        pip install -r requirements.txt
+        pip freeze > requirements-all.txt
+    - name: Check python
+      id: license_check_report
+      uses: pilosus/action-pip-license-checker@v0.6.1
+      with:
+        requirements: 'requirements-all.txt'
+        fail: 'Copyleft,Other,Error'
```
shir22, line 76: Can we change to an include(white)-list approach and not an an exclude(black)-list one?

(or alternatively have the 'fail' categories interpreted dynamically somehow as all of the categories that aren't Permissive. In case they change...)

shir22 (changes requested):

.github/workflows/build.yml
ItayGabbay: 1. Done
2. Done
3. Not easily possible :/

.github/workflows/build.yml
shir22: No option, but it includes all non-permissive

shir22 (approved):

PR #240, opened 2021-12-09
[MLC-241] output order

yromanyshyn:


deepchecks/base/display_suite.py
```diff
@@ -78,6 +79,8 @@ def display_suite_result(suite_name: str, results: List[Union[CheckResult, Check
             name = result.check.name()
             others_table.append([name, msg, 1])
 
+    display_table = sorted(display_table, key=lambda it: it.get_conditions_sort_value())
```
matanper, line 82: will it result is same order as conditions table? 
maybe better to sort the `results` var and than condition & display will run on same sort

yromanyshyn: 1. Sorry that have not been responding to this PR for so long!
2. yup, you are right, it will not work/ I modified the code


changes:

    added `def priority() -> int` property to the CheckResult, that will:
        - return 1 if at least one condition did not pass and is of category FAIL
        - return 2 if at least one condition did not pass and is of category WARN
        - return 3 if check result does not have assigned conditions
        - return 4 if all conditions passed
    
    renamed `Condition.get_sort_value() -> int` to `Condition.priority() -> int`, it returns:
        - 1 if condition did not pass and is of category FAIL
        - 2 if condition did not pass and is of category WARN
        - 3 if condition passed

matanper (approved):

PR #242, opened 2021-12-10
[Refactoring] move `validate_*` methods from  the `Dataclass` to the `BaseCheck` 

yromanyshyn:
Purpose of the PR:
- move validation methods (methods that start with `validate_`) from `Dataset` class to the `BaseCheck` class

Reasoning:
- despite the fact that it looks like a totally logical to define all those methods within the Dataset class, I think they are not needed there
- hardcoded messages like `Check requires ...` make them totally unrelevant to the Dataset context, and also make them`s unusable anywhere except check instance context
- they are used only by checks

list of changes (not full):
- I moved validation methods from the `Dataset` class to the `BaseCheck` class
- I modified those methods signature to make them easier to use
- I added the next `classmethod`s to the Dataset class: 
   + `share_same_label(*datasets) -> bool`
   + `share_same_categorical_features(*datasets) -> bool`
   + `share_same_features(*datasets) -> bool`
- I renamed `Dataset.filter_columns_with_validation` to `Dataset.select`
- I added the next functions to the `deepchecks.utils.validation module`:
   + `ensure_dataset(obj: object, error_message: Optional[str] = None) -> base.Dataset`
   + `ensure_not_empty_dataset(obj: object, error_messages: Optional[t.Dict[str, str]] = None) -> base.Dataset`
   + `ensure_dataframe(obj: object, error_message: Optional[str] = None) -> pd.DataFrame`
   + `ensure_not_empty_dataframe(obj: object, error_messages: Optional[t.Dict[str, str]] = None) -> pd.DataFrame`

Note: This PR is not finished, before fixing all checks and making sure that all tests pass, I would like to hear your thoughts and comments

### DIFF
--- a/deepchecks/base/check.py
+++ b/deepchecks/base/check.py
@@ -9,6 +9,7 @@
 # ----------------------------------------------------------------------------
 #
 """Module containing all the base classes for checks."""
+import typing as t
 import abc
 import enum
 import re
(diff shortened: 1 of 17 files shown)


ItayGabbay (commented):
Nice PR!

Overall, I'm in favor of this change. I agree it's better to have the datasets validation in the check scope and not on the dataset scope since the checks are the only ones that require something on the dataset.

The only thing I'd change is that the validation functions will receive a list of datasets [], instead of *datasets. in my opinion, this is more readable when using the validation functions.

yromanyshyn:
@ItayGabbay 

>Nice PR!
> Overall, I'm in favor of this change. I agree it's better to have the datasets validation in the check scope and not on the dataset scope since the checks are the only ones that require something on the dataset.
> 
> The only thing I'd change is that the validation functions will receive a list of datasets [], instead of *datasets. in my opinion, this is more readable when using the validation functions.
> 

Thanks) 
Personaly I think that this is more readable:
`validate(dataset1, dataset2)`  instead of `validate([dataset1, dataset2])`
but, ok

@matanper @noamzbr do you have any thoughts/comments regarding this?

Also, this branch is already far behind the main, it will be too time-consuming, and error-prone to try to resolve all merge conflicts, therefore I will split it into several smaller PRs

yromanyshyn:
fixed by #640

PR #243, opened 2021-12-10
[MLC-364] replaced occurrence of 'metric' noun with 'score(r)' within 'deepchecks.checks.performance' package

yromanyshyn:
@noamzbr what about the occurrence of 'metric in other packages?


deepchecks/checks/performance/simple_model_comparison.py
```diff
@@ -126,27 +131,30 @@ def _find_score(self, train_ds: Dataset, test_ds: Dataset, task_type: ModelType,
...
-        if self.metric is not None:
-            scorer = validate_scorer(self.metric, model, train_ds)
-            metric_name = self.metric_name or self.metric if isinstance(self.metric, str) else 'User metric'
+        if self.scorer is not None:
+            scorer = validate_scorer(self.scorer, model, train_ds)
+            scorer_name = self.scorer if isinstance(self.scorer, str) else 'User Scorer'
         else:
-            metric_name = DEFAULT_SINGLE_METRIC[task_type]
-            scorer = DEFAULT_METRICS_DICT[task_type][metric_name]
+            scorer_name = DEFAULT_SINGLE_METRIC[task_type]
```
noamzbr, line 151: Also rename these const dicts to DEFAULT_SINGLE_SCORER etc

noamzbr (commented):
@yromanyshyn Could you give me an example of usage of this terminology in other packages that we should consider?

@shir22 Please have a look, what do you think? I think it's good we cleared up the terminology, and we should add a link to here https://scikit-learn.org/stable/modules/model_evaluation.html in out user guide to explain what we mean by that usage throughout the package.

deepchecks/checks/performance/performance_report.py
```diff
@@ -64,12 +67,12 @@ def add_condition_score_not_less_than(self, min_score: float):
...
             min_score (float): Minimal score to pass.
         """
-        name = f'Metrics score is not less than {min_score}'
+        name = f'Score is not less than {min_score}'
 
         def condition(result, min_score):
             not_passed = {k: v for k, v in result.items() if v < min_score}
             if not_passed:
-                details = f'Metrics with lower score: {not_passed}'
+                details = f'Scores that are less than {min_score}: {not_passed}'
```
matanper, line 75: no need to repeat the `min_score` since it's already in the name

deepchecks/checks/performance/simple_model_comparison.py
```diff
@@ -126,27 +131,30 @@ def _find_score(self, train_ds: Dataset, test_ds: Dataset, task_type: ModelType,
             elif task_type in (ModelType.BINARY, ModelType.MULTICLASS):
                 clf = DecisionTreeClassifier(max_depth=self.max_depth,
                                              random_state=self.random_state, class_weight='balanced')
-            if clf:
+
+            if clf: # FIXME: is 'clf' possibly unbound?
```
matanper, line 135: looks like instead of this if it need to be an `else: raise exception` after the elif

yromanyshyn: updated

deepchecks/checks/performance/performance_report.py
yromanyshyn: updated

deepchecks/checks/performance/simple_model_comparison.py
yromanyshyn: updated

yromanyshyn:
>@yromanyshyn Could you give me an example of usage of this terminology in other packages that we should consider?

@noamzbr in next modules:
- `deepchecks/checks/methodology/boosting_overfit.py`
- `deepchecks/checks/methodology/performance_overfit.py`

noamzbr:
> 

Oh sorry didn't realize that. Yes we should replace there as well, I just forgot about these occurrences.

deepchecks/checks/methodology/boosting_overfit.py
```diff
@@ -113,19 +113,19 @@ class BoostingOverfit(TrainTestBaseCheck):
...
 
     The check runs a pred-defined number of steps, and in each step it limits the boosting model to use up to X
-    estimators (number of estimators is monotonic increasing). It plots the given metric calculated for each step for
+    estimators (number of estimators is monotonic increasing). It plots the given score calculated for each step for
     both the train dataset and the test dataset.
 
     Args:
-        metric (Union[Callable, str]): Metric to use verify the model, either function or sklearn scorer name.
-        metric_name (str): Name to be displayed in the plot on y-axis. must be used together with 'metric'
+        scorer (Union[Callable, str]): Scorer to use verify the model, either function or sklearn scorer name.
```
shir22, line 120:
```suggestion
        scorer (Union[Callable, str]): Scorer used to verify the model, either function or sklearn scorer name.
```

shir22: and a question - @noamzbr how is it possible with a function? isn't it a scorer object? https://scikit-learn.org/stable/modules/model_evaluation.html#defining-your-scoring-strategy-from-metric-functions

shir22 (changes requested):

deepchecks/checks/methodology/boosting_overfit.py
noamzbr: A scoring object is a function (Callable), they say object but mean a Callable (they even state it explicitly later on)

shir22:
@noamzbr I thinks it's great that we cleared the terminology...!
I think after the small fixes we're good to go :)

Simultaneously - the next step (in a separate task) is to align how/where we explain the general scorer usage concept.
I think its best if we have a concise text that will appear directly in the docstring of all of the functions that use it, with links to relevant places in sklearn docs. In addition we should add something to our user guide but I think the user experience will be nicer if they get the direct explanation (and link to origin) in each check's arguments, WDYT?

deepchecks/checks/methodology/boosting_overfit.py
shir22: 👍

deepchecks/checks/methodology/performance_overfit.py
```diff
@@ -39,11 +39,12 @@ class TrainTestDifferenceOverfit(TrainTestBaseCheck):
     the difference in performance between the two datasets. If no 'alternative_scorers' are supplied, the check would
     use a list of default scorers. If they are supplied, 'alternative_scorers' must be a dictionary, with the keys
     being score names and the values being either a name of an sklearn scoring function
-    (https://scikit-learn.org/stable/modules/model_evaluation.html#scoring) or an sklearn scoring function.
+    (https://scikit-learn.org/stable/modules/model_evaluation.html#common-cases-predefined-values)
+    or an sklearn scorer object (https://scikit-learn.org/stable/modules/model_evaluation.html#defining-your-scoring-strategy-from-metric-functions).
```
shir22, line 43: Possible to format it so #plyint disable won't be necessary?

shir22 (changes requested):

deepchecks/checks/methodology/performance_overfit.py
yromanyshyn: fixed

deepchecks/checks/methodology/performance_overfit.py
```diff
@@ -34,24 +34,30 @@
...
-    (https://scikit-learn.org/stable/modules/model_evaluation.html#scoring) or an sklearn scoring function.
+    The check would display the selected scores for the training and test data, helping the user visualize
+    the difference in performance between the two datasets. If no 'alternative_scorers' are supplied, the check would
+    use a list of default scorers. If they are supplied, 'alternative_scorers' must be a dictionary, with the keys
+    being score names and the values being either a name of an sklearn scoring function
+    (`Scikit-learn scorers`_) or an sklearn scorer object (`Scikit-learn - defining scoring strategy`_).
 
     Args:
-        alternative_metrics (Dict[str, Callable]): An optional dictionary of metric name or scorer functions
+        alternative_scorers (Dict[str, Callable]): An optional dictionary of scorer name or scorer functions
```
shir22, line 46: This looks good.
Lets see how it looks in the API reference (and in the auto-complete in jupyter), and then update it to be the documentation also for other classes that can receive other scorers (e.g. in boosting_overfit)

shir22 (approved):

PR #245, opened 2021-12-11
Mlc 225 domain classifier

noamzbr:


deepchecks/checks/distribution/whole_dataset_drift.py
```diff
@@ -0,0 +1,234 @@
...
+
+        Check trains a model to distinguish between train and test datasets, called a domain classifier. Then,
+        it calculates the feature importance for the domain classifier model. The result of the check is based on the
+        AUC of the domain classifier model, and the check displays the change in distribution between train and test
+        for the top features according to the calculated feature importance.
+
+        Args:
+            n_top_features (int):
+                Amount of columns to show ordered by domain classifier feature importance.
+            min_feature_importance (float): Minimum feature importance to show in the check display.
```
nirhutnik, line 49: As "feature importance" is not exactly a well-defined size but more of a comparable number, do you feel comfortable with filtering by min?

nirhutnik: Also, should say that you can either use top n or min feature importance, but not both

deepchecks/checks/distribution/whole_dataset_drift.py
```diff
@@ -0,0 +1,234 @@
...
+            max_num_categories (int):
+                Only for categorical columns. Max number of allowed categories. If there are more,
+                they are binned into an "Other" category in the display. If max_num_categories=None, there is no limit.
```
nirhutnik, line 52: Unclear explanation. Doesn't mean "allowed categories" IMO, as it won't fail on more categories.

nirhutnik: Should also be clear that this refers to category encoding (and the fact that we use one-hot)

deepchecks/checks/distribution/whole_dataset_drift.py
```diff
@@ -0,0 +1,234 @@
...
+import matplotlib.pyplot as plt
+
+__all__ = ['WholeDatasetDrift']
+
+
+class WholeDatasetDrift (TrainTestBaseCheck):
+    """
+        Calculate drift between train and test datasets as a whole, using a model trained to distinguish between them.
+
+        Check trains a model to distinguish between train and test datasets, called a domain classifier. Then,
```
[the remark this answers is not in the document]
nirhutnik: Also, as the phrased is used throughout the documentation, I would make the words "Domain Classifier" more dominant and maybe define in a different line

deepchecks/checks/distribution/whole_dataset_drift.py
```diff
@@ -0,0 +1,234 @@
...
+
+        Args:
+            n_top_features (int):
+                Amount of columns to show ordered by domain classifier feature importance.
+            min_feature_importance (float): Minimum feature importance to show in the check display.
+            max_num_categories (int):
+                Only for categorical columns. Max number of allowed categories. If there are more,
+                they are binned into an "Other" category in the display. If max_num_categories=None, there is no limit.
+            sample_size (int):
+                Amount of rows to use from each dataset for the training of the domain classifier.
```
nirhutnik, line 54: Max num, not amount

deepchecks/checks/distribution/whole_dataset_drift.py
```diff
@@ -0,0 +1,234 @@
...
+            random_state (int):
+                Random seed for the check.
```
nirhutnik, line 56: Do you want to explain for what it is used? for sampling and model training, I guess?

deepchecks/checks/distribution/whole_dataset_drift.py
```diff
@@ -0,0 +1,234 @@
...
+        self.min_feature_importance = min_feature_importance
+        self.max_num_categories = max_num_categories
+        self.sample_size = sample_size
+        self.random_state = random_state
+
+    def run(self, train_dataset, test_dataset, model=None) -> CheckResult:
+        """Run check.
+
+        Args:
+            train_dataset (Dataset): The training dataset object. Must contain a label column.
```
nirhutnik, line 80: Why must contain a label?

deepchecks/checks/distribution/whole_dataset_drift.py
```diff
@@ -0,0 +1,234 @@
...
+                         comparing the train and test distributions.
+
+        Raises:
+            DeepchecksValueError: If the object is not a Dataset or DataFrame instance
+        """
+
+        train_dataset = Dataset.validate_dataset_or_dataframe(train_dataset)
+        test_dataset = Dataset.validate_dataset_or_dataframe(test_dataset)
+
+        features = train_dataset.validate_shared_features(test_dataset, self.__class__.__name__)
```
nirhutnik, line 98: I think Matan updated the name of the class as an attribute of self, so you can use that instead of self.__class__.__name__. I think

deepchecks/checks/distribution/whole_dataset_drift.py
```diff
@@ -0,0 +1,234 @@
...
+        cat_features = train_dataset.validate_shared_categorical_features(test_dataset, self.__class__.__name__)
+        self._cat_features = cat_features
+
+        model = self._generate_model(list(set(features) - set(cat_features)), cat_features)
```
nirhutnik, line 102: Maybe call the variable "domain_classifier"?

deepchecks/checks/distribution/whole_dataset_drift.py
nirhutnik: Also, you only use 70% of that sample to train, and the rest to test, so the wording here is inaccurate

deepchecks/checks/distribution/whole_dataset_drift.py
```diff
@@ -0,0 +1,234 @@
...
+
+        # create new dataset, with label denoting whether sample belongs to test dataset
+        domain_class_df = pd.concat([train_sample_df, test_sample_df])
+        domain_class_labels = pd.Series([0] * len(train_sample_df) + [1] * len(test_sample_df))
+
+        from sklearn.model_selection import train_test_split
+        x_train, x_test, y_train, y_test = train_test_split(domain_class_df, domain_class_labels,
+                                                            stratify=domain_class_labels,
+                                                            random_state=self.random_state,
+                                                            test_size=0.3)
```
nirhutnik, line 116: nitpicking and no one will use it, and yet - I think test_size should be a check parameter and not hard-coded here

deepchecks/checks/distribution/whole_dataset_drift.py
```diff
@@ -0,0 +1,234 @@
...
+
+        model = model.fit(x_train, y_train)
+
+        y_test.name = 'belongs_to_test'
+        domain_test_dataset = Dataset(pd.concat([x_test.reset_index(drop=True), y_test.reset_index(drop=True)], axis=1),
+                                      cat_features=cat_features, label='belongs_to_test')
+
+        # calculate feature importance of model, containing the information of what features can separate the datasets
```
nirhutnik, line 124: "which features separate the dataset best" - WDYT?

deepchecks/checks/distribution/whole_dataset_drift.py
```diff
@@ -0,0 +1,234 @@
...
+        }
+
+        headnote = f"""<span>
+                    The shown features are the features that are most important for the domain classifier - the model
+                    trained to distinguish between the train and test datasets.<br> The percents of explained dataset 
+                    difference are the calculated feature importance values for the feature.
+                </span>"""
+
+        top_fi = fi_ser.head(self.n_top_features)
+        top_fi = top_fi.loc[top_fi > self.min_feature_importance]
```
nirhutnik, line 141: should be one or the other IMO, not both

deepchecks/checks/distribution/whole_dataset_drift.py
```diff
@@ -0,0 +1,234 @@
...
+        class PandasSimpleImputer(SimpleImputer):
+            """A wrapper around `SimpleImputer` to return data frames with columns.
+            """
+
+            def fit(self, x, y=None):
+                self.columns = x.columns
+                return super().fit(x, y)
+
+            def transform(self, x):
+                return pd.DataFrame(super().transform(x), columns=self.columns)
```
nirhutnik, line 194: Why not put in dist_utils? Weird to me that you define a useful class inside a private function

deepchecks/checks/distribution/whole_dataset_drift.py
```diff
@@ -0,0 +1,234 @@
...
+                   ("model", HistGradientBoostingClassifier(
+                       max_depth=2, max_iter=10, random_state=self.random_state,
+                       categorical_features=[False] * len(numerical_columns) + [True] * len(categorical_columns)
+                                                            ))])
+
+    def add_condition_overall_drift_value_not_greater_than(self, max_drift_value: float = 0.3):
+        """Add condition.
+
+        Overall drift score, calculated as (2 * AUC - 1) for the AUC of the dataset discriminator model, is not greater
+        than the specified value.
```
nirhutnik, line 219: Please explain here why and what is the meaning of this value

nirhutnik (changes requested):

deepchecks/checks/distribution/__init__.py
```diff
@@ -11,3 +11,4 @@
 """Module containing all data distribution checks."""
 from .trust_score_comparison import *
 from .train_test_drift import *
+from .whole_dataset_drift import *
```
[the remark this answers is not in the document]
noamzbr: @shir22 WDYT?

deepchecks/checks/distribution/whole_dataset_drift.py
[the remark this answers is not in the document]
noamzbr: modified a bit

deepchecks/checks/distribution/whole_dataset_drift.py
noamzbr: it's actually is an "AND".
Also, added some explenation

deepchecks/checks/distribution/whole_dataset_drift.py
noamzbr: It's actually only used here for the display plots, but edited.

deepchecks/checks/distribution/whole_dataset_drift.py
noamzbr: Don't think it's needed

deepchecks/checks/distribution/whole_dataset_drift.py
noamzbr: IMO both is fine, I do want to hard limit to a certain number, and also cut off unseals features.

deepchecks/checks/distribution/__init__.py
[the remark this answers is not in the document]
shir22: While I agree that "whole" isn't scientific, I think that multivariate drift isn't a good match to this scenario.
Since here we're explicitly talking about Dataset 1 vs. Dataset 2 -> are they drifted one compared to the other, while not necessarily the same dataset, and necessarily all of the features. I think multivariate drift doesn't encapsulate that meaning and may confuse with other options.

Completely open to alternative suggestions, but between the two offered here, whole_dataset_drift is clearer and less confusing IMO.

deepchecks/checks/distribution/whole_dataset_drift.py
shir22: What happens if no features pass this minimum?

deepchecks/checks/distribution/train_test_drift.py
```diff
@@ -27,113 +27,9 @@
...
-    if val_max == val_min:
-        return 0
-
-    dist1 = (dist1 - val_min) / (val_max - val_min)
-    dist2 = (dist2 - val_min) / (val_max - val_min)
-
-    return wasserstein_distance(dist1, dist2)
-
-
 class TrainTestDrift(TrainTestBaseCheck):
```
shir22, line 30: Separate thought - shouldn't this be named something like "TrainTestPerFeatureDrift"?
for a clearer indication of what it does

shir22 (changes requested):
Main review is on notebook

notebooks/checks/distribution/whole_dataset_drift.ipynb
```diff
@@ -0,0 +1,280 @@
+{
```
shir22, line 1: <p>The thing I'm most missing in this check's output is the "bottom line", which should appear first. </p><p>A clear indicator to the question "are they drifted"? and "how much"? (which is also the results on which the condition is applied, right?)</p><p> </p><p>The rest of the graphs are great to explain the drift, but that's interesting mainly if they are drifted...</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/deepchecks/deepchecks/pull/245/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

shir22: and please add conditions' demonstration to notebook

deepchecks/checks/distribution/train_test_drift.py
noamzbr: Seems a bit clumsy no? 
Maybe TrainTestFeatureDrift is a good compromise
Or drop the TrainTest?

notebooks/checks/distribution/whole_dataset_drift.ipynb
noamzbr: Condition - will do
Bottom line - I can add a traffic light bar like in the regular drift as the first plot.

As for your other question on "what happens if no features pass" - then none are shown, but feature importance tends to be distributed non-evenly, so as long as your threshold is less then 0.1 my guess is that something will be shown.

deepchecks/checks/distribution/__init__.py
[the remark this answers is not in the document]
noamzbr: drift_between_datasets is not in our format, and between_datasets_drift is not 100% clear, but it may be better then whole_dataset_drift so I'm willing to roll

deepchecks/checks/distribution/whole_dataset_drift.py
nirhutnik: Disagree but letting go

deepchecks/checks/distribution/__init__.py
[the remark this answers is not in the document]
nirhutnik: I don't like either but I have no idea at the moment.

nirhutnik (dismissed):

deepchecks/checks/distribution/__init__.py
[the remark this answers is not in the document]
noamzbr: I'm leaving it be for the duration of this PR

notebooks/examples/quickstart_in_5_minutes.ipynb
```diff
@@ -142,7 +142,16 @@
      "start_time": "2021-12-08T21:43:44.423627Z"
```
shir22, line 142: <p>?</p><p>(fix error / disable warning for specific versions?)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/deepchecks/deepchecks/pull/245/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

shir22, line 142: <p>How come in the output here we don't see anything?</p><p>Something weird here (on one hand no "nothing found" but on the other hand empty display, without the "traffic light" showing that the drift is minimal / non-existant)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/deepchecks/deepchecks/pull/245/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

notebooks/checks/distribution/whole_dataset_drift.ipynb
```diff
@@ -0,0 +1,480 @@
...
```
shir22, line 1: <p>Waiting to have our own datasets :)</p><p>(cause was thinking about "how can we take this code to a dataset_utils file, and then just call one function", and then understood that it is redundant.)</p><p>However we need to add some credit for uci in notebook, no?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/deepchecks/deepchecks/pull/245/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

shir22, line 1: <p>Title of the "drift bar" should has a redundant "Drift Score" in it (appears twice)</p><p> </p><p>And I think the design clarity / spacing should be improved a bit -</p><ol><li>Spacing between description and plot (maybe add new line before the drift bar?) and maybe add a title "Whole dataset total drift score" or something like that (right above the drift bar)</li><li>and also make somehow the distinction between the "total drift score" and the "main features contributing" a bit clearer? (new line? a physical line? something like that)</li></ol>
<br/>

 _Reply via <a href='https://app.reviewnb.com/deepchecks/deepchecks/pull/245/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

shir22, line 1: <p>Based on this it does look that the default condition (0.3) may be a bit high (given that this one looks clear and is only 0.32), but I guess it's better to err to the higher side</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/deepchecks/deepchecks/pull/245/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

shir22 (changes requested):
Reviewed notebooks, the main significant comment is on the quickstart's output (bug?)

shir22 (approved):

PR #247, opened 2021-12-12
added model wrapper

JKL98ISR:


JKL98ISR:
I am sincerely not sure what to think about it.
It works but should we do it?

JKL98ISR:
@ItayGabbay Just thought about something. maybe just add caching in the feature importance calculation file? like a dict of model, dataset

JKL98ISR:
> @ItayGabbay Just thought about something. maybe just add caching in the feature importance calculation file? like a dict of model, dataset

though it will remove the option of cached predictions though

yromanyshyn:
> @ItayGabbay Just thought about something. maybe just add caching in the feature importance calculation file? like a dict of model, dataset

@JKL98ISR try to use `lru_cache` decorator from `functools` module or take a look at the [cachetools](https://github.com/tkem/cachetools) package

ItayGabbay:
@JKL98ISR as @yromanyshyn mentioned, I agree that lru_cache may solve some problems here. Overall I think this is really important since the feature importance calculation is a heavy process. 
I don't mind skipping the predictions cache for now if that makes things more complicated.

noamzbr (commented):
General question - why not have the WrapperModel implement predict and predict_proba and handle it internally?

Also feature importance can be saved in an internal feature_importance_ and then no need to modify the feature importance func

matanper:
Another problem, in `TrustScoreComparison` we are predicting on a sample, which is dataframe. so you the model wrapper won't work there

JKL98ISR:
switched to lru

PR #252, opened 2021-12-13
added to datetime formatting

JKL98ISR:


deepchecks/base/dataset.py
```diff
@@ -68,7 +68,7 @@ def __init__(
         use_index: bool = False,
         index: t.Optional[Hashable] = None,
         date: t.Optional[Hashable] = None,
-        date_unit_type: t.Optional[Hashable] = None,
+        date_args: t.Optional[t.Dict] = {},
```
ItayGabbay, line 71: Please edit the docstring of the class and specify the args needed

ItayGabbay (changes requested):

deepchecks/base/dataset.py
ItayGabbay: Yup, a link would be fine. Also you changed the name of the parameter, so editing is mandatory.

ItayGabbay (dismissed):

PR #261, opened 2021-12-14
removed_check_is_fitted

JKL98ISR:


deepchecks/utils/validation.py
```diff
@@ -64,16 +65,6 @@ def validate_model(dataset: 'base.Dataset', model: t.Any):
             'But function received empty dataset.'
         ))
 
-    try:
```
noamzbr, line 67: No need to delete that. Note that if feature_names_in_ doesn't exist, then it's pass and no harm it's done.
It's only a way to get a more specific error message.

noamzbr (changes requested):

noamzbr (approved):

PR #265, opened 2021-12-14
Relax label constraints

noamzbr:
Remove the requirement for consecutive labels


deepchecks/checks/distribution/trust_score_comparison.py
```diff
@@ -130,15 +130,23 @@ def run(self, train_dataset, test_dataset, model=None) -> CheckResult:
             max_num_categories=self.max_number_categories
         )
```
nirhutnik, line 132:
```suggestion
# Trust Score model expects labels to be consecutive integers from 0 to n-1, so we transform our label to this format:
```

nirhutnik (approved):

nirhutnik (approved):

PR #266, opened 2021-12-15
fix tqdm problems with kaggle

JKL98ISR:


deepchecks/base/display_suite.py
```diff
@@ -30,12 +31,7 @@ class ProgressBar:
     def __init__(self, name, length):
         """Initialize progress bar."""
         shared_args = {'total': length, 'desc': name, 'unit': ' Check', 'leave': False, 'file': sys.stdout}
```
matanper, line 33: you don't need `shared_args` just put them in the tqdm init

matanper (approved):

PR #268, opened 2021-12-15
Add plotly support to check display

matanper:


requirements.txt
```diff
@@ -3,11 +3,11 @@ numpy>=1.19
...
-seaborn>=0.11.2
-matplotlib>=3.3.3
 ipywidgets>=7.6.5
 tqdm>=4.62.3
 category-encoders>=2.3.0
 scipy>=1.5.0
 typing_extensions>=4.0.1
-dataclasses>=0.6; python_version < '3.7'
\ No newline at end of file
+dataclasses>=0.6; python_version < '3.7'
```
shir22, line 11: Since when do we have this python version limitation?

matanper: I have no idea, I didn't add it. I think @benisraeldan had problems with the dataclass, so maybe he knows

ItayGabbay: It's not a limit. It's just we need this packet when the python version is below 3.7. Above 3.7 it's integrated in python itself.

ItayGabbay (approved):

PR #269, opened 2021-12-15
Added pipeline support for boosting overfit

JKL98ISR:


JKL98ISR:
we need to add tests for pipeline usage
we also never tested the usage in the features before

matanper (approved):

ItayGabbay:
Yup, I agree. This can't be merged without unit tests of the pipeline usecase.

ItayGabbay (approved):

PR #271, opened 2021-12-16
Update label docs and add validation

matanper:


deepchecks/base/dataset.py
```diff
@@ -115,7 +113,16 @@ def __init__(
         if label is not None:
             if label.shape[0] != self._data.shape[0]:
                 raise DeepchecksValueError('Number of samples of label and data must be equal')
-            label_name = label_name or 'target'
+            # Make tests to prevent overriding user column
+            if label_name is None:
+                label_name = 'target'
+                if label_name in self._data.columns:
+                    raise DeepchecksValueError(f'Data have column with name "{label_name}", use label_name parameter'
```
shir22, line 120: Data has

shir22: Why is thir validation nested inside the "if label_name=None"?

shir22 (changes requested):

deepchecks/base/dataset.py
matanper: To write another message to the user based on he if passed `label_name` or not (if he didn't pass label_name but there is still a clash with the default name, `target`)

shir22 (approved):

PR #276, opened 2021-12-20
small documentation improvements

yromanyshyn:


ItayGabbay (approved):

docs/source/conf.py
```diff
@@ -0,0 +1,530 @@
...
+    result = f"* :doc:`API Reference - {notebook_name} <{apipath}>`"
+    return result
+
+
+
+# TODO: Not sure if we need this
+
+# class AutoAutoSummary(Autosummary):
+
+#     option_spec = {
```
benisraeldan, line 467: you're right we don't need this, you can remove

docs/source/conf.py
```diff
@@ -0,0 +1,530 @@
...
+
+
+CURRENT_DIR = pathlib.Path(__file__).parent
+PROJECT_DIR = CURRENT_DIR.parent.parent
+
+
+sys.path.insert(0, str(PROJECT_DIR.absolute()))
+
+
+import deepchecks.version
```
benisraeldan, line 30: I can't import this locally, can you remove it from here, or add it to the package?

makefile
```diff
@@ -17,14 +17,24 @@ REQUIRE = requirements.txt
...
+>>>>>>> main
 repo = pypi
 
 # System Envs
 BIN := $(ENV)/bin
+<<<<<<< HEAD
+pythonpath := PYTHONPATH=.
+=======
 pythonpath = PYTHONPATH=.
+>>>>>>> main
```
benisraeldan: fix merge conflicts

yromanyshyn: already fixed)

makefile
```diff
@@ -17,14 +17,24 @@ REQUIRE = requirements.txt
 ext_py := $(shell which python3 || which python)
 
 # Override by putting in commandline python=XXX when needed.
+<<<<<<< HEAD
+python = $(shell echo ${ext_py} | rev | cut -d '/' -f 1 | rev)
+TESTDIR = $(shell realpath tests)
+ENV = $(shell realpath venv)
```
benisraeldan, line 22: this doesn't work locally on my machine

docs/source/conf.py
yromanyshyn: do you mean that you get ImportError when trying to import `deepchecks.version`?

makefile
yromanyshyn: do you use windows? (I pushed fix for this)

docs/source/conf.py
yromanyshyn: removed

makefile
benisraeldan: nope, I'm on mac

benisraeldan:
right now we have duplicate notebooks, is there a way to reference them from outside the source directory? or should we move them to the source and keep them update and tested there?

yromanyshyn:
>right now we have duplicate notebooks, is there a way to reference them from outside the source directory? or should we move them to the source and keep them update and tested there?

@benisraeldan 

It looks like it is not possible to reference documents outside of the source dir in the rst/ at least I was not able to achieve this

Think we need to move notebooks to the docs/source folder

makefile
yromanyshyn: weird/ Did I understand you right that there is no `realpath` cmd in mac env?

yromanyshyn: if so, I think this cmd will install it - `brew install coreutils`

yromanyshyn:
> >right now we have duplicate notebooks, is there a way to reference them from outside the source directory? or should we move them to the source and keep them update and tested there?
> 
> @benisraeldan
> 
> It looks like it is not possible to reference documents outside of the source dir in the rst/ at least I was not able to achieve this
> 
> Think we need to move notebooks to the docs/source folder

@ItayGabbay any thoughts regarding that?

docs/source/conf.py
benisraeldan: yup

PR #279, opened 2021-12-20
Fixes to dataset creation logic

noamzbr:
Fixes to the following issues:

1. Enable using the dataset index as date
2. Problem with merge of label column (can create nans)
3. Suite generator names are similar to suites and under same hierarchy

Point 1 wasn't done yet, but the outlines of how I want to do it are in the change to the dataset init docs. Please have a look there and tell me what you think.


deepchecks/base/dataset.py
```diff
@@ -88,13 +89,20 @@ def __init__(
             label_name (Optional[Hashable]):
                 If `label` is given, then this name is used as the column name for the labels.
                 If `label` is none, then looks for this name in the data dataframe.
-            use_default_index (bool, default False):
-                Whether to use the dataframe index as the index column, for index related checks. can't be used
-                together with `index_name`
             index_name (Optional[Hashable]):
-                Name of the index column in the DataFrame. can't be used together with `use_default_index`
+                Name of the index column in the dataframe. If create_index_from_dataframe_index is True and index_name
```
noamzbr, line 93: here is the place to look @shir22 @ItayGabbay

deepchecks/suites/__init__.py
```diff
@@ -9,7 +9,7 @@
 # ----------------------------------------------------------------------------
 #
 """Module contains all prebuilt suites."""
-from .integrity_suite import *
-from .methodology_suite import *
-from .performance_suite import *
-from .overall_suite import *
+from .integrity import *
```
shir22, line 12: This may be a bit confusing given that (for some) these are also the submodules of the check, no?
(e.g. Integrity checks)
Mabye should be integrity_suites and so forth?

deepchecks/base/dataset.py
shir22: In general this looks good.
A few things that I found not inuitive:
1. The wording "create index" - I think it may confuse the user - if they're passing the index column and telling you "that is the index", what exactly are you "creating"?
2. Somewhat similar but different for "date", what does it mean to "create date".
3. I think in general our approach of "we can have a single index column" and "we can have a single date column", and each are treated differently isn't very clear to the user (what we use each for, and why can there be only a single of each), but I think this is something we should iterate with feedback to improve the clarity and usability of this

deepchecks/base/dataset.py
```diff
@@ -124,15 +134,16 @@ def __init__(
                 if label_name in self._data.columns:
                     raise DeepchecksValueError('Can\'t pass label with label_name that exists in the data. change '
                                                'the label_name parameter')
-            self._data[label_name] = label
+            # Cast to numpy to promise no index collisions
```
shir22, line 137: I don't like this... This is something that the user should explicitly, and make sure they understand what they're doing.

I think we should give the user an indicative error and tell them that the indices of their labels and samples don't match. It might well be that they have a problem (e.g. maybe they shuffled them differently or sampled them differently?)
Personally, I had a bug that had to do with something like that in a past DS project, and it was very annoying to find it, and it feels almost impossible to find it when this is implicitly done.

deepchecks/base/dataset.py
```diff
@@ -124,15 +134,16 @@ def __init__(
...
+            self._data[label_name] = np.array(label)
 
-        if use_default_index is True and index_name is not None:
-            raise DeepchecksValueError('parameter use_default_index cannot be True if index is given')
+        if create_index_from_dataframe_index is True and index_name is not None:
```
shir22, line 140: Guessing here you just renamed the previous code as part of the refactor cause it's not yet the behavior describe above :)

deepchecks/base/dataset.py
noamzbr: 1. Maybe "set_Index.."?
2. same
3. Don't have an idea for an alternative simple solution to start from...

deepchecks/suites/__init__.py
noamzbr: It's the exact same names as with the checks, except overall that exists here. But this will change soon when we change the default suites we offer. IMO it's clear that if it's under suites then it's a suite, and also we can't have it be "integrity_suites" because that's confusing with the actual suite that is inside the module

deepchecks/base/dataset.py
noamzbr: OK, will raise a dedicated error instead if label is a series or dataframe

deepchecks/base/dataset.py
noamzbr: yep(:

deepchecks/suites/__init__.py
shir22: yes understood that double-naming is problematic, that's why offered integrity_suites
(and then inside it there will be integrity_suite)
but get your case, if it's inside the suites, then seems fine (overlap will be partial, according to the suite types...)
coming to think about it - why don't we just have a single suites file? (I know it used to be like that, and then changed because we had multiple suites of each type, but will this be the case now?)

deepchecks/base/dataset.py
noamzbr: we can change the whole concept of "date" to "datetime", you think it's better?

deepchecks/suites/__init__.py
noamzbr: so I'm keeping this like that, and when we get to the reorg then it'll all be in one file

ItayGabbay (approved):
Overall and in continuing our discussion I think it's too complicated, but approving for now. 
Let's discuss it in the next sprint.

deepchecks/base/dataset.py
```diff
@@ -424,32 +481,52 @@ def index_col(self) -> t.Optional[pd.Series]:
...
+        """If datetime column exists, return its name.
 
         Returns:
-           (str) date column name
+           (str) datetime name
         """
-        return self._date_name
+        return self._datetime_name
+
+    def get_datetime_column_from_index(self, datetime_name):
```
matanper, line 507: like yurii did in another PR, if this method doesn't use self, and called just once in init, than move it outside or make it classs method

deepchecks/base/dataset.py
```diff
@@ -512,6 +589,12 @@ def columns_info(self) -> t.Dict[Hashable, str]:
             columns[column] = value
         return columns
 
+    def check_compatible_labels(self):
```
matanper, line 592: remove this

deepchecks/base/dataset.py
```diff
@@ -35,108 +35,150 @@ class Dataset:
...
-                error_message += ' If you attempted to use the dataframe index, set use_default_index to True instead.'
-            raise DeepchecksValueError(error_message)
-
-        if date_name is not None and date_name not in self._data.columns:
-            raise DeepchecksValueError(f'date column {date_name} not found in dataset columns')
+            # If passed label is a pandas object, check that indexes match, else set column as is with provided values
+            if isinstance(label, (pd.Series, pd.DataFrame)):
+                pd.testing.assert_index_equal(self._data.index, label.index)
+                self._data[label_name] = label
+            else:
```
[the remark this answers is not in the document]
noamzbr: numpy also has shape

deepchecks/base/dataset.py
noamzbr: does use init

deepchecks/base/dataset.py
noamzbr: removed by @yromanyshyn in another PR

deepchecks/base/dataset.py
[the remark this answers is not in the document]
matanper: But before calling shape you need to validate the instance is pandas/numpy and not list for example

deepchecks/base/dataset.py
matanper: English?

PR #282, opened 2021-12-20
Mlc 300

ItayGabbay:


noamzbr:
Why? In most repos notebooks sit either under "notebooks" or more commonly "examples"

PR #284, opened 2021-12-20
improve wording

benisraeldan:
Class Performance Imbalance - Bad condition name and description (too much redundant text)


deepchecks/checks/performance/class_performance_imbalance.py
```diff
@@ -34,11 +34,11 @@
...
 
 
-CP = t.TypeVar('CP', bound='ClassPerformanceImbalance')
+CP = t.TypeVar('CP', bound='ClassPerformance')
 
 
-class ClassPerformanceImbalance(SingleDatasetBaseCheck):
-    """Visualize class imbalance by displaying the difference between class score values.
+class ClassPerformance(SingleDatasetBaseCheck):
+    """Visualize class performance compared to other classes.
```
noamzbr, line 41: mmm still not really clear, maybe write 
Visualize performance per class - aiding detection of class imbalance

But don't put a lot of effort into it because next sprint there is a task merging this with performance report and train-test overfit

noamzbr (approved):

PR #289, opened 2021-12-20
[Refactoring] Moved `is_categorical` and `infer_categorical_features` to the utils package

yromanyshyn:
list of changes:
- moved `is_categorical` function from Dataset to the utils package
- moved `infere_categorical_columns` function from Dataset to the utils package
- renamed `Dataset.filter_columns_with_validation` into `Dataset.select`

### DIFF
--- a/deepchecks/base/dataset.py
+++ b/deepchecks/base/dataset.py
@@ -16,9 +16,9 @@ import logging
 
 import numpy as np
 import pandas as pd
-from pandas.core.dtypes.common import is_float_dtype
 
-from deepchecks.utils.dataframes import filter_columns_with_validation
+from deepchecks.utils.dataframes import select_from_dataframe
+from deepchecks.utils.features import is_categorical, infer_categorical_features
 from deepchecks.utils.strings import is_string_column
 from deepchecks.utils.typing import Hashable
 from deepchecks.errors import DeepchecksValueError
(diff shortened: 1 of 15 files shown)


deepchecks/base/dataset.py
```diff
@@ -349,43 +355,48 @@ def n_samples(self) -> int:
...
         return self.data.shape[0]
 
-    def __len__(self):
+    def __len__(self) -> int:
         """Return number of samples in the member dataframe."""
         return self.n_samples
 
-    def infer_categorical_features(self) -> t.List[Hashable]:
+    @staticmethod
+    def _infer_categorical_features(
```
matanper, line 363: why it's better as static method if it's need so many values from self?

yromanyshyn: I think the next reasons should be followed when deciding if declaring a function as a method instance:
- it represents/modifies object state during its life-cycle
- it is a private method that works with object inner state

`infer_categorical_features` is definitely just a utility function 
that is called only once during dataset initialization, also, 
it neither represents object state nor modifies it

passing arguments explicitly makes it more obvious what does this function do

also, do not you consider it strange that we do not pass arguments to it 
(we use self for this) but we return value from it which is then assigned to the self/

matanper (approved):

PR #296, opened 2021-12-22
[MLC-379] fixed suite result display 

yromanyshyn:


deepchecks/base/display_suite.py
```diff
@@ -113,13 +135,13 @@ def display_suite_result(suite_name: str, results: List[Union[CheckResult, Check
...
-    first_3 = []
-    i = 0
-    while len(first_3) < 3 and i < len(results):
-        curr = results[i]
-        curr_name = curr.check.name()
-        if curr_name not in first_3:
-            first_3.append(curr_name)
-        i += 1
-    return ', '.join(first_3)
+# def get_first_3(results: List[Union[CheckResult, CheckFailure]]):
```
shir22, line 138: Suite may have several similar checks, so this function is relevant for returning unique names., so the list of names will look funny)
If len of unique names is less than 3 than the other wording you proposed should be used

deepchecks/base/display_suite.py
```diff
@@ -70,24 +75,41 @@ def display_suite_result(suite_name: str, results: List[Union[CheckResult, Check
             msg = result.exception.__class__.__name__ + ': ' + str(result.exception)
             name = result.check.name()
             others_table.append([name, msg, 1])
+        else:
```
shir22, line 78: why fail all suite display if only single (or few) result(s) is problematic

deepchecks/base/display_suite.py
```diff
@@ -70,24 +75,41 @@ def display_suite_result(suite_name: str, results: List[Union[CheckResult, Check
...
-    Suites, checks and conditions can all be modified (see tutorial [link]).</p>
-    {bold_hr}<h2>Conditions Summary</h2>
-    """
-    display_html(html, raw=True)
+
+    check_names = list(set(it.check.name() for it in results))
+    prologue = (
+        f"The suite is composed of various checks such as: {', '.join(check_names[:3])}, etc..."
+        if len(check_names) > 3
+        else f"The suite is composed of next checks: {', '.join(check_names)}."
```
shir22, line 98:
```suggestion
        else f"The suite is composed of the following checks: {', '.join(check_names)}."
```

deepchecks/base/display_suite.py
```diff
@@ -70,24 +75,41 @@ def display_suite_result(suite_name: str, results: List[Union[CheckResult, Check
...
-    Each check may contain conditions (which results in {icons}), as well as other outputs such as plots or tables.<br>
-    Suites, checks and conditions can all be modified (see tutorial [link]).</p>
-    {bold_hr}<h2>Conditions Summary</h2>
-    """
-    display_html(html, raw=True)
+
+    check_names = list(set(it.check.name() for it in results))
+    prologue = (
+        f"The suite is composed of various checks such as: {', '.join(check_names[:3])}, etc..."
+        if len(check_names) > 3
```
shir22, line 97: see comment bellow about unique names (for both lines)

shir22 (changes requested):

deepchecks/base/display_suite.py
yromanyshyn: > Suite may have several similar checks,

yup, I know, but set drops all duplicates

```
check_names = list(set(it.check.name() for it in results))
```

deepchecks/base/display_suite.py
yromanyshyn: this is an internal function, if it receives something unexpected that means we have a bug in the code. WIthout `else` we just ignoring those bugs

deepchecks/base/display_suite.py
```diff
@@ -54,6 +55,10 @@ def get_display_exists_icon(exists: bool):
 
 def display_suite_result(suite_name: str, results: List[Union[CheckResult, CheckFailure]]):
     """Display results of suite in IPython."""
+    if len(results) == 0:
```
matanper, line 58: I think this should be impossible, maybe we need in a suite creation to ensure there is at least 1 check

deepchecks/base/display_suite.py
matanper: we have function `wrap_run` where we validate each return value of a check. I'll add over there assertion the return type is CheckResult, and than we won't have to worry about that

deepchecks/base/display_suite.py
matanper: On second thought it is possible. you can create a suite with only train-test checks, and than run suite with single dataset so nothing will run

deepchecks/base/display_suite.py
yromanyshyn: ok, but it does not change anything/
writing `if-elif` statement without `else` is a bad practice
you never know in which part of the package your function will be used and what will be passed to it.

in the end, it is just one line of the code/

deepchecks/base/display_suite.py
yromanyshyn: >I think this should be impossible

it is possible, currently, we do not do any validation for this, not in init method not in run method

>maybe we need in a suite creation to ensure there is at least 1 check

we have the `add` method within the Suite which assumes that the suite is a mutable type and that we can add checks to it after its creation

matanper:
@yromanyshyn added a few changes, please review and than merge

matanper (approved):

yromanyshyn:
> @yromanyshyn added a few changes, please review and than merge

@matanper looks good, thanks, but we also need @shir22 approval (merge is blocked without her approval)

yromanyshyn:
oh, actually never mind/ I actually can merge

PR #297, opened 2021-12-22
bug in dataset class?

yromanyshyn:
I am confused/ Why do we need this method?
GitLens is saying that I am the last who modified this method, I probably could accidentally remove its body/

but I see your tests @noamzbr, and guess that this method is actually not needed/?


noamzbr (commented):
Hi @yromanyshyn, you're right, we don't need check_compatible_labels anymore. I've removed the validations inside it in a PR yesterday.

yromanyshyn:
@noamzbr ok, I removed this method from dataset module, please approve)

noamzbr (approved):

PR #315, opened 2021-12-24
[MLC-388] added limit on dataset features number for features importance calculation

yromanyshyn:
short list of changes:

#### 1)
   
added `deepchecks.utils.features._NUMBER_OF_FEATURES_LIMIT = 100` to limit features importance calculation only for 
the datasets that contain a fewer number of features than the specified limit

#### 2)
   
added next functions:
    + `deepchecks.utils.features.set_number_of_features_limit`
    + `deepchecks.utils.features.get_number_of_features_limit`
  to allow user to change `deepchecks.utils.features._NUMBER_OF_FEATURES_LIMIT`

after some simple experimentations, it become obvious that the `calculate_feature_importance` execution time is too dependant upon the type of the model (and probably upon of bunch of other variables) so it is not the best idea to leave the user without the possibility to redefine that limit.

[collected execution time statistics ](https://drive.google.com/file/d/1GaJ0fvS6uO2LhJyEYoD0DvE0heU0AszY/view?usp=sharing)

#### 3)

after the above changes, some modifications were also needed for the next checks
    - `WholeDatasetDrift`
    - `SegmentPerformance`

#### 4)

refactored `calculate_feature_importance` function


noamzbr (approved):
approved pending setting the default to 200

ItayGabbay (approved):

PR #325, opened 2021-12-27
fix calibration score check

benisraeldan:
Reproduced with dataset where the labels were strings and not numbered classes.

### DIFF
--- a/deepchecks/checks/performance/calibration_score.py
+++ b/deepchecks/checks/performance/calibration_score.py
@@ -49,15 +49,15 @@ class CalibrationScore(SingleDatasetBaseCheck):
         y_pred = model.predict_proba(ds_x)
 
         briers_scores = {}
-        unique_labels = dataset.label_col.unique()
+        unique_labels = model.classes_
 
         if len(unique_labels) == 2:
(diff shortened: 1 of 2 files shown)


deepchecks/checks/performance/calibration_score.py
```diff
@@ -49,15 +49,15 @@ def _calibration_score(self, dataset: Dataset, model):
         y_pred = model.predict_proba(ds_x)
 
         briers_scores = {}
-        unique_labels = dataset.label_col.unique()
+        unique_labels = model.classes_
```
ItayGabbay, line 52: What is the difference between classes_ and label_col.unique()?

ItayGabbay (commented):
As I understand, the problem was that the labels are strings and the predictions are just arrays. So you used classes_ in order to get the correct "order" of labels?
Correct me if I'm wrong

deepchecks/checks/performance/calibration_score.py
benisraeldan: exactly what you said in the comment

ItayGabbay (approved):

PR #335, opened 2021-12-28
* Adding the datasets module

ItayGabbay:
* Modifying some docs


shir22 (changes requested):
Looks great!
Main comments:
1. State more clearly the typical "target / purpose" of the dataset at the beginning
2. For the load_data's docstring's first part - Seems that rst syntax will indeed be useful here, good time to add this extension to our sphinx

noamzbr (commented):
1. didn't we want to nest avocado under dataset.regression?
2. Making the sanity check that the dataset is under a permissive license. Is it required to post that license here?

ItayGabbay:
@shir22 @noamzbr done

ItayGabbay:
image

noamzbr (approved):

deepchecks/datasets/regression/avocado.py
```diff
@@ -0,0 +1,131 @@
...
+             - real, string
+           * - Targets
+             - real 0.44 - 3.25
+
+    Description:
+        .. list-table:: Dataset Description
+           :widths: 50 50
+           :header-rows: 1
+
+           * - Column name
```
shir22, line 58: Had a look at kaggle's description for this dataset...
I think we should add also the description of the columns that have one:
image
(including an inline explanation for "what is PLU")

shir22 (changes requested):
added a final comment

btw had a look at the database license -
https://opendatacommons.org/licenses/odbl/1-0/
it and all of its derivatives need to be licensed the same and have a copy of the license

deepchecks/datasets/regression/avocado.py
ItayGabbay: image

deepchecks/datasets/regression/avocado.py
```diff
@@ -0,0 +1,150 @@
...
+
+_MODEL_URL = 'https://figshare.com/ndownloader/files/32393723'
+_FULL_DATA_URL = 'https://figshare.com/ndownloader/files/32393729'
+_TRAIN_DATA_URL = 'https://figshare.com/ndownloader/files/32393732'
+_TEST_DATA_URL = 'https://figshare.com/ndownloader/files/32393726'
+_target = 'AveragePrice'
+_CAT_FEATURES = ['region', 'type']
+
+
+def load_data(data_format: str = 'Dataset', as_train_test: bool = True) -> \
```
matanper, line 28: a. as I wrote in the slack, why do we need all those options?
b. I'd prefer to have boolean instead of data_format (like sklearn, `as_frame`)

ItayGabbay: About a. - we want to let the user explore the package but also give him the opportunity to experiment a bit with the checks logic. He can do that by manipulating existing datasets and compare the suite results, for example. + It really isn't so hard to implement it...

About b. - This might be an option. But don't you think that it's not so clear what will be the return type if I specify False? I also think that in sklearn it's not clear... that's why I prefer they mention it explicitly

matanper: a. exactly for the things you wrote he needs train + test datasets. no reason to give him dataframe or single dataset

shir22 (approved):

PR #337, opened 2021-12-28
infer task type

benisraeldan:
infer task type from dataset


deepchecks/base/dataset.py
```diff
@@ -420,6 +441,20 @@ def train_test_split(self,
...
 
+    @staticmethod
+    def _infer_label_type(
+            label_col: pd.Series,
+            max_categorical_ratio: float,
+            max_categories: int,
+            max_float_categories: int
+    ):
+        if not is_numeric_dtype(label_col):
+            return 'classification label'
```
ItayGabbay, line 452: I would remove the "label" word from the type.

deepchecks/base/dataset.py
```diff
@@ -420,6 +441,20 @@ def train_test_split(self,
...
+        elif is_categorical(label_col, max_categorical_ratio, max_categories, max_float_categories):
+            return 'classification label'
+        else:
+            return 'regression label'
```
ItayGabbay, line 456: I just hope it's a good heuristic....

deepchecks/base/dataset.py
benisraeldan: @nirhutnik - having it just classification/regression is weird for a label type

ItayGabbay: at least let's replace the space with _

deepchecks/base/dataset.py
noamzbr: I'm also not sure about it, but it's better than using something like https://scikit-learn.org/stable/modules/generated/sklearn.utils.multiclass.type_of_target.html#sklearn.utils.multiclass.type_of_target - this will for e.g. infer a "test score" that is ints from 0 to 100 as "classification" although it's clearly regression. 

We're open to suggestions

ItayGabbay (approved):

PR #341, opened 2021-12-29
Simple model multi metric

matanper:


deepchecks/checks/distribution/preprocessing.py
```diff
@@ -20,70 +20,73 @@
...
-    Returns:
-        (DataFrame, DataFrame): returns both datasets transformed.
     """
-    x_baseline = baseline_features.copy()
-    x_test = test_features.copy()
-    non_categorical_columns = list(set(test_features.columns) - set(categorical_columns))
 
-    # Impute all-nan cols to all-zero:
-    def impute_whole_series_to_zero(s: pd.Series):
+    def __init__(self, baseline_features: pd.DataFrame, categorical_columns: List[Hashable], max_num_categories: int):
```
noamzbr, line 39: maybe better to have this follow the way scikit learn does that. Look for e.g. at https://github.com/scikit-learn/scikit-learn/blob/7e1e6d09bcc2eaeba98f7e737aac2ac782f0e5f1/sklearn/preprocessing/_data.py#L632 and what it inherits from.
Long story short - categorical_columns: List[Hashable], max_num_categories defined in init, fitting done in fit(), transforming in transform(), fit_transform() does both one after the other.

deepchecks/checks/performance/simple_model_comparison.py
```diff
@@ -127,90 +180,89 @@ def _find_score(self, train_ds: Dataset, test_ds: Dataset, task_type: ModelType,
...
+        predictions = self.predict(a)
+
+        def create_proba(x):
+            proba = np.zeros(classes_num)
+            proba[x] = 1
+            return proba
+        return np.apply_along_axis(create_proba, axis=1, arr=predictions)
+
+
+class ScaledModel:
```
noamzbr, line 257: way to do this is with [pipelines](https://scikit-learn.org/stable/modules/generated/sklearn.pipeline.Pipeline.html)
Have a look at other usage in the package

deepchecks/checks/performance/simple_model_comparison.py
```diff
@@ -127,90 +180,89 @@ def _find_score(self, train_ds: Dataset, test_ds: Dataset, task_type: ModelType,
...
-        fig.update_layout(width=600, height=500)
-        fig.update_yaxes(title=score_name)
-
-        return CheckResult({'given_model_score': pred_score,
-                            'simple_model_score': simple_score,
-                            'ratio': ratio},
-                           display=[text, fig])
-
-    def add_condition_ratio_not_less_than(self, min_allowed_ratio: float = 1.1):
+    def add_condition_ratio_not_less_than(self, min_allowed_ratio: float = 1.1, classes: List[Hashable] = None):
```
noamzbr, line 192: Let me know what you think, but it may be better to set this by default only on a single metric (e,g. f1)

deepchecks/checks/performance/simple_model_comparison.py
```diff
@@ -127,90 +180,89 @@ def _find_score(self, train_ds: Dataset, test_ds: Dataset, task_type: ModelType,
...
+            for df in metrics_dfs:
+                origin_score = df[df['Type'] == 'Origin']['Value'].iloc[0]
+                simple_score = df[df['Type'] == 'Simple']['Value'].iloc[0]
+                ratio = get_scores_ratio(simple_score, origin_score, max_ratio)
+                if ratio < min_allowed_ratio:
+                    not_passing_metrics.update(df['Metric'].tolist())
+
+            if not_passing_metrics:
+                not_passing_metrics = sorted(not_passing_metrics)
+                msg = f'Metrics with scores ratio lower than threshold: {", ".join(not_passing_metrics)}'
```
noamzbr, line 227: Think we should report not passing classes.

deepchecks/checks/performance/simple_model_comparison.py
matanper: it's not possible to set a default since you don't know the model type

deepchecks/checks/performance/simple_model_comparison.py
matanper: I think it will be too long message

deepchecks/checks/distribution/preprocessing.py
```diff
@@ -9,81 +9,94 @@
...
+__all__ = ['ScaledNumerics']
 
 
-def preprocess_dataset_to_scaled_numerics(
-    baseline_features: pd.DataFrame,
-    test_features: pd.DataFrame,
-    categorical_columns: List[Hashable],
-    max_num_categories: int
-) -> Tuple[pd.DataFrame, pd.DataFrame]:
+class ScaledNumerics(TransformerMixin):
```
noamzbr, line 28: also BaseEstimator, as it implements fit

noamzbr (approved):

PR #344, opened 2021-12-29
add minimum outlier difference

benisraeldan:
outlier length sometimes shows outliers that are too close to the normal range, we want to filter out these
e.g.:

default | 2 - 2 | 3 - 3 | 222
-- | -- | -- | --


noamzbr (changes requested):
By default, this check should not run on categorical columns. It should run only on string columns that have many unique values (that we assume contain text of some kind).
Please add this behavior and add a flag to control it.

noamzbr (approved):
Approved, pending setting default min_unique_value_ratio to default value

PR #368, opened 2021-12-30
drop na_action from column.map, we already dropna

benisraeldan:


benisraeldan:
na_action is not implemented in some scenarios in df.Series.map

matanper (approved):

PR #379, opened 2022-01-01
Mlc 456 unused feature bug

nirhutnik:
Found the bug(s) - 
naive_encoder was not returning deterministic output (numeric_features list was not keeping its sorting).

Also, the variance series was using the feature_importance indexing, which was wrong as it returns in a different column order from the naive_encoder.


noamzbr (approved):
Thanks for fixing my bugs(:

PR #380, opened 2022-01-01
Label + prediction distribution

nirhutnik:


deepchecks/checks/distribution/train_test_label_drift.py
```diff
@@ -0,0 +1,230 @@
...
+# Copyright (C) 2021 Deepchecks (https://www.deepchecks.com)
+#
+# This file is part of Deepchecks.
+# Deepchecks is distributed under the terms of the GNU Affero General
+# Public License (version 3 or later).
+# You should have received a copy of the GNU Affero General Public License
+# along with Deepchecks.  If not, see <http://www.gnu.org/licenses/>.
+# ----------------------------------------------------------------------------
+#
+"""Module contains Train Test Drift check."""
```
noamzbr, line 11:
```suggestion
"""Module contains Train Test label Drift check."""
```

deepchecks/checks/distribution/train_test_label_drift.py
```diff
@@ -0,0 +1,230 @@
...
+                display: label distribution graph, comparing the train and test distributions.
+        """
+        train_dataset = Dataset.validate_dataset(train_dataset)
+        test_dataset = Dataset.validate_dataset(test_dataset)
+
+        drift_score, method, display = self._calc_drift_per_column(
+            train_column=train_dataset.label_col,
+            test_column=test_dataset.label_col,
+            column_name=train_dataset.label_name,
+            column_type='categorical' if train_dataset.label_type == 'classification_label' else 'numerical',
```
noamzbr, line 95: If a model was passed, we should use it to determine the type

deepchecks/checks/distribution/train_test_label_drift.py
```diff
@@ -0,0 +1,230 @@
...
+            The Drift score is a measure for the difference between two distributions, in this check - the test
+            and train distributions.<br> The check shows the drift score and distributions for the label.
+        </span>"""
+
+        displays = [headnote, display]
+        values_dict = {'Drift score': drift_score, 'Method': method}
+
+        return CheckResult(value=values_dict, display=displays, header='Train Test Label Drift')
+
+    def _calc_drift_per_column(self, train_column: pd.Series, test_column: pd.Series, column_name: Hashable,
```
noamzbr, line 108: Code duplication with regular drift check. Should be util or base class.

deepchecks/checks/distribution/train_test_label_drift.py
```diff
@@ -0,0 +1,230 @@
...
+                return_str = f'Label has PSI over {max_allowed_psi_score} - Drift score is {drift_score:.2f}'
+                return ConditionResult(False, return_str)
+            elif method == "Earth Mover's Distance" and has_failed:
+                return_str = f'Label has Earth Mover\'s Distance over {max_allowed_earth_movers_score} - ' \
+                             f'Drift score is {drift_score:.2f}'
+                return ConditionResult(False, return_str)
+
+            return ConditionResult(True)
+
+        return self.add_condition(f'PSI and Earth Mover\'s Distance cannot be greater than {max_allowed_psi_score} or '
```
noamzbr, line 228:
```suggestion
        return self.add_condition(f'PSI and Earth Mover\'s Distance for label drift cannot be greater than {max_allowed_psi_score} or '
```

deepchecks/checks/distribution/train_test_label_drift.py
```diff
@@ -0,0 +1,230 @@
...
+from deepchecks.utils.typing import Hashable
+import plotly.graph_objects as go
+from plotly.subplots import make_subplots
+
+__all__ = ['TrainTestLabelDrift']
+
+
+class TrainTestLabelDrift(TrainTestBaseCheck):
+    """
+    Calculate label drift between train dataset and test dataset, using statistical measures.
```
noamzbr, line 29: Somehow this line does not appear in the notebook, maybe rerun and re-commit?

noamzbr (changes requested):
minor comments

deepchecks/checks/distribution/train_test_label_drift.py
nirhutnik: Disagree. This is purely on the data and the user can better control the type if inferred from dataset / directly given

deepchecks/checks/distribution/train_test_label_drift.py
nirhutnik: Actually, not completely duplicated as the feature importance thingy is in there. 
Thought about consolidating and splitting to smaller functions that can be reused, but seems too complicated for now. I suggest we open a ticket for it for later.

deepchecks/checks/distribution/train_test_label_drift.py
nirhutnik: Weird, I see it. Rerunning anyway

deepchecks/checks/distribution/train_test_label_drift.py
noamzbr: mm well can concede, if only on the grounds of this not running with a model most of the time

deepchecks/checks/distribution/train_test_label_drift.py
noamzbr: yeah you can open an issue for that

noamzbr (approved):

PR #387, opened 2022-01-02
Add explanatation about scorers 

matanper:


noamzbr (commented):
Add it also in performance report

noamzbr (approved):

PR #397, opened 2022-01-02
assume task type from model or label

### LINKED ISSUE #390: BUG:New Label train test in regression task
When running the full suite on a regression task, I encouner 2 differnet bugs:

1. This check is appearing twice
2. This check is appearing (In regression i don't expect it to be there)
image

benisraeldan:
- assume new task type from model without using is_regression attribute from sklearn
 - remove NewLabelTrainTest duplication from full_suite
  
fix #390


deepchecks/checks/integrity/label_ambiguity.py
```diff
@@ -60,8 +61,9 @@ def run(self, dataset: Dataset, model=None) -> CheckResult:
         dataset: Dataset = Dataset.validate_dataset(dataset)
         dataset = dataset.select(self.columns, self.ignore_columns)
 
-        if (model and hasattr(model, 'is_regression') and model.is_regression()) or \
-                (dataset.label_type == 'regression_label'):
+        if model:
```
ItayGabbay, line 64: If you use the model then add it to the docstring

ItayGabbay (approved):

PR #415, opened 2022-01-02
Roc plotly

JKL98ISR:


ItayGabbay (approved):

deepchecks/checks/performance/roc_report.py
```diff
@@ -57,45 +57,55 @@ def _roc_report(self, dataset: Dataset, model):
...
         multi_y = (np.array(ds_y)[:, None] == np.unique(ds_y)).astype(int)
-        n_classes = len(dataset.classes)
         y_pred_prob = model.predict_proba(ds_x)
 
         fpr = {}
         tpr = {}
         roc_auc = {}
-        for i in range(n_classes):
+        for i, _ in enumerate(dataset_classes):
             if i in self.excluded_classes:
```
matanper, line 68: I think we need to change this to work directly on the class, not the class index

JKL98ISR: why?

JKL98ISR: I'm using an array

matanper: I'm not talking about the array, if the labels are 'a', 'b', 'c' the user will exclude [a,b] and not [1,2]

JKL98ISR: oh yeah forgot about the excluded classes

JKL98ISR: changed it so it'll skip by name, I still need the relative index because of the prediction series

PR #436, opened 2022-01-03
Add to suite output the checks that haven't run because of missing parameters

### LINKED ISSUE #430: Suite with non-compatible checks return "empty suite"
For example, when a suite has a check that requires two datasets but suite is given only one, it will print “empty suite”. Instead, this needs to be in the “checks that couldn’t run” table.

matanper:
resolve #430 

image

### DIFF
--- a/deepchecks/base/check.py
+++ b/deepchecks/base/check.py
@@ -14,6 +14,7 @@ import abc
 import enum
 import inspect
 import re
+import traceback
 from collections import OrderedDict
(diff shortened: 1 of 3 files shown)


noamzbr (approved):
Looks good, are there any notebooks to re-run for this?

matanper:
> Looks good, are there any notebooks to re-run for this?

I don't know but we have the automatic rerun anyway

PR #444, opened 2022-01-03
Performance report degradation fix

JKL98ISR:


ItayGabbay (approved):

matanper:
We have a global function to calculate ratio maybe better to use it

JKL98ISR:
> We have a global function to calculate ratio maybe better to use it

it's not the same scenario

matanper:
> > We have a global function to calculate ratio maybe better to use it
> 
> it's not the same scenario

what's the difference? that you have the diff? won't it be better to do `global_ratio(score1 - score2, score2)` and handle all the edge cases in single function?

JKL98ISR:
> > > We have a global function to calculate ratio maybe better to use it
> > 
> > 
> > it's not the same scenario
> 
> what's the difference? that you have the diff? won't it be better to do `global_ratio(score1 - score2, score2)` and handle all the edge cases in single function?

oh I see, ok

JKL98ISR:
@matanper I can't because i want (0 - 0)/0 = 0 to act more like any other number case, so I cant use that function.
best I can do in that function is return 1 if both are 0 which is incorrect

matanper:
the main problem I currently see is what to do with metrics that have a range like [-1,1]. you can have score1 = 0, and score2=-0.01. in cases like that it seems that the percentage from the overall range is more important than the relative between the scores. @noamzbr WDYT?

deepchecks/checks/performance/performance_report.py
```diff
@@ -153,6 +153,13 @@ def add_condition_train_test_performance_degradation_ratio_not_greater_than(self
         Args:
             threshold: maximum degradation ratio allowed (value between 0 to 1)
         """
+        def _ratio_calc(score_1, score_2):
```
noamzbr, line 156:
```suggestion
        def _ratio_of_change_calc(score_1, score_2):
```

deepchecks/checks/performance/performance_report.py
```diff
@@ -169,7 +176,7 @@ def condition(check_result: pd.DataFrame) -> ConditionResult:
                     test_scores_dict = dict(zip(test_scores_class['Metric'], test_scores_class['Value']))
                     train_scores_dict = dict(zip(train_scores_class['Metric'], train_scores_class['Value']))
                     # Calculate percentage of change from train to test
-                    diff = {score_name: ((score - test_scores_dict[score_name]) / score)
+                    diff = {score_name: _ratio_calc(score, test_scores_dict[score_name])
                             for score_name, score in train_scores_dict.items()}
                     failed_scores = [k for k, v in diff.items() if abs(v) > threshold]
```
noamzbr, line 181: Reading the description I would assume only score_train > score_test can raise the condition, but here I see abs()

noamzbr (commented):
In general - I think we should improve the naming here, making it clearer it refers to a change in performance, as measured by percent.

in model_error_analysis we called it performance_relative_difference for e.g.

noamzbr:
> the main problem I currently see is what to do with metrics that have a range like [-1,1]. you can have score1 = 0, and score2=-0.01. in cases like that it seems that the percentage from the overall range is more important than the relative between the scores. @noamzbr WDYT?

Agreed and I have the same problem with simple model report - when it seems like accuracy 0.96 is very close to 0.99 and should trigger the "your model is really close to a simple model" condition, but in fact if the imbalance is 96% than improving to 0.99 is actually really good.

I can't see how to solve that without defining the upper range. This can be done perhaps by running a scorer with prediction = labels. Then we can calc the % change compared to the range. WDYT?

noamzbr:
> @noamzbr there are no scorers with infinite upper range?

RMSE / MSE and so on. For them, the range could be from 0 to their current max value

matanper:
> > @noamzbr there are no scorers with infinite upper range?
> 
> RMSE / MSE and so on. For them, the range could be from 0 to their current max value

not really, their upper value is 0 (best score) so they are infinite from the bottom which we care less if we want to compare to the best possible score.

noamzbr (approved):

PR #449, opened 2022-01-03
catch condition errors

JKL98ISR:


matanper (changes requested):
Add test of failing condition

JKL98ISR:
> Add test of failing condition

added

matanper (approved):

PR #454, opened 2022-01-03
Bugfix/fix 453

### LINKED ISSUE #453: [BUG] using .head(2) to validate scorers fails when sample contains single label
This happens inside `validate_fitting` when should_return_array = False.

Running `self._run_score(model, df.head(2), dataset)` when head contains only single kind of label raises the following bug:

> TypeError: issubclass() arg 2 must be a class or tuple of classes

noamzbr:
resolves #453

### DIFF
--- a/deepchecks/utils/metrics.py
+++ b/deepchecks/utils/metrics.py
@@ -137,7 +137,7 @@ class DeepcheckScorer:
         """Validate given scorer for the model and dataset."""
         df = self.filter_nulls(dataset)
         if should_return_array:
-            # In order for multiclass scorer to return array in right length need to pass him samples from all labels
+            # In order for scorer to return array in right length need to pass him samples from all labels
             single_label_data = df.groupby(dataset.label_name).head(1)
             result = self._run_score(model, single_label_data, dataset)
             if not isinstance(result, np.ndarray):
@@ -157,7 +157,9 @@ class DeepcheckScorer:
                                                   f'returned only {len(result)} elements in the score array value')
 
         else:
-            result = self._run_score(model, df.head(2), dataset)
+            # In order for scorer to return array in right length need to pass him samples from all labels
+            single_label_data = df.groupby(dataset.label_name).head(1)
+            result = self._run_score(model, single_label_data, dataset)
             if not isinstance(result, Number):
                 raise errors.DeepchecksValueError(f'Expected scorer {self.name} to return number '
                                                   f'but got: {type(result).__name__}')
--- a/deepchecks/utils/metrics.py
+++ b/deepchecks/utils/metrics.py
@@ -157,7 +157,7 @@ class DeepcheckScorer:
                                                   f'returned only {len(result)} elements in the score array value')
 
         else:
-            # In order for scorer to return array in right length need to pass him samples from all labels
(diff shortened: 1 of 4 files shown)


deepchecks/utils/metrics.py
```diff
@@ -157,7 +157,9 @@ def validate_fitting(self, model, dataset: 'base.Dataset', should_return_array:
                                                   f'returned only {len(result)} elements in the score array value')
 
         else:
-            result = self._run_score(model, df.head(2), dataset)
+            # In order for scorer to return array in right length need to pass him samples from all labels
+            single_label_data = df.groupby(dataset.label_name).head(1)
```
matanper, line 161: in here the scorer should return number so no need for the group by (and also comment is incorrect)

tests/checks/performance/performance_report_test.py
```diff
@@ -69,6 +70,26 @@ def test_classification(iris_split_dataset_and_model):
                 metric_col = class_col.loc[class_col['Metric'] == metric]
                 assert_that(metric_col['Value'] , close_to(1, 0.3))
 
+
+def test_classification_binary(iris_dataset_single_class_labeled):
+    # Arrange
+    train, test = train_test_split(iris_dataset_single_class_labeled.data, test_size=0.33, random_state=42)
+    train_ds = iris_dataset_single_class_labeled.copy(train)
```
matanper, line 77: why do you have to copy?

tests/checks/performance/performance_report_test.py
```diff
@@ -69,6 +70,26 @@ def test_classification(iris_split_dataset_and_model):
...
+    clf = RandomForestClassifier(random_state=0)
+    clf.fit(train_ds.features_columns, train_ds.label_col)
+    check = PerformanceReport()
+
+    # Act X
+    result = check.run(train_ds, test_ds, clf).value
+    # Assert
+    for dataset in ['Test', 'Train']:
+        dataset_col = result.loc[result['Dataset'] == dataset]
+        for metric in ['Accuracy (Default)', 'Precision (Default)', 'Recall (Default)']:
```
matanper, line 88: maybe use the default dict `keys()` for future changes

deepchecks/utils/metrics.py
noamzbr: The groupby fixes the bug - in order to run, the sklearn scorer must have at least one sample of each label

About the comment-  will fix

tests/checks/performance/performance_report_test.py
noamzbr: It's a way to split to train and test a dataset that arrived un-splitted

tests/checks/performance/performance_report_test.py
noamzbr: fixed

tests/checks/performance/performance_report_test.py
matanper: how is that a split?

noamzbr: first I do train_test_split, then I create a dataset from each of the splitted dataframes copying the metadata from the original dataset

matanper (approved):

PR #456, opened 2022-01-03
remove label_name from dataset constructor

### LINKED ISSUE #423: [Feature] label field should be able to get label column name and label series
### Currently
`Dataset` object can get both `label` and `label_name` as parameters.

### suggestion
The parameter `label` should be able to get a `pd.Series` and identify it as such and also be able to get a `string`/`int` and identify it as column in the DataFrame

(same for `index` and `date`)

matanper:
resolve #423


deepchecks/base/dataset.py
```diff
@@ -43,17 +43,13 @@ class Dataset:
     Args:
         df (pandas.DataFrame):
             A pandas DataFrame containing data relevant for the training or validating of a ML models.
-        label (pandas.Series)
-            A pandas series containing data of the labels. Will be joined to the data dataframe with the name
-            given by `label_name` parameter or 'target' by default.
+        label (t.Union[Hashable, pd.Series, pd.DataFrame, np.ndarray])
+            label column provided as name in the dataframe or label object which will be concatenated to the dataframe
```
shir22, line 47:
```suggestion
            label column provided either as a string with the name of an existing column in the DataFrame or a label object including the label data (pandas Series/DataFrame or a numpy array) that will be concatenated to the data in the DataFrame
```

deepchecks/base/dataset.py
```diff
@@ -122,28 +117,41 @@ def __init__(
...
-                pd.testing.assert_index_equal(self._data.index, label.index)
-                self._data[label_name] = label
+        if label is None:
+            label_name = None
+        else:
+            if isinstance(label, (pd.Series, pd.DataFrame, np.ndarray)):
+                # Validate label shape
+                if label.shape[0] != self._data.shape[0]:
+                    raise DeepchecksValueError('Number of samples of label and data must be equal')
+                if len(label.shape) > 1 and label.shape[1] != 1:
```
shir22, line 127: doesn't cover cases like if someone transfers a 3 (or more) dimensional array

deepchecks/base/dataset.py
```diff
@@ -122,28 +117,41 @@ def __init__(
...
+                    raise DeepchecksValueError('Label must be a column vector')
+
+                # Get label name
+                if isinstance(label, (pd.Series, pd.DataFrame)) and label.name is not None:
```
shir22, line 131: if it's a DataFrame we should take the name of the (single) column, and not of the DataFrame itself, no?

deepchecks/base/dataset.py
```diff
@@ -122,28 +117,41 @@ def __init__(
...
+                    label_name = label.name
+                    if label_name in self._data.columns:
+                        raise DeepchecksValueError(f'Data has column with name "{label_name}", use pandas rename to'
+                                                   f' change label name or remove the column from the dataframe')
+                else:
+                    label_name = 'target'
```
shir22, line 137: should be written in parameter documentation?
that we take the name of the series / column in dataframe, and if it's a numpy array / no name then we call it 'target'?

tests/base/dataset_test.py
```diff
@@ -747,26 +746,60 @@ def test_train_test_split_changed(iris):
...
 def test_set_label_type(iris):
     # Arrange
-    label = iris['target']
+    label = iris['target'].rename('actual')
     data = iris.drop('target', axis=1)
-    dataset = Dataset(data, label, label_name='actual', label_type='regression_label')
+    dataset = Dataset(data, label, label_type='regression_label')
     # Assert
     assert_that(dataset.label_type, is_('regression_label'))
+
```
shir22, line 772: Tests for invalid label types?
e.g. DataFrame with multiple columns, numpy array with incorrect dimensions, string that doesn't exist in dataframe, etc.

shir22 (changes requested):

deepchecks/base/dataset.py
matanper: updated

deepchecks/base/dataset.py
matanper: do we support these use cases?

deepchecks/base/dataset.py
matanper: fixed

deepchecks/base/dataset.py
matanper: added

deepchecks/base/dataset.py
```diff
@@ -128,12 +132,19 @@ def __init__(
...
                 # Get label name
-                if isinstance(label, (pd.Series, pd.DataFrame)) and label.name is not None:
+                if isinstance(label, pd.Series) and label.name is not None:
                     label_name = label.name
                     if label_name in self._data.columns:
                         raise DeepchecksValueError(f'Data has column with name "{label_name}", use pandas rename to'
                                                    f' change label name or remove the column from the dataframe')
-                else:
+                elif isinstance(label, pd.DataFrame):
+                    label_name = label.columns[0]
```
shir22, line 141: Validation that supplied DataFrame has indeed only one column? (e.g. user can mix it up and send us a DataFrame with "all data incl. label")

deepchecks/base/dataset.py
```diff
@@ -44,7 +44,12 @@ class Dataset:
         df (pandas.DataFrame):
             A pandas DataFrame containing data relevant for the training or validating of a ML models.
         label (t.Union[Hashable, pd.Series, pd.DataFrame, np.ndarray])
-            label column provided as name in the dataframe or label object which will be concatenated to the dataframe
+            label column provided either as a string with the name of an existing column in the DataFrame or a label
+            object including the label data (pandas Series/DataFrame or a numpy array) that will be concatenated to the
+            data in the DataFrame. in case of label data the following logic is applied to set the label name:
+                - Series: takes the series name or 'target' if name is empty
```
shir22, line 50: Great :)

deepchecks/base/dataset.py
shir22: doesn't seem like something we should support...
in other words - exactly my question (that we should verify that the user passed valid input)
that here someone can pass a 3d np.ndarray, e.g. of shape (<num_samples_in_data>, 1, 42) , then the code will run successfully....

(That's also what I meant in the tests - verifying that it fails as it should if someone sends a DataFrame with more than one column, and same for ndarrays with too many dimensions)

btw coming to thing of it - for numpy I would sugggest separate treatment - using np.squeeze, and then verifying that it's a 1d array with the correct length (cause np arrays have various weird dimension behaviors.... and this also makes us note care if it's a "column vector" or a "row vector")

tests/base/dataset_test.py
shir22: lol TBH I didn't know we had a "label type" argument, cool
I actually meant the "improper shapes of np arrays and/or dataframes"

shir22 (changes requested):

ItayGabbay (commented):
I am not sure we want to support a pd.DataFrame as an accepted type for label...

deepchecks/base/dataset.py
```diff
@@ -43,17 +43,18 @@ class Dataset:
...
             A pandas DataFrame containing data relevant for the training or validating of a ML models.
-        label (pandas.Series)
-            A pandas series containing data of the labels. Will be joined to the data dataframe with the name
-            given by `label_name` parameter or 'target' by default.
+        label (t.Union[Hashable, pd.Series, pd.DataFrame, np.ndarray])
+            label column provided either as a string with the name of an existing column in the DataFrame or a label
+            object including the label data (pandas Series/DataFrame or a numpy array) that will be concatenated to the
+            data in the DataFrame. in case of label data the following logic is applied to set the label name:
+                - Series: takes the series name or 'target' if name is empty
+                - DataFrame: expect single column in the dataframe and use its name
```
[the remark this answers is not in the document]
noamzbr: examine the following usecase:

`Dataset(df[feature_list], label=df[['target']])`

In that case the label will be passed as dataframe. Agreed that it's not very likely, but could happen and don't think we should forbid that.

deepchecks/base/dataset.py
matanper: we looked at pandas, and they are not allowing 3+ dimensions. so right now I'll check for a maximum of 2-dim, and that at least one of them is equal 1 (meaning we accept either a column or row vector). we can support any 2-dim shape with squeeze but it might be confusing

shir22: I was talking about numpy for the 3d (plus)
pandas indeed has only up to 2-dim
but numpy has unlimited (that's why offered the squeeze)

matanper: yea I meant we looked at how pandas handled numpy arrays as input

noamzbr: Yeah pandas doesn't allow more then 2-d and I agree that there is no reason to allow more. The user can squeeze by himself IMO

noamzbr (approved):
Please open an issue to also support lists in a future PR

deepchecks/base/dataset.py
shir22: oh, ok, cool :)

shir22: oh, cool :)

PR #461, opened 2022-01-04
[issue-439] added 'Number of samples' column to the result dataframe;

### LINKED ISSUE #439: Add count to classification report
When plotting the metrics per class, it is useful to see how many samples belong to each class

yromanyshyn:
list of changes:
- added 'Number of samples' column to the result data frame;
- modified bar plot, now it also shows 'Number of samples' value within hover box;

resolves #439


noamzbr (approved):
Really liked the solution of adding it as hover data!
Could you please add this also to Simple Model Comparison?

yromanyshyn:
>Really liked the solution of adding it as hover data!
Could you please add this also to Simple Model Comparison?

I will create a separate branch for it

PR #463, opened 2022-01-04
Updating long description

ItayGabbay:


setup.py
```diff
@@ -36,8 +82,8 @@ def readme():
     packages=setuptools.find_packages(),
     install_requires=install_requires,
     license_files = ('LICENSE', ),
-    description = 'Package for validating your machine learning model and data',
-    long_description=readme(),
+    description = DOCLINES[0],
+    long_description="\n".join(DOCLINES[2:]),
```
noamzbr, line 86: Isn't it a bit cumbersome to update this on top of the readme? Can't we have something programmatic taking the readme until Usage Examples?

ItayGabbay: I find it a bit dangerous since we have rst directives in the readme (that can be added anywhere) that are not supported in pypi. This may crash the release. 
I don't think what I wrote here is gonna change a lot...

noamzbr: I'll roll with that

noamzbr (approved):

PR #468, opened 2022-01-04
pipeline model can also no be of type BaseEstimator

### LINKED ISSUE #425: [BUG] Model Identification in sklearn Pipeline should look at the last step
```
pipe = Pipeline(steps=[
    ('transform', transformers),
    ('handle_nans', SimpleImputer(strategy='most_frequent')),
    ('modle', clf)
])

```

Boosting Overfit | DeepchecksValueError: Unsupported model of type: SimpleImputer
-- | --

Boosting Overfit check should have identified the pipeline as the last step (`clf` which was `catboost`)

benisraeldan:
resolves #425 

the last step in the pipeline must be an estimator, but doesn't have to inherit from sklearn BaseEstimator

### DIFF
--- a/deepchecks/utils/model.py
+++ b/deepchecks/utils/model.py
@@ -14,8 +14,6 @@ from typing import Union
 from sklearn.pipeline import Pipeline
 from sklearn.base import BaseEstimator
 
-from deepchecks.errors import DeepchecksValueError
-
 
 __all__ = ['get_model_of_pipeline']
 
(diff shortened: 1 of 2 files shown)


matanper:
We have `model_type_validation`, wonder if we should use it in the function or in the checks

benisraeldan:
> We have `model_type_validation`, wonder if we should use it in the function or in the checks

added `get_model_of_pipeline` to `model_type_validation`

matanper (approved):

PR #485, opened 2022-01-04
add dataset script and notebook for phishing url example

noamzbr:
Add an example notebook for the usage of deepchecks package in a realistic usecase.

@ItayGabbay Could you please see that the dataset script is OK? And also how to properly make this render in the docs
@shir22 - need to think where the notebook should sit in docs and what should reference it.


shir22:
Notebook should sit under guides folder (docs/source/examples/guides)
in a new folder inside, I thought of `use_cases`, WDYT?

ItayGabbay (commented):
Dataset script looks fine. 
The only thing left is to render the notebook. Without it the nightly CI will fail

ItayGabbay (approved):

PR #488, opened 2022-01-04
Change condition of simple model check to use gain

### LINKED ISSUE #474: Condition on SimpleModelComparison should be gain of model over simple model
Currently, the default conditions checks that the model performance is better than the simple model by computing the ratio between model performance and simple model performance.

This is misleading in highly unbalanced datasets, as @shaypal5 has pointed out. 

Proposed solution - define the condition as "what percent of the remaining (unachieved) performance of the simple model did the user model manage to achieve". This quantity will be called gain and be measures as:
change_in_performance / (max_possible_performance - baseline_performance)

matanper:
resolve #474


tests/checks/performance/simple_model_comparison_test.py
```diff
@@ -197,5 +197,24 @@ def test_regression_tree_max_depth(diabetes_split_dataset_and_model):
...
+
+
+def assert_classification(result, metrics=None):
+    metrics = metrics or [DEFAULT_SINGLE_SCORER_MULTICLASS_NON_AVG]
+    class_matchers = {clas: has_entries({'Origin': close_to(1, 1), 'Simple': close_to(1, 1)})
+                      for clas in result['classes']}
+    matchers = {metric: has_entries(class_matchers) for metric in metrics}
+    assert_that(result['scores'], has_entries(matchers))
+    assert_that(result['scorers_perfect'], has_entries({metric: is_(1) for metric in metrics}))
+    assert_that(result['classes'], contains_exactly(0, 1, 2))
```
noamzbr, line 220: Don't want to thread on the toes of this [issue](https://github.com/deepchecks/deepchecks/issues/491), but may want to add a test for string labels...

noamzbr (commented):
Comment about the notebook - why set the condition minimum condition to 1.2? Seems like it should have triggered on any value (including the default), as the simple model outright beats the use model

noamzbr (commented):
maximum_ratio - docstring explaining this quantity can be improved

noamzbr (approved):

PR #500, opened 2022-01-05
Update the Dataset Object Documentation

### LINKED ISSUE #484: [DOCS] Update Dataset Object in docs

benisraeldan:
#### Reference Issues/PRs

Resolves #484


ItayGabbay (commented):
I really like the writing style!

Some comments:
- I didn't comment on wordings
- I think we need to show more functionalities about the dataset, for example functions like copy, from_numpy, datetime parsing, passing labels as series, and more.
- We can take inspiration from a not-so-related library which in my opinion has a great quickstart:

docs/source/user-guide/dataset_object.rst
```diff
@@ -3,8 +3,139 @@
...
+A Quick Introduction
+====================
+
+The Dataset Object is a simple and robust Object to store tabular data and metadata relevant to Machine Learning.
+
+Common information used in Machine Learning that is aggregated in the Dataset:
+ - features
+    List of column names. This is the features that are passed to the model. If not defined, columns not defined as something else is considered a feature.
+ - cat_features
+    List of column names. A subset of the features. Categorical features normally require some preprocessing before being passed to the model.
```
ItayGabbay, line 17:
```suggestion
    List of column names. A subset of the features. Categorical features normally require some preprocessing before being passed to the model. If not specified, the categorical features are inferred automatically from the data itself.
```

docs/source/user-guide/dataset_object.rst
```diff
@@ -3,8 +3,139 @@
...
+ - label
```
ItayGabbay, line 18: Actually, when reading the doc I think we should rename the label parameter to target.
What do you guys think?
@shir22 @noamzbr @benisraeldan @matanper  @nirhutnik @JKL98ISR

docs/source/user-guide/dataset_object.rst
```diff
@@ -3,8 +3,139 @@
...
+   Either name of a column, or `pd.Series`. The classes of the classification problem or values of a regression problem.
+ - index_name
+    Name of the index column. This can be useful to track indexes duplicate etc. Like the other arguments, this is not manditory.
+ - date_name
```
ItayGabbay, line 22: I think we should specify the accepted format with an example

docs/source/user-guide/dataset_object.rst
```diff
@@ -3,8 +3,139 @@
...
+
+
+A Quick Introduction
+====================
+
+The Dataset Object is a simple and robust Object to store tabular data and metadata relevant to Machine Learning.
+
+Common information used in Machine Learning that is aggregated in the Dataset:
+ - features
+    List of column names. This is the features that are passed to the model. If not defined, columns not defined as something else is considered a feature.
```
ItayGabbay, line 15:
```suggestion
    List of column names. This parameter specifies the features the model is trained on. If not defined, columns that are not defined as label, date, or index are considered as features.
```

docs/source/user-guide/dataset_object.rst
ItayGabbay: And also we should line here to the categorical inference heuristic detailed below.

docs/source/user-guide/dataset_object.rst
```diff
@@ -3,8 +3,139 @@
...
+    List of column names. A subset of the features. Categorical features normally require some preprocessing before being passed to the model.
+ - label
+   Either name of a column, or `pd.Series`. The classes of the classification problem or values of a regression problem.
+ - index_name
+    Name of the index column. This can be useful to track indexes duplicate etc. Like the other arguments, this is not manditory.
+ - date_name
+    Name of the date column. The date the sample was acquired (Useful for TimeSeries problems). This column is standardized using `pd.Timestamp`.
+
+All these and more can be defined and accessed on the Dataset Object.
+This allows the user the flexibility to define these column names whichever they seam fit,
```
ItayGabbay, line 26:
```suggestion
This allows the user the flexibility to define these column names whichever they seem fit,
```

ItayGabbay (changes requested):

docs/source/user-guide/dataset_object.rst
shir22: yes, can be good - though this is also something that I think should be explicit and clear in the API Reference (also so we won't need to maintain double...)

docs/source/user-guide/dataset_object.rst
shir22: See my general note above about phrasing

shir22 (changes requested):

docs/source/user-guide/dataset_object.rst
```diff
@@ -3,76 +3,74 @@
...
+This allows the user the flexibility to define these column names whichever they seem fit,
+and allows use of the these tags to standardize development around this data.
+
+
+The Basics
+==========
+
+Lets start with building a simple pd.DataFrame that we might already know as a popular data structure to store data.
+But we also have other metadata that we have other use for, and not only as features for the model.
+
```
shir22, line 31: All of the code blocks should be [literal blocks ](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#literal-blocks) (right?)
(like in the various numpy code examples)

docs/source/user-guide/dataset_object.rst
```diff
@@ -3,76 +3,74 @@
...
-    * - convert_date
-      - bool
-      - Whether to convert date column if it's numeric to date
-      - True
+
+
+A Quick Introduction
+====================
+
+The Dataset Object is a simple and robust Object to store tabular data and metadata relevant to Machine Learning.
```
shir22, line 11:
```suggestion
The Dataset Object is a simple and robust Object to store tabular data and metadata relevant for Machine Learning.
```

docs/source/user-guide/dataset_object.rst
```diff
@@ -3,76 +3,74 @@
...
+It allows deepchecks to understand the context of the data with minimal hassle for the user.
+
+Common information used in Machine Learning that are aggregated in the Dataset:
```
shir22, line 14:
```suggestion
The Dataset's common metadata properties:
```

docs/source/user-guide/dataset_object.rst
```diff
@@ -3,76 +3,74 @@
...
+ - **cat_features** - Categorical features normally require some preprocessing before being passed to the model. If not specified, the categorical features are inferred automatically from the data itself. (`Inferring Features And Categorical Features`_ )
+ - **label** - The classes of the classification problem or values of a regression problem.
```
shir22, line 16: put this one first.
phrasing should be clear and short (that this represents the target that we're trying to predict in the data. 
(more high level, you wrote it very technical)

docs/source/user-guide/dataset_object.rst
```diff
@@ -3,76 +3,74 @@
...
+
+
+A Quick Introduction
+====================
+
+The Dataset Object is a simple and robust Object to store tabular data and metadata relevant to Machine Learning.
+It allows deepchecks to understand the context of the data with minimal hassle for the user.
+
+Common information used in Machine Learning that are aggregated in the Dataset:
+ - **cat_features** - Categorical features normally require some preprocessing before being passed to the model. If not specified, the categorical features are inferred automatically from the data itself. (`Inferring Features And Categorical Features`_ )
```
shir22, line 15: Would skip the first part, too specific for here. Would have a very short description - something like "The features that are of categorical type"

docs/source/user-guide/dataset_object.rst
```diff
@@ -3,76 +3,74 @@
...
+ - **label** - The classes of the classification problem or values of a regression problem.
+ - **index_name** - This can be useful to track indexes duplicate etc. Like the other arguments, this is not mandatory.
```
shir22, line 17: State something about the fact that some datasets have a unique index that may be meaningful (e.g. represents a serial number)

docs/source/user-guide/dataset_object.rst
```diff
@@ -3,76 +3,74 @@
...
+ - **date_name** - The date the sample was acquired (Useful for TimeSeries problems). This column is standardized using `pd.Timestamp`.
```
shir22, line 18: Not necessarily...
A date column representing the sample (wasn't necessarily acquired at that time), may be the time at which it was acquired. (for example - may also be "number of clicks on day X, while they where all acquired post-mortem)

docs/source/user-guide/dataset_object.rst
```diff
@@ -3,76 +3,74 @@
...
+ - **features** - Specifies the columns used by model for training. Used for defining only a subset of the columns in the data as features. If not defined, all columns that aren't explicitly specified as `label`, `date`, or `index` are considered as features.
+
+All these and more can be defined and accessed on the Dataset Object.
```
shir22, line 21: Link here explicitly to the API reference at the bottom?

docs/source/user-guide/dataset_object.rst
```diff
@@ -3,76 +3,74 @@
...
+This allows the user the flexibility to define these column names whichever they seem fit,
+and allows use of the these tags to standardize development around this data.
+
+
+The Basics
+==========
+
+Lets start with building a simple pd.DataFrame that we might already know as a popular data structure to store data.
+But we also have other metadata that we have other use for, and not only as features for the model.
```
shir22, line 30:
```suggestion
Lets start with building a simple `pd.DataFrame` that we might already know as a popular data structure to store data, and demonstrate how the Dataset object should be defined.
```

docs/source/user-guide/dataset_object.rst
```diff
@@ -3,76 +3,74 @@
...
+
+    >>> d = {"id": [1,2,3,4],
+    ...      "feature1": [0.1,0.3,0.2,0.6],
+    ...      "feature2": [4,5,6,7],
+    ...      "cat_feature": [0,0,0,1],
```
shir22, line 35:
```suggestion
    ...      "categorical_feature": [0,0,0,1],
```

docs/source/user-guide/dataset_object.rst
```diff
@@ -3,76 +3,74 @@
...
+    ...      "class": [1,2,1,2]}
+    ... df = pd.DataFrame(d)
+    ... dataset = Dataset(df, label="class", index_name="id", cat_features=["category"])
```
shir22, line 38:
```suggestion
    ... dataset = Dataset(df, label="class", index_name="id", cat_features=["categorical_feature"])
```

docs/source/user-guide/dataset_object.rst
```diff
@@ -3,76 +3,74 @@
...
+
+Now we have built a Dataset that we can easily pull features and labels from:
```
shir22, line 40:
```suggestion
Now that we have a Dataset we can also inspect its defined features and label:
```

docs/source/user-guide/dataset_object.rst
```diff
@@ -3,76 +3,74 @@
...
+
+    >>> dataset.features
+    ['feature1', 'feature2', 'category']
+    >>> dataset.label_name
+    ['class']
+
+We can also get pd.DataFrames from columns that interest us e.g.:
```
shir22, line 47: Maybe more interesting to demonstrate instead (or to state both options and show one of them) - how to show the whole dataframe with `dataset.data`

docs/source/user-guide/dataset_object.rst
```diff
@@ -3,76 +3,74 @@
...
+
+    >>> dataset.features_columns
+        feature1	feature2	category
+    0	0.1	        4	        0
+    1	0.3	        5	        0
+    2	0.2	        6	        0
+    3	0.6	        7	        1
+
+Now we can see that only features columns are pulled, which is save us the hassle
+of tracking externally what is a feature column and what isn't one.
```
shir22, line 57: a bit redundant (long, doesn't really add information)

docs/source/user-guide/dataset_object.rst
```diff
@@ -3,76 +3,74 @@
...
+
+    >>> dataset.label_col
+        class
+    0	1
+    1	2
+    2	1
+    3	2
+
+
+now instead of needing keeping track fo the label column, we can track it within the the Dataset Object.
```
shir22, line 70: ok now I get it - you want to show that this can also be extracted directly - to get the relevant slice of the dataframe, with the specifc class parameters.
This is good.
But the explanation should be improved. We're demonstrating convenient usage options, that can be useful for the user, however no need to try to make it sound fancy, can keep it simple.

docs/source/user-guide/dataset_object.rst
```diff
@@ -3,76 +3,74 @@
...
 
 Inferring Features And Categorical Features
-================================================
+===========================================
 Dataset defines which columns of the data are features and of them which are categorical features.
 For features, if parameter `features` not passed explicitly, all will be considered features apart from the label, index and date.
 For categorical features, if parameter `cat_features` not passed explicitly, the following logic runs on every column to determine
```
shir22, line 76:
```suggestion
The Dataset defines which columns of the data are considered as the model's features, and of them, which are treated as categorical features.
For features, if parameter `features` not passed explicitly, all columns besides the the `label`, `index `and `date `will be considered as features.
For categorical features, if the parameter `cat_features` was not passed explicitly, the following inference logic will run on each column to determine
```

shir22, line 76: and replace sentence after (which it won't let me review on since it's not part of this PR, as:
how it will be treated.

(instead of `whether the column is categorical or not:`)

docs/source/user-guide/dataset_object.rst
```diff
@@ -83,46 +81,57 @@ whether the column is categorical or not:
 * else:
     *  number of unique < `max_categories` AND (number of unique / number of samples) < `max_categorical_ratio`
 
-Creating a Dataset
-======================
+Useful Functions
+================
+
+Create Dataset From a Numpy Arrays
```
shir22, line 87:
```suggestion
Create a Dataset from Numpy Arrays
```

docs/source/user-guide/dataset_object.rst
```diff
@@ -83,46 +81,57 @@ whether the column is categorical or not:
...
+    ... )
+
+All the rest of the Dataset's properties can be passed also as a regular keyword arguments:
+
+    >>> Dataset.from_numpy(
+    ...     features, labels,
+    ...     feature_names=['feat1', 'feat2', 'feat3',],
+    ...     label_name='target',
+    ...     max_float_categories=10
+    ... )
```
shir22, line 111:
```suggestion
All the rest of the Dataset's properties can be passed also as regular keyword arguments:

    >>> Dataset.from_numpy(
    ...     features, labels,
    ...     feature_names=['feat1', 'feat2', 'feat3',],
    ...     label_name='target',
    ...     max_float_categories=10
    ... )
```

Suggested a small grammar fix, but not sure - why is this needed - what does it demonstrate?

docs/source/user-guide/dataset_object.rst
```diff
@@ -83,46 +81,57 @@ whether the column is categorical or not:
...
 
-From a DataFrame
+
+Train Test Split
 ~~~~~~~~~~~~~~~~
-The default dataset constructor is expecting to get a dataframe. the rest of the properties
-are optional, but if your data have date/index/label you would want to define them.
 
-.. code-block:: python
+Using the `sklearn.model_selection.train_test_split <https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.train_test_split.html>`_ but also copies the metadata to each instance of the split.
```
shir22, line 117:
```suggestion
Uses internally `sklearn.model_selection.train_test_split <https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.train_test_split.html>`_ (so same arguments can be passed) and also copies the metadata to each instance of the split and returns two Datasets.
```

docs/source/user-guide/dataset_object.rst
```diff
@@ -83,46 +81,57 @@ whether the column is categorical or not:
...
 
-    Dataset(my_dataframe, features=['feat1', 'feat2', 'feat3'], label='target', index='id', date='timestamp')
+    >>> train_dataset, test_dataset = dataset.train_test_split()
```
shir22, line 119: not clear here if "dataset" is an existing Dataset that you called dataset, or if it's the class
give an example that gives the whole context (includes the actual creation).
in other words - where is the original dataset here?

and can give a clearer example, e.g. use the` train_test_split'`s popular `stratify` argument, and send the label column to it

docs/source/user-guide/dataset_object.rst
```diff
@@ -83,46 +81,57 @@ whether the column is categorical or not:
...
 
-.. code-block:: python
+Copy
+~~~~
 
-    features = np.array([[0.25, 0.3, 0.3], [0.14, 0.75, 0.3], [0.23, 0.39, 0.1]])
-    labels = np.array([0.1, 0.1, 0.7])
-    dataset_with_labels = Dataset.from_numpy(features, labels)
-    dataset_without_labels = Dataset.from_numpy(features)
+Copy allows copying of metadata on a new dataframe. This can be used on data that already is passed the train test split.
```
shir22, line 125: rephrase
(second sentence is not clear, but in general this line can be improved)

shir22: also maybe give a different example, to not confuse the user the the train_test_split which is right above
(they might think "wait, after I do the train_test_split I need to copy the metadata?"

docs/source/user-guide/dataset_object.rst
```diff
@@ -83,46 +81,57 @@ whether the column is categorical or not:
...
+    >>> test_dataset = train_dataset.copy(test_df)
 
-.. code-block:: python
 
-    Dataset.from_numpy(
-        features, labels,
-        feature_names=['feat1', 'feat2', 'feat3',],
-        label_name='target'
-    )
+Full API Reference
```
shir22, line 130:
```suggestion
Dataset API Reference
```

shir22: Do we want the API reference itself? maybe only a link to it in several places?

shir22 (changes requested):

docs/source/user-guide/dataset_object.rst
matanper: Did you run this code example? I think this doesn't work since the features names won't match the column names

docs/source/user-guide/dataset_object.rst
benisraeldan: I'm open to suggestions for linking, I couldn't find how to link to the generated API

ItayGabbay (approved):

PR #501, opened 2022-01-05
fix errors related to string labels

JKL98ISR:
fixed issue in Model Error Analysis and Calibration Score

### DIFF
--- a/deepchecks/checks/performance/calibration_score.py
+++ b/deepchecks/checks/performance/calibration_score.py
@@ -52,7 +52,7 @@ class CalibrationScore(SingleDatasetBaseCheck):
         briers_scores = {}
 
         if len(dataset.classes) == 2:
-            briers_scores[0] = brier_score_loss(ds_y, y_pred[:, 1])
+            briers_scores[0] = brier_score_loss(ds_y, y_pred[:, 1], pos_label=dataset.classes[1])
         else:
             for class_index, class_name in enumerate(dataset.classes):
                 prob_pos = y_pred[:, class_index]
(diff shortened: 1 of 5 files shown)


deepchecks/checks/performance/model_error_analysis.py
```diff
@@ -147,11 +148,13 @@ def run(self, train_dataset: Dataset, test_dataset: Dataset, model=None) -> Chec
...
-            def scoring_func(dataset):
+            def scoring_func(dataset: Dataset):
                 return per_sample_mse(dataset.label_col, model.predict(dataset.features_columns))
         else:
-            def scoring_func(dataset):
-                return per_sample_binary_cross_entropy(dataset.label_col,
+            def scoring_func(dataset: Dataset):
+                le = preprocessing.LabelEncoder()
+                encoded_label = le.fit_transform(dataset.label_col)
+                return per_sample_binary_cross_entropy(encoded_label,
```
nirhutnik, line 157: Why not do the encoding outside of the function? I think that the implicit assumption that LabelEncoder.fit_transform will give the same result for train and test might cause bugs in some future.

deepchecks/checks/performance/calibration_score.py
```diff
@@ -52,7 +52,7 @@ def _calibration_score(self, dataset: Dataset, model):
         briers_scores = {}
 
         if len(dataset.classes) == 2:
-            briers_scores[0] = brier_score_loss(ds_y, y_pred[:, 1])
+            briers_scores[0] = brier_score_loss(ds_y, y_pred[:, 1], pos_label=dataset.classes[1])
```
nirhutnik, line 55: Still needed if the results are passed as [0,1]?

nirhutnik (changes requested):

deepchecks/checks/performance/model_error_analysis.py
JKL98ISR: did a fix

deepchecks/checks/performance/calibration_score.py
JKL98ISR: no but it doesn't matter

nirhutnik (approved):

PR #504, opened 2022-01-05
Fix string labels in performance report / simple model comparison 

JKL98ISR:


deepchecks/checks/performance/performance_report.py
```diff
@@ -97,14 +97,14 @@ def _performance_report(self, train_dataset: Dataset, test_dataset: Dataset, mod
         validate_model(test_dataset, model)
 
         task_type = task_type_check(model, train_dataset)
-        clasess = test_dataset.classes
+        clasess = train_dataset.classes
```
matanper, line 100: why did you change to train?

JKL98ISR: test can contain classes not in train

JKL98ISR: missing labels from train should do an error either way no?

JKL98ISR: not an error just bad score*

matanper (approved):

PR #551, opened 2022-01-09
Filter null labels before sampling for more accurate metrics

### LINKED ISSUE #547: Add additional model comparison checks
Currently, only model comparison check compares the performance of two models. 

Additional suggested model comparisons:

1. Change in unused features
2. Difference in computation time 
3. Difference in calibration

### LINKED ISSUE #549: Improve metric calculation with many missing labels
Currently, missing label filtering in metric-related checks is done at the moment of calculation, after sampling was already done. This could result in bad metrics when a large percent of labels are missing. 
Better practice to implement - first filter missing labels, then sample.

matanper:
Resolve #549


deepchecks/base/dataset.py
```diff
@@ -420,19 +420,25 @@ def copy(self: TDataset, new_data: pd.DataFrame) -> TDataset:
                    convert_datetime=False, max_categorical_ratio=self._max_categorical_ratio,
                    max_categories=self._max_categories, label_type=self.label_type)
 
-    def sample(self: TDataset, n_samples: int, replace: bool = False, random_state: t.Optional[int] = None) -> TDataset:
+    def sample(self: TDataset, n_samples: int, replace: bool = False, random_state: t.Optional[int] = None,
+               drop_na_label: bool = True) -> TDataset:
```
noamzbr, line 424: This should be true only when sampling in metric related checks, not in all checks

noamzbr (changes requested):

noamzbr (changes requested):
add to all performance related checks please

matanper:
> add to all performance related checks please

There are no others

noamzbr (approved):
lol right, approved

PR #578, opened 2022-01-10
StringLengthOutOfBounds: Add examples of outliers to result

### LINKED ISSUE #506: String Length Out Of Bounds should show outlier examples
Currently only tells you there are outlier samples and what length they are, but doesn't show you an example of these.

matanper:
resolve #506


docs/source/examples/checks/performance/plotly_question.ipynb
```diff
@@ -0,0 +1,1079 @@
+{
```
nirhutnik: What is this file?

nirhutnik (changes requested):

nirhutnik (approved):

PR #579, opened 2022-01-10
[issue-342] added docs link check CI/CD workflow

### LINKED ISSUE #342: Consider using Lychee to validate our links
@ItayGabbay we can take some inspiration from https://github.com/eyarz/pink-bunny-ears and use https://github.com/lycheeverse/lychee-action to constantly validate our links.

_Originally posted by @noamzbr in https://github.com/deepchecks/deepchecks/pull/340#pullrequestreview-841416537_

yromanyshyn:
added link checking CI/CD workflow 

resolves #342


ItayGabbay:
@yromanyshyn does it check internal links also?

yromanyshyn:
@ItayGabbay 

yes, it tries to resolve internal links to other parts of the doc relatively to the `--base` dir

.github/workflows/documentation.yaml
```diff
@@ -0,0 +1,100 @@
...
+        run: |
+          set -uxo pipefail
+          
+          curl \
+            -L --output lychee-v0.8.2-x86_64-unknown-linux-gnu.tar.gz \
+            https://github.com/lycheeverse/lychee/releases/download/v0.8.2/lychee-v0.8.2-x86_64-unknown-linux-gnu.tar.gz
+          
+          tar -xvzf lychee-v0.8.2-x86_64-unknown-linux-gnu.tar.gz
+
+          echo "$GITHUB_TOKEN"
```
ItayGabbay, line 79: Please remove this

.github/workflows/documentation.yaml
```diff
@@ -0,0 +1,100 @@
...
+
+          ./lychee \
+            README.rst FAQ.rst "./docs/build/html/**/*.html" \
+            --base ./docs/build/html \
```
ItayGabbay, line 83: Why are we checking this on the html and not on the rst? this means that every pr we need to run make docs before testing lychee?

ItayGabbay (changes requested):

.github/workflows/documentation.yaml
yromanyshyn: OMG, sorry, I have been testing it locally and forgot to remove this

.github/workflows/documentation.yaml
yromanyshyn: do you mean, why we are not checking docs source code?

good question, I also was thinking about it/ should we check docs source code or the result of the build.
I think the last one is better because a lot of stuff is expanded during the build, like interlinks between different parts of the documentation, etc.

I think we can cache a result of the docs build ('build' folder), and then use it during the next build, but this question needs additional exploration. But, it will make workflow more complicated - not sure if it is worth/ Maybe, let's run it only on the release?

yromanyshyn:
@ItayGabbay link-check failed, maintainability badge in the README.md leads to the 404 page

image

.github/workflows/documentation.yaml
ItayGabbay: Got it. in addition - can it also validate links in the README?

yromanyshyn: yes, it checks FAQ README and all generated html files

```
README.rst FAQ.rst "./docs/build/html/**/*.html"
```

ItayGabbay: We have some more readme (For example inside the examples/ folder, or under suites/. I think it's better to find all rst files (which are not under docs/) and validate them

yromanyshyn: yup, You are right, I changed the glob pattern to "./**/*.rst"

ItayGabbay (approved):

Issue #371, opened 2021-12-30
feature importance crashes if not supplied with a standard sklearn model


benisraeldan:
>TypeError: If no scoring is specified, the estimator passed should have a 'score' method. The estimator my_dumb_model() does not.

permutational importance assumes model has `fit` and `score` attributes. We need to check if it has these attributes.
If it does not have a score function, use a default from the assumed task type.

noamzbr:
@benisraeldan Just to clarify - this does work on the standard models we support right? (sklearn, lightgbm, catboost, xgboost)

PR #598, opened 2022-01-12
chose scorer based on score and fit on model in permutational feature importance

benisraeldan:
#### Reference Issues/PRs

Resolves: #371 

 Scorer can be set on the `model`, but can also be not set so we chose from our defaults based on the type of model


ItayGabbay:
I feel I can't review it well enough as I'm not so familiar with this part of the code. better @noamzbr @JKL98ISR @matanper will review that

deepchecks/utils/features.py
```diff
@@ -219,13 +220,22 @@ def _calc_importance(
         raise errors.DeepchecksTimeoutError('Permutation importance calculation was not projected to finish in'
                                             f' {permutation_importance_timeout} seconds.')
 
+    if hasattr(model, 'score'):
+        if hasattr(model, 'fit'):
+            scoring = None
```
matanper, line 225: can you explain why do you check for fit?

noamzbr: agree, we don't use fit anywhere

deepchecks/utils/features.py
```diff
@@ -219,13 +220,22 @@ def _calc_importance(
...
+
     r = permutation_importance(
         model,
         dataset_sample.features_columns,
         dataset_sample.label_col,
         n_repeats=n_repeats,
         random_state=random_state,
-        n_jobs=-1
+        n_jobs=-1,
+        scoring=scoring
```
matanper, line 238: if this is scorer from the model (model.score) it doesn't make sense we did the calculation of time on our default scorer

PR #610, opened 2022-01-13
Improve display

JKL98ISR:


deepchecks/base/display_suite.py
```diff
@@ -70,17 +91,34 @@ def display_suite_result(suite_name: str, results: List[Union[CheckResult, Check
...
             if result.have_conditions():
                 checks_with_conditions.append(result)
-            if result.have_display():
-                display_table.append(result)
-            else:
+                if result.have_display():
+                    checks_w_condition_display.append(result)
+            elif result.have_display():
+                checks_wo_conditions_display.append(result)
+            if not result.have_display():
```
matanper, line 121: no need the if here, just else

deepchecks/base/display_suite.py
```diff
@@ -70,17 +91,34 @@ def display_suite_result(suite_name: str, results: List[Union[CheckResult, Check
         unique_id = ''
     else:
         unique_id = get_random_string()
+    is_widgets = is_widgets_enabled()
+    if is_widgets:
```
matanper, line 95: IMO having many `if is_widgets` makes the code very hard to read. 
I'd prefer to have 2 functions, display_as_widget, display_as_html even in the price of some code duplication

deepchecks/base/check.py
```diff
@@ -85,41 +87,83 @@ def __init__(self, value, header: str = None, display: Any = None):
...
+                    else:
+                        display_html(check_html, raw=True)
+                        item.show()
+                    check_html = ''
                 elif callable(item):
                     try:
                         item()
-                        plt.show()
+                        if as_widget:
+                            plt_out = widgets.Output()
```
matanper, line 138: by the way, did you test it? I remember I tried something similar and couldn't make it work

deepchecks/base/check.py
```diff
@@ -85,41 +87,86 @@ def __init__(self, value, header: str = None, display: Any = None):
...
+                    check_html += item
                 elif isinstance(item, BaseFigure):
-                    item.show()
+                    if as_widget:
+                        box_children.append(widgets.HTML(check_html))
+                        box_children.append(go.FigureWidget(data=item))
+                    else:
+                        display_html(check_html, raw=True)
+                        item.show()
+                    check_html = ''
```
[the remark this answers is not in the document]
JKL98ISR: Yes, first 1 is that plotly as html doesn't work in widget. Second one is that it is much bigger

deepchecks/base/display_suite.py
JKL98ISR: It means any update to the code will be very hard

deepchecks/base/display_suite.py
JKL98ISR: we need it because maybe it doesn't have display

deepchecks/base/check.py
```diff
@@ -85,41 +87,86 @@ def __init__(self, value, header: str = None, display: Any = None):
...
+        if not self.display:
+            check_html += '<p><b>&#x2713;</b> Nothing found</p>'
         if unique_id:
-            display_html(f'<br><a href="#summary_{unique_id}" style="font-size: 14px">Go to top</a>', raw=True)
+            check_html += f'<br><a href="#summary_{unique_id}" style="font-size: 14px">Go to top</a>'
+        if as_widget:
+            box_children.append(widgets.HTML(check_html))
+            box.children = box_children
+            return box
+        display_html(check_html, raw=True)
```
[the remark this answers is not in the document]
JKL98ISR: it's not an html option, It's either live display or widget

deepchecks/base/check.py
```diff
@@ -85,41 +87,86 @@ def __init__(self, value, header: str = None, display: Any = None):
...
-            summary = next((s for s in docs.split('\n') if not re.match('^\\s*$', s)), '')
-            display_html(f'<p>{summary}</p>', raw=True)
+            summary = get_docs_summary(self.check)
+            check_html += f'<p>{summary}</p>'
         if self.conditions_results and show_conditions:
-            display_html('<h5>Conditions Summary</h5>', raw=True)
-            display_conditions_table(self, unique_id)
+            check_html += '<h5>Conditions Summary</h5>'
+            check_html += get_conditions_table_display(self, unique_id)
+            check_html += '<h5>Additional Outputs</h5>'
```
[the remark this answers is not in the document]
JKL98ISR: why?

deepchecks/base/check.py
```diff
@@ -85,41 +87,86 @@ def __init__(self, value, header: str = None, display: Any = None):
...
+                        display_html(check_html, raw=True)
+                        item.show()
+                    check_html = ''
                 elif callable(item):
                     try:
                         item()
-                        plt.show()
+                        if as_widget:
+                            plt_out = widgets.Output()
+                            with plt_out:
```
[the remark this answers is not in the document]
JKL98ISR: done

deepchecks/base/display_suite.py
matanper: I disagree, right now every update is very hard since the function is very complicated

JKL98ISR: I remove some if is_widgets

deepchecks/base/check.py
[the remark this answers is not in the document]
matanper: because if it's false you don't want to show this header

deepchecks/base/check.py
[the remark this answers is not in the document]
matanper: sounds weird it doesn't work. still I would change the way you work with the check_html (would do something an array containing str/widgets and some logic to append to the end where if it's HTML it adds, else add new item)

deepchecks/base/check.py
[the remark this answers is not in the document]
JKL98ISR: ok

deepchecks/base/check.py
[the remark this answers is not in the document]
JKL98ISR: I don't understand

deepchecks/base/display_suite.py
JKL98ISR: there aren't many differences in the logic, I don't think we should want to split it

matanper: @ItayGabbay do you have opinion?

deepchecks/base/check.py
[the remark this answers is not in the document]
matanper: my bottom line is that this hybrid of "add widget or live display" is very messy. I'd prefer to separate it into steps, first create all necessary objects, and in the end display/return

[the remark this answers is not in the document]
JKL98ISR: I can't group live displays, only thing you can do is split the function completely

deepchecks/base/display_suite.py
ItayGabbay: I think we should separate it to functions

JKL98ISR: done

deepchecks/base/display_pandas.py
```diff
@@ -12,30 +12,25 @@
...
 import warnings
 
-from IPython.core.display import display_html
 import pandas as pd
 from pandas.io.formats.style import Styler
 
-from . import check  # pylint: disable=unused-import
+from deepchecks.utils.strings import get_docs_summary, get_ellipsis
 
+from . import check  # pylint: disable=unused-import
```
matanper, line 20: why that line is needed?

deepchecks/base/display_suite.py
```diff
@@ -16,19 +16,40 @@
...
 
 
 __all__ = ['display_suite_result', 'ProgressBar']
 
 
+def _get_check_widget(check_res: CheckResult, unique_id: str) -> widgets.VBox:
+    return check_res.display_check(unique_id=unique_id, as_widget=True)
+
+
+def _add_widget_classes(widget: widgets.HTML):
```
matanper, line 38: maybe explain what are this classes?

deepchecks/base/display_suite.py
```diff
@@ -61,7 +82,130 @@ def get_display_exists_icon(exists: bool):
...
+    others_tab = widgets.VBox()
+    tab.children = [condition_tab, checks_wo_tab, others_tab]
+    tab.set_title(0, 'Checks With Conditions')
+    tab.set_title(1, 'Checks Without Conditions')
+    tab.set_title(2, 'Checks Without Output')
+    display_html('<style>.jupyter-widgets.widget-tab > .p-TabBar .p-TabBar-tab {flex: 0 1 auto}</style>',
+                 raw=True)
+
+    if checks_with_conditions:
+        cond_html_h2 = '<h2>Conditions Summary</h2>'
```
matanper, line 106: maybe put all the shared fixed texts outside in a constants

deepchecks/base/display_pandas.py
JKL98ISR: no sure why it's showing that I added that, but it's to be able to use CheckResult typing without circular input - I think

deepchecks/base/display_suite.py
JKL98ISR: done

deepchecks/base/display_suite.py
JKL98ISR: done

ItayGabbay (approved):
I really think the `display_check` can be simplified. Worth thinking about that - currently this function is complicated and not so maintainable. However approving for now

PR #614, opened 2022-01-13
[issue-572] automatic generation of the checks api references

### LINKED ISSUE #572: Fix API reference to load components recursively

yromanyshyn:
added possibility to automatically generate checks API references

resolves #572 

@ItayGabbay 

I did not manage to make sphinx `autosummary` 
recursively generate checks API references and 
to keep them look like it is now

When the`:recursive:` option is added to the `autosummary` 
directive, it follows directory (package) structure and as a result, 
it creates additional sublevels in the toctree (screenshot 1)

I was exploring ways to tweak autosummary behavior 
but it cannot be done easily or without some unobvious hacks

So, I decided to write a small script that will generate rst 
files for checks automatically and with similar structure and 
content that we wrote manually

I do not like this solution/ but it is simple whereas the hacks 
that I tried did not work or were unobvious and required understanding 
of how `autosummary` works underhood 
 
note: we generate other parts of the API Reference (suites, utils)  
recursively (with 'recursive' option)

**screenshot 1**
Screenshot from 2022-01-13 12-23-11


yromanyshyn:
the problem can be solved by using the dev branch of the sphinx, closing (will recreate)

PR #618, opened 2022-01-14
[issue-572] automatic api references generation for the checks

### LINKED ISSUE #572: Fix API reference to load components recursively

yromanyshyn:
main change: 

now, sphinx generates API documentation for the checks automatically (**partially**). If someone adds a new check type to the existing checks packages (distribution, integrity, ...), documentation for it will be generated automatically. But if you added a new module to the `deepchecks.checks` package, documentation for it will not be generated, you will need to add that new module to the `autosummary` directive within `api/checks.rst` file manually. This limitation is caused by the fact that `deepchecks.checks` package imports all existing checks, as result we cannot pass it directly to the `autosummary` directive, because it will build toctree with all checks on the same level.

list of other changes:
- fixed navigation sidebars on the `examples/use-cases/use-cases` page
- fixed some rough edges within autosummary templates
- refactoring

resolves #572


ItayGabbay:
@yromanyshyn the only issue now is that the links from the notebooks to the API reference are missing.

yromanyshyn:
@ItayGabbay 

>@yromanyshyn the only issue now is that the links from the notebooks to the API reference are missing.

do you mean those links that are generated within `nbsphinx_prolog`? (yup, they are not rendered/)

I think I have not seen any manually written links to the API references/

ItayGabbay:
@yromanyshyn Yes, I'm talking about the `nbsphinx_prolog`

yromanyshyn:
@ItayGabbay I fixed that

ItayGabbay (approved):

PR #640, opened 2022-01-17
[issue-561] error handling

### LINKED ISSUE #561: Update error handling and display
For all errors that are related to "metadata not relevant" - e.g. not a boosting model, no date/index column, etc. - make it a different type of error and update suite text to be something that doesn't sound like an error

For example, instead of - `DeepchecksValueError: Expected model to be a type from ['regression'], but received model of type: multiclass`
maybe something like: Check relevant for models of type ['regression'], received model of type 'multiclass'.

yromanyshyn:
(intended to resolve #561, see also PR #242)

This issue actually brings up a more broad concern about how we handle exceptions and dataset validation.

In general
----------------

Our `Dataset` class is loosely defined and has too many possible states. We can define a dataset with features but without a label, with a label but without features, or even entirely empty, without features and label.

This "shapelessness" of the dataset class has a negative effect. Whenever and wherever we want to do something with a dataset instance, first we need to verify presence of the features or a label. For example, if we have a check that requires a dataset with features and a label, and uses some public function from another `deepchecks` module that also requires a dataset with features and a label, then we need to do validation twice, separately in the check and separately in the public function. 

We cannot rely on the fact that whoever will use that public function, will always provide a valid instance of the dataset, and we also want to raise an error with a more specific message in the check, instead of allowing the public function to handle this.

doing the same thing twice (possible even more times) is something incorrect/

I think we should impose more restrictions upon the `Dataset` type, but I am not going to speak about it here, instead, I want to speak about how we handling this currently, and how we should change it.

We have tried to solve this problem with help of a bunch of` Dataset.validate_*` methods that we use almost everywhere when we need to validate a dataset instance. Honestly, it did not facilitate not solve the problem.
1. we just replaced two lines of code (`if ... the raise ...`) with one
2. those methods raise errors with messages that are relative only to the checks

The only reason to use those methods is to unify error messages within some scope, and that is useful only within checks scope.

What I have done and what I suggest
---------------------------------------------------

**point 1**

Added `DatasetValidationError` and `ModelValidationError` error types, that should be 
used in cases when dataset instance or model instance did not meet imposed requirements.

(to some degree it is a special case of the `DeepchecksValueError`/)

**point 2**

Removed `validate_*` methods from the `Dataset` class.

**point 3**

Added private dataset validation methods to the `BaseCheck` class that are used only
to unify error messages across all checks.

**point 4**

Added next methods to the `Dataset` class:

- `def ensure_not_empty_dataset(cls, obj: t.Any, *, cast: bool = False) -> 'Dataset': ...`
- `def datasets_share_features(cls, datasets: t.List['Dataset']) -> bool: ...`
- `def datasets_share_categorical_features(cls, datasets: t.List['Dataset']) -> bool: ...`
- `def datasets_share_label(cls, datasets: t.List['Dataset']) -> bool: ...` 

**point 5**

Rephrased some error messages by following @shir22  advice

**Note that PR is in draft state.**

**TODO:**
- [x] fix tests
- [x] add tests for new `Dataset` methods
- [x] fix comments
- [x] fix codestyle
- [x] rephrase error messages by folowing @shir22 advice (where it is possible)

**To think about:**
- should we also compare label/features dtypes within `Dataset.datasets_share_label`, `Dataset.datasets_share_features`?


deepchecks/base/dataset.py
```diff
@@ -781,109 +783,213 @@ def select(
...
+            AssertionError:
+                'datasets' parameter is not instance of a list;
+                'datasets' is empty or contains only one dataset;
         """
-        Dataset.validate_dataset(other)
-        if sorted(self.features) == sorted(other.features):
-            return list(self.features)
-        else:
-            raise DeepchecksValueError('Check requires datasets to share the same features')
+        assert isinstance(datasets, list)
```
matanper, line 824: not sure I like the use of assert, what happens if it fails?

matanper (commented):
I agree with your points. not sure about the classmethods on base check, a bit weird. but overall good

deepchecks/base/dataset.py
yromanyshyn: it is common in the python world to use assert to verify that input parameters of the function satisfy its contract (signature). Ff `datasets` is not a list then `AssertionError` will be raised which means that function is used improperly 

I will add a message to that assert to make it cleaner what happened

deepchecks/base/check.py
```diff
@@ -332,6 +336,67 @@ def name(cls):
...
+                'inferred incorrectly as categorical features. To fix this, manually edit the '
+                'categorical features using Dataset(cat_features=<list_of_features>'
+            )
+        return datasets[0].cat_features
+
+    @classmethod
+    def _datasets_share_label(cls, datasets: List['Dataset']) -> Hashable:
+        """TODO: add coments"""
+        if Dataset.datasets_share_label(datasets) is False:
+            raise DatasetValidationError('Check requires datasets to have and to share the same label')
```
shir22, line 366:
```suggestion
            raise DatasetValidationError('Check requires Datasets to have and to share the same label')
```

deepchecks/base/check.py
```diff
@@ -332,6 +336,67 @@ def name(cls):
...
+        return cast(Hashable, datasets[0].label_name)
+
+    @classmethod
+    def _dataset_has_label(cls, dataset: Dataset) -> pd.Series:
+        """TODO: add comments"""
+        if dataset.label_col is None:
+            raise DatasetValidationError('Datasets without label are irrelevant to the check')
```
shir22, line 373:
```suggestion
            raise DatasetValidationError('Check is irrelevant for Datasets without label')
```

deepchecks/base/check.py
```diff
@@ -332,6 +336,67 @@ def name(cls):
...
+        return dataset.label_col
+
+    @classmethod
+    def _dataset_has_features(cls, dataset: Dataset) -> pd.DataFrame:
+        """TODO: add comments"""
+        if (
+            dataset.features_columns is None
+            or len(dataset.features_columns.columns) == 0
+        ):
+            raise DatasetValidationError('Datasets without features are irrelevant to the check')
```
shir22, line 383:
```suggestion
            raise DatasetValidationError('Check is irrelevant for Datasets without features')
```

deepchecks/base/check.py
```diff
@@ -332,6 +336,67 @@ def name(cls):
...
+        return dataset.features_columns
+
+    @classmethod
+    def _dataset_has_date(cls, dataset: Dataset) -> pd.Series:
+        """TODO: add comments"""
+        if dataset.datetime_col is None:
+            raise DatasetValidationError('Datasets without datetime column are irrelevant to the check')
```
shir22, line 390:
```suggestion
            raise DatasetValidationError('Check is irrelevant for Datasets without datetime column ')
```

deepchecks/base/check.py
```diff
@@ -332,6 +336,67 @@ def name(cls):
...
+        return dataset.datetime_col
+
+    @classmethod
+    def _dataset_has_index(cls, dataset: Dataset) -> pd.Series:
+        """TODO: add comments"""
+        if dataset.index_col is None:
+            raise DatasetValidationError('Datasets without index are irrelevant to the check')
```
shir22, line 397:
```suggestion
            raise DatasetValidationError('Check is irrelevant for Datasets without an index')
```

deepchecks/base/check.py
```diff
@@ -332,6 +336,67 @@ def name(cls):
         """Name of class in split camel case."""
         return split_camel_case(cls.__name__)
 
+    # NOTE: next set of private functions exists to unify error messages across all checks
```
shir22, line 339: General note - 
One more error that is currently a "DeepchecksValueError" is in the BoostingOverfit check... 
(if model is not of type boosting. not sure where this should be but it also belongs to these types of "irrelevant data/model for check")

ItayGabbay (commented):
I agree with your comments.
One important note: We currently add computer vision checks as a namespaced package. In addition, I moved all current checks and suits to be in a `tabular` namespaced packages, and some the core functionality and classes (including basecheck, basesuites, etc.) to a `base` namespaced package. (This is currently a draft).

I wonder where the validations should happen? I don't think it should be inside the base subpackage since they are specific to tabular datasets and not related to CV.

deepchecks/base/dataset.py
yromanyshyn: oh, also regarding this

> ..not sure about the classmethods ..

they do not depend upon instance variables (state), so I decided to declare thems as classmethods

yromanyshyn:
@ItayGabbay 

> wonder where the validations should happen? I don't think it should be inside the base subpackage since they are specific to tabular datasets and not related to CV.

oh/ yup you are right
then it will be better to put those methods into separate modules, like
- `deepchecks.tabular.checks._dataset_validation`
- `deepchecks.vision.checks._dataset_validation`

something like that

deepchecks/base/check.py
```diff
@@ -332,6 +338,157 @@ def name(cls):
...
+
+        Raises:
+            DatasetValidationError: if dataset does not have index;
+        """
+        if dataset.index_col is None:
+            raise DatasetValidationError('Check is irrelevant for Datasets without an index')
+        return dataset.index_col
+
+    @staticmethod
+    def _verify_model_type(
```
JKL98ISR, line 471: why it should be in the Check class? I think will be better in utild or dataset class

yromanyshyn: - it is used only within checks to unify error message
- it raises  an exception with check related message

ItayGabbay:
image

deepchecks/base/check.py
```diff
@@ -9,30 +9,36 @@
 # ----------------------------------------------------------------------------
 #
 """Module containing all the base classes for checks."""
-# pylint: disable=broad-except
+# pylint: disable=broad-except,inconsistent-quotes
```
benisraeldan, line 12: why disable this lint?

yromanyshyn:
@ItayGabbay did not get you? was not it expected? (error name is not show only for `DatasetValidationError`, `ModelValidationError` and `DeepchecksProcessError`)

ItayGabbay:
> @ItayGabbay did not get you? was not it expected? (error name is not show only for `DatasetValidationError`, `ModelValidationError` and `DeepchecksProcessError`)

It's exactly like how I expected. I attached the screenshot so everyone will see the output :) 
great work.

deepchecks/checks/methodology/identifier_leakage.py
```diff
@@ -51,14 +52,17 @@ def run(self, dataset: Dataset, model=None) -> CheckResult:
...
+        self._dataset_has_label(dataset)
 
+        ppscore_params = ppscore_params or {}
         relevant_columns = list(filter(None, [dataset.datetime_name, dataset.index_name, dataset.label_name]))
 
         if len(relevant_columns) == 1:
-            raise DeepchecksValueError('Dataset needs to have a date or index column.')
+            raise DatasetValidationError(
+                'Datasets without index or date column '
+                'are irrelevant to the current check'
```
shir22, line 64:
```suggestion
                'Check is irrelevant for '
                'Datasets without index or date column'
```

deepchecks/checks/distribution/trust_score_comparison.py
```diff
@@ -72,35 +72,40 @@ def run(self, train_dataset, test_dataset, model=None) -> CheckResult:
...
-        train_dataset.validate_shared_features(test_dataset)
+        train_dataset = Dataset.ensure_not_empty_dataset(train_dataset)
+        train_label = self._dataset_has_label(train_dataset)
+
+        features_list = self._datasets_share_features([train_dataset, test_dataset])
+        label_name = self._datasets_share_label([train_dataset, test_dataset])
 
         if test_dataset.n_samples < self.min_test_samples:
             msg = ('Number of samples in test dataset have not passed the minimum. you can change '
                    'minimum samples needed to run with parameter "min_test_samples"')
```
shir22, line 89:
```suggestion
            msg = ('Number of samples in test dataset has not passed the minimum. You can change '
                   'the minimum number of samples required for the check to run with the parameter "min_test_samples"')
```

deepchecks/base/check.py
yromanyshyn: I wanted to disable it for two specific lines, but when I added directive at the end of the lines it caused `line too long` error.
Yup, you are right) it is better to use it only for that lines, fixed

deepchecks/checks/methodology/boosting_overfit.py
```diff
@@ -61,7 +61,7 @@ def predict_proba(self, x):
         elif self.model_class == 'CatBoostClassifier':
             return self.model.predict_proba(x, ntree_end=self.step)
         else:
-            raise DeepchecksValueError(f'Unsupported model of type: {self.model_class}')
+            raise ModelValidationError(f'Unsupported model of type: {self.model_class}')
```
shir22, line 64: This error message (and also in lines 77 and 93) should be
"Check is relevant for Boosting models of type <list_of_supported model types>, but received model of type <model_class>"

tests/checks/performance/calibration_score_test.py
```diff
@@ -11,29 +11,39 @@
...
     # Assert
     train, _, clf = diabetes_split_dataset_and_model
-    assert_that(calling(CalibrationScore().run).with_args(train, clf),
-                raises(DeepchecksValueError, r'Expected model to be a type from'
-                                           r' \[\'multiclass\', \'binary\'\], but received model of type: regression'))
+    assert_that(
+        calling(CalibrationScore().run).with_args(train, clf),
+        raises(
+            ModelValidationError,
+            r'Check relevant for models of type \[\'multiclass\', \'binary\'\], '
```
shir22, line 43:
```suggestion
            r'Check is relevant for models of type \[\'multiclass\', \'binary\'\], '
```

deepchecks/base/check.py
```diff
@@ -332,6 +338,157 @@ def name(cls):
...
+        Returns:
+            ModelType: type of the provided model
+
+        Raises:
+            ModelValidationError: if unexpected model type is provided;
+        """
+        task_type = task_type_check(model, dataset)
+        if task_type not in expected_types:
+            raise ModelValidationError(
+                f'Check relevant for models of type {[e.value.lower() for e in expected_types]}, '
```
shir22, line 487:
```suggestion
                f'Check is relevant for models of type {[e.value.lower() for e in expected_types]}, '
```

tests/checks/performance/regression_error_distribution_test.py
```diff
@@ -14,30 +14,37 @@
...
     # Assert
     _, test, clf = iris_split_dataset_and_model
-    assert_that(calling(RegressionErrorDistribution().run).with_args(test, clf),
-                raises(DeepchecksValueError, r'Expected model to be a type from'
-                                           r' \[\'regression\'\], but received model of type: multiclass'))
+    assert_that(
+        calling(RegressionErrorDistribution().run).with_args(test, clf),
+        raises(
+            ModelValidationError, 
+            r'Check relevant for models of type \[\'regression\'\], '
```
shir22, line 45:
```suggestion
            r'Check is relevant for models of type \[\'regression\'\], '
```

tests/checks/performance/roc_report_test.py
```diff
@@ -17,30 +17,37 @@
...
     # Assert
     train, _, clf = diabetes_split_dataset_and_model
-    assert_that(calling(RocReport().run).with_args(train, clf),
-                raises(DeepchecksValueError, r'Expected model to be a type from'
-                                           r' \[\'multiclass\', \'binary\'\], but received model of type: regression'))
+    assert_that(
+        calling(RocReport().run).with_args(train, clf),
+        raises(
+            ModelValidationError, 
+            r'Check relevant for models of type '
```
shir22, line 48:
```suggestion
            r'Check is relevant for models of type '
```

shir22 (changes requested):

shir22:
and +1 to Itay's comment, completely agree
the print screen did enable to see it in action (and therefore my minor phrasing comments), anyway looks great!

deepchecks/checks/methodology/identifier_leakage.py
yromanyshyn: fixed

deepchecks/checks/distribution/trust_score_comparison.py
yromanyshyn: fixed

deepchecks/checks/methodology/boosting_overfit.py
yromanyshyn: fixed

tests/checks/performance/calibration_score_test.py
yromanyshyn: fixed

deepchecks/base/check.py
yromanyshyn: fixed

tests/checks/performance/confusion_matrix_report_test.py
```diff
@@ -12,29 +12,38 @@
...
     # Assert
     train, _, clf = diabetes_split_dataset_and_model
-    assert_that(calling(ConfusionMatrixReport().run).with_args(train, clf),
-                raises(DeepchecksValueError, r'Expected model to be a type from'
-                                           r' \[\'multiclass\', \'binary\'\], but received model of type: regression'))
+    assert_that(
+        calling(ConfusionMatrixReport().run).with_args(train, clf),
+        raises(
+            ModelValidationError,
+            r'Check relevant for models of type \[\'multiclass\', \'binary\'\], '
```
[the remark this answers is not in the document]
yromanyshyn: fixed

tests/checks/performance/regression_systematic_error_test.py
```diff
@@ -9,38 +9,41 @@
...
     # Assert
     _, test, clf = iris_split_dataset_and_model
-    assert_that(calling(RegressionSystematicError().run).with_args(test, clf),
-                raises(DeepchecksValueError, r'Expected model to be a type from'
-                                             r' \[\'regression\'\], but received model of type: multiclass'))
+    assert_that(
+        calling(RegressionSystematicError().run).with_args(test, clf),
+        raises(
+            ModelValidationError, 
+            r'Check relevant for models of type '
```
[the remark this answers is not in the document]
yromanyshyn: fixed

tests/checks/performance/roc_report_test.py
yromanyshyn: fixed

yromanyshyn:
@shir22  I fixed your remarks, thanks!

shir22 (approved):

ItayGabbay (approved):

PR #643, opened 2022-01-17
Add supported models to user guide documentation

### LINKED ISSUE #566: Supported models docs - user guide
Add a "User Guide" about supported models, how we infer model type and what exactly needs to be implemented for a custom model.
including explanation about feature importance and how it's calculated.

benisraeldan:
#### Reference Issues/PRs

Resolves #566 

Adds explination of supported models to user guide


docs/source/user-guide/supported_models.rst
```diff
@@ -0,0 +1,56 @@
+.. _supported_models:
+
+====================
+Supported Models
+====================
+
+Deepchecks supports a minmal version of the sklearn api.
```
ItayGabbay, line 7:
```suggestion
Deepchecks supports a minimal version of the sklearn api.
```

docs/source/user-guide/supported_models.rst
```diff
@@ -0,0 +1,56 @@
...
+...     def predict(X: pd.DataFrame) -> pd.Series:
+...         ...
+...     @property
+...     def feature_importances_(self):
+...         ...
+
+Check Specific APIs
+-------------------
+
+Some checks require specific apis to run. For example, ``BoostingOverfit`` requires model to be a `Boosting model`.
```
ItayGabbay, line 55:
```suggestion
Some checks require specific apis to run. For example, ``BoostingOverfit`` requires model to be a `Boosting model`. Examples to such model include XGBoost, LightGBM, CatBoost or any other gbm implementation.
```

ItayGabbay, line 55: Do we want to add which properties the model needs to be considered as a compatible model to the boosting overfit check?

docs/source/user-guide/supported_models.rst
ItayGabbay, line 7: I would rephrase it a bit different:

Similar to scikit-learn API, a model will be compatible with deepchecks' checks if it implements some minimal functionality. Below are the minimal requirements for each task type

docs/source/user-guide/supported_models.rst
```diff
@@ -0,0 +1,56 @@
...
+...     def predict(X: pd.DataFrame) -> pd.Series:
+...         ...
+
+Classification
+--------------
+
+For classification models, we require also the ``predict_proba`` function:
+
+>>> class simple_classification_model:
+...     def predict(X: pd.DataFrame) -> pd.Series:
```
ItayGabbay, line 30: Maybe we should include the return format?

docs/source/user-guide/supported_models.rst
```diff
@@ -0,0 +1,56 @@
...
+...         ...
+
+
+Other Helpful APIs
+==================
+
+Feature importance
+-------------------
+
+Deepchecks can calculate feature importance using ``permutational_importance``.
```
ItayGabbay, line 42: I think some elaborations on permutational importance are missing. Maybe we can send them to sklearn docs, or document here.

docs/source/user-guide/supported_models.rst
```diff
@@ -0,0 +1,56 @@
...
+But it can support builtin feature importance property: ``feature_importances_`` or ``coef_`` for linear model:
+
+>>> class simple_importance_model:
+...     def predict(X: pd.DataFrame) -> pd.Series:
+...         ...
+...     @property
+...     def feature_importances_(self):
+...         ...
```
ItayGabbay, line 50: Also here, I think an output format is important

ItayGabbay (changes requested):

docs/source/user-guide/supported_models.rst
```diff
@@ -0,0 +1,56 @@
+.. _supported_models:
```
shir22, line 1: just a note - no need for a tag since documents can be directly referred to using a "doc" link... (anyway, no problem for adding either)

docs/source/user-guide/supported_models.rst
shir22: +1

docs/source/user-guide/supported_models.rst
```diff
@@ -0,0 +1,56 @@
...
+
+Required API
+=============
+
+Regression
+----------
+
+All that is necessary for for regression models is the ``predict`` function.
+
+Example of an simple model:
```
shir22, line 18:
```suggestion
Example of a simple model:
```

docs/source/user-guide/supported_models.rst
```diff
@@ -0,0 +1,56 @@
...
+Deepchecks supports a minmal version of the sklearn api.
+
+
+Required API
+=============
+
+Regression
+----------
+
+All that is necessary for for regression models is the ``predict`` function.
```
shir22, line 16:
```suggestion
All that is necessary for a regression models is the ``predict`` function.
```

docs/source/user-guide/supported_models.rst
shir22: indeed

want to note that these examples are a bit confusing as it seems that they should all receive a dataframe and return a series, while this isn't necessarily the (only) standard

docs/source/user-guide/supported_models.rst
shir22: it's also confusing that it looks like a "term". if it's a link to a specific function then you can link here to it's API reference. otherwise it will look like something that people want (and try to) click but it leads them no where and remains unclear

docs/source/user-guide/supported_models.rst
shir22: Same comment here - BoostingOverfit - better to make it a link to the API reference

docs/source/user-guide/supported_models.rst
```diff
@@ -0,0 +1,56 @@
+.. _supported_models:
+
+====================
+Supported Models
+====================
+
+Deepchecks supports a minmal version of the sklearn api.
+
```
shir22, line 8: This whole part should be much more elaborate...
a sentence about motivation (what happens if you don't supply us with a compatible model? then we can't run all checks that require predictions - e.g. all performance related etc...)
Examples of common models that support the scikit-learn api (all scikit learn models obviously, but also other models that we suppport as-is... this should be the first thing - so the user quickly understands "what are the straight forward things we support easily", without needing to go into the nitty-gritty of the API etc.)

shir22 (changes requested):

docs/source/user-guide/supported_models.rst
ItayGabbay: +1

shir22: Hi, I see you updated here, but I still don't feel this is "user facing".
The user comes here to try and understand:
Is my model supported?
The first paragraph should lead him as quickly as possible to this answer...

docs/source/user-guide/supported_models.rst
```diff
@@ -0,0 +1,65 @@
...
+...         ...
+
+
+Other Helpful APIs
+==================
+
+Feature importance
+-------------------
+
+Deepchecks can calculate feature importance using `sklearn permutational_importance <https://scikit-learn.org/stable/modules/permutation_importance.html>`_.
```
shir22, line 51: will this link open in a new tab as required for external links?

docs/source/user-guide/supported_models.rst
```diff
@@ -0,0 +1,65 @@
...
+Below are the minimal requirements for each task type:
+
+Required API
+=============
+
+Regression
+----------
+
+All that is necessary for a regression model is the ``predict`` function.
+The predict function should expect an `array-like` or `sparse matrix` of shape ``(n_samples, n_features)``
```
shir22, line 19: consider linking to array-like:
https://scikit-learn.org/stable/glossary.html#term-array-like

docs/source/user-guide/supported_models.rst
```diff
@@ -0,0 +1,65 @@
...
+Example of a simple model:
+
+>>> class simple_regression_model:
+...     def predict(X: pd.DataFrame) -> pd.Series:
+...         ...
+
+Classification
+--------------
+
+For classification models, we require also the ``predict_proba`` function.
```
shir22, line 32: Do we require "also" or only?
I'm thinking if there is a way to put these two together (seems as if it's not that convenient to read... cause if I have a classification model, then I jump here, but then it says "we require also", so it doesn't make much sense...)

docs/source/user-guide/supported_models.rst
```diff
@@ -0,0 +1,65 @@
...
+The predict function should expect an `array-like` or `sparse matrix` of shape ``(n_samples, n_features)``
+and is expected to return an `ndarray` of shape ``(n_samples,n_classes)``, an array containing the predicted probability of each class per sample.
+
+
+
+>>> class simple_classification_model:
```
shir22, line 38: and then the examples can be one after the other.... the regression and the classification
btw consider using a different signature, for the clarity that both are acceptable (e.g. with numpy array)

docs/source/user-guide/supported_models.rst
shir22, line 51: The explanation is good, but I'm missing - what is the default behavior...
"it can  do this and it can support that - but what actually happens"?

and - where is the feature importance used?
and how can I get it if I want?

docs/source/user-guide/supported_models.rst
```diff
@@ -0,0 +1,65 @@
...
+...         ...
+...     @property
+...     def feature_importances_(self):
+...         ...
+
+Check Specific APIs
+-------------------
+
+Some checks require specific apis to run. For example, :doc:`BoostingOverfit </api/base/generated/deepchecks.checks.methodology.boosting_overfit.BoostingOverfit>`
+requires model to be a `Boosting model`. Examples to such model include XGBoost, LightGBM, CatBoost or any other gbm implementation.
```
shir22, line 65: what is the "specific api" in this case?
e.g. if I have a custom boosting model, what should be implemented?

shir22 (changes requested):

docs/source/user-guide/supported_models.rst
benisraeldan: No, but other links don't either, we should probably change it in the template html/js

ItayGabbay: @benisraeldan you can use html tags inside rst. so just use 
<a href=<link> target=_blank></a>

shir22 (approved):

ItayGabbay (approved):

PR #664, opened 2022-01-18
Fix distribution plots outliers

### LINKED ISSUE #663: [FEAT] Distribution plots should handle outliers
When we have numeric columns with outliers, the distribution plots don't cut them out, so it is displayed with a very long tail which makes it harder to look at the interesting part.

Solution:\
Should handle this in the global function we use for the distribution plots. possible to totally cut off the tail, or just "hide" it and allow the user to still scroll there.

matanper:
Fix density plot to work on data with a single value.
Add cut-off of outliers in distribution plots.
Change trust score comparison to use the global distribution plot function.

resolve #663


shir22 (approved):
change default of quantile_cut as discussed :)

deepchecks/utils/distribution/plot.py
```diff
@@ -133,7 +133,13 @@ def feature_distribution_traces(train_column,
             min(np.quantile(train_column, quantile_cut), np.quantile(test_column, quantile_cut)),
             max(np.quantile(train_column, 1 - quantile_cut), np.quantile(test_column, 1 - quantile_cut))
         )
-        xs = np.linspace(x_range[0], x_range[1], 40)
+        # Heuristically take points on x-axis to show on the plot
+        # The intuition is the graph will look "smooth" wherever we will zoom it
+        xs = sorted(np.concatenate((
```
shir22, line 138: don't need to also drop duplicates here?

ItayGabbay (approved):

noamzbr (approved):

PR #665, opened 2022-01-18
added option to save checks as json

JKL98ISR:


deepchecks/base/check.py
```diff
@@ -165,6 +189,72 @@ def display_check(self, unique_id: str = None, as_widget: bool = False,
...
 
+    def _display_to_json(self):
+        displays = []
+        old_backend = matplotlib.get_backend()
+        displays.append(('header', f'<h4>{self.get_header()}</h4>'))
+        displays.append(('docs_header', f'<p>{get_docs_summary(self.check)}</p>'))
+        if self.conditions_results:
+            displays.append(('conditions_header', _CONDITIONS_HEADER))
+            cond_df = get_conditions_table(self)
+            if isinstance(cond_df, Styler):
```
matanper, line 198: `get_conditions_table` doesn't return always same type?

deepchecks/base/check.py
```diff
@@ -165,6 +189,72 @@ def display_check(self, unique_id: str = None, as_widget: bool = False,
...
+        check_name = self.check.name()
+        parameters = self.check.params()
+        result_json = {'name': check_name, 'parameters': parameters}
+        
+        if isinstance(self.value, pd.DataFrame):
+            result_json['value'] = self.value.to_json()
+        elif isinstance(self.value, np.ndarray):
+            result_json['value'] = self.value.tolist()
+        else:
+            result_json['value'] = self.value
```
matanper, line 234: what will happen with dict that contains dataframe/ndarray?

deepchecks/base/check.py
```diff
@@ -165,6 +189,72 @@ def display_check(self, unique_id: str = None, as_widget: bool = False,
...
+            result_json['display'] = display_json
+        return jsonpickle.dumps(result_json)
+
+    @staticmethod
+    def display_from_json(json_data):
+        json_data = jsonpickle.loads(json_data)
+        if json_data.get('display') is None:
+            return
+        for display_type, value in json_data['display']:
+            if display_type in ['html', 'plt'] or 'header' in display_type:
```
matanper: why not `display_type in ['html', 'plt', 'header']`?

deepchecks/base/check.py
```diff
@@ -165,6 +189,72 @@ def display_check(self, unique_id: str = None, as_widget: bool = False,
...
+                display_html(value, raw=True)
+            elif display_type in ['conditions', 'dataframe']:
+                df: pd.DataFrame = pd.read_json(value, orient='records')
+                display_html(dataframe_to_html(df), raw=True)
+            elif display_type == 'plotly':
+                plotly_json = io.StringIO(value)
+                plotly.io.read_json(plotly_json).show()
```
matanper, line 253: add another "else: raise exception", bad practice to have elif without else

deepchecks/base/check.py
```diff
@@ -165,6 +189,72 @@ def display_check(self, unique_id: str = None, as_widget: bool = False,
...
+            result_json['value'] = self.value.tolist()
+        else:
+            result_json['value'] = self.value
+        if with_display:
+            display_json = self._display_to_json()
+            result_json['display'] = display_json
+        return jsonpickle.dumps(result_json)
+
+    @staticmethod
+    def display_from_json(json_data):
```
matanper, line 241: so we only support display from json, but not reconstructing the real object?

deepchecks/base/check.py
JKL98ISR: jsonpickle will handle it, though now it isn't happening. I can't go into dicts recursivly

deepchecks/base/check.py
JKL98ISR: yes

ItayGabbay (approved):

PR #695, opened 2022-01-19
Fix Train Test Samples Mix

### LINKED ISSUE #659: [BUG] Train Test Samples Mix error in phishing notebook
**Describe the bug**
In the url phishing notebook, when running the suite produced by train_test_validation the Train Test Sample Mix check raises an error:
`TypeError: int() argument must be a string, a bytes-like object or a number, not 'Timestamp'`

**To Reproduce**
Run the following cell in the url phishing notebook:
`vsuite.run(model=logreg, train_dataset=ds_train, test_dataset=ds_test)`

**Expected behavior**
Check should run.
After fixing this bug, rerun the phishing notebook and commit the result as part of the PR.

**Screenshots**
image

 - OS: Mac
 - Python Version: ?
 - Deepchecks Version: main (0.3.0)

matanper:
resolve #659 

1. Fix limitation dataframe index must be numeric
2. Fix bug that the check expected dataframe index to be unique
3. Update result to return the duplicates dataframe


ItayGabbay:
Do we want this for 0.3?

noamzbr (approved):
Approved, though I would like to check that the example notebook remained the same. 

And yes, why not can add to 0.3.0

PR #700, opened 2022-01-19
Add first draft check context

matanper:


deepchecks/base/dataset.py
```diff
@@ -783,7 +783,7 @@ def ensure_not_empty_dataset(cls, obj: t.Any, *, cast: bool = False) -> 'Dataset
                 )
 
     @classmethod
-    def datasets_share_features(cls, datasets: t.List['Dataset']) -> bool:
+    def datasets_share_features(cls, *datasets: t.List['Dataset']) -> bool:
```
yromanyshyn, line 786: change `datasets` type from `List[Dataset]` to `Dataset`, currently it means `Iterable[List[Dataset]]`

deepchecks/base/check_context.py
```diff
@@ -0,0 +1,181 @@
...
+            user_scorers = self._user_non_avg_scorers if multiclass_non_avg else self._user_scorers
+            if user_scorers:
+                scorer_name = next(iter(self._user_non_avg_scorers))
+                scorer = self._user_non_avg_scorers[scorer_name]
+            else:
+                scorer = None
+        return get_single_scorer_or_default(self.task_type, self.model, self.train, scorer, multiclass_non_avg)
+
+
+class SingleDatasetBaseCheck(BaseCheck):
```
noamzbr, line 143: Why are the subclasses of check inside a module called context?

deepchecks/base/check_context.py
```diff
@@ -0,0 +1,181 @@
...
+        if train and model:
+            validate_model(train, model)
+
+        self._train = train
+        self._test = test
+        self._model = model
+        self._feature_importance_mode = feature_importance_mode
+        self._features_importance = None
+        self._task_type = None
+        self._user_scorers = scorers
```
noamzbr, line 52:
```suggestion
        self._user_avg_scorers = scorers
```

deepchecks/base/check_context.py
```diff
@@ -0,0 +1,181 @@
...
+
+    @property
+    def features(self):
+        if not self.train.features_columns:
+            raise DatasetValidationError('Check is irrelevant for Datasets without features')
+        return self.train.features
+
+    @property
+    def date_name(self):
+        if self.train.datetime_name is None:
```
noamzbr, line 89: not correct (same for index) - datetime can be in dataframe index. Current way of checking if datetime is not set is by checking if self.datetime_col is None.

also, call this datatime

deepchecks/base/check_context.py
```diff
@@ -0,0 +1,181 @@
...
+
+
+class CheckRunContext:
+
+    def __init__(self,
+                 train=None,
+                 test=None,
+                 model=None,
+                 feature_importance_mode: str = 'model',
+                 scorers: Mapping[str, Union[str, Callable]] = None,
```
noamzbr, line 19:
```suggestion
                 avg_scorers: Mapping[str, Union[str, Callable]] = None,
```

noamzbr (commented):
main comment - how does the user manually set feature importance?

deepchecks/base/check_context.py
matanper: This is just temporary easier for me to update

deepchecks/base/check_context.py
matanper: why? if avg is the default behavior

noamzbr: does the user know that when setting these parameters? I thought that in this case stating stuff explicitly is better.

matanper: I think he knows. this is also the default behavior of skelarn

deepchecks/base/dataset.py
```diff
@@ -825,22 +798,22 @@ def datasets_share_categorical_features(cls, datasets: t.List['Dataset']) -> boo
                 'datasets' parameter is not a list;
                 'datasets' contains less than one dataset;
         """
-        assert isinstance(datasets, list), "'datasets' must be a list"
+        assert isinstance(datasets, tuple), "'datasets' must be a list"
```
yromanyshyn, line 801: should be removed, it always will be a tuple

ItayGabbay (approved):
Good luck for us

deepchecks/base/check_context.py
```diff
@@ -0,0 +1,279 @@
...
+        self._model = model
+        self._feature_importance_force_permutation = feature_importance_force_permutation
+        self._features_importance = features_importance
+        self._feature_importance_timeout = feature_importance_timeout
+        self._calculated_importance = False
+        self._importance_type = None
+        self._validated_model = False
+        self._task_type = None
+        self._user_scorers = scorers
+        self._user_non_avg_scorers = non_avg_scorers
```
shir22, line 103: do we have a better name for this? e.g. per_class_scorers

matanper: Possible will change in another PR

PR #734, opened 2022-01-24
Rename scorers related parameters

matanper:


deepchecks/base/check_context.py
```diff
@@ -43,8 +43,8 @@ class CheckRunContext:
         timeout in second for the permutation features importance calculation
     scorers : Mapping[str, Union[str, Callable]] , default: None
         dict of scorers names to scorer sklearn_name/function
-    non_avg_scorers : Mapping[str, Union[str, Callable]] , default: None
-        dict of scorers for multiclass without averaging of the classes
+    scorers_per_class : Mapping[str, Union[str, Callable]] , default: None
+        dict of scorers for classification without averaging of the classes
```
shir22, line 47: WDYT about adding a link to sklearn - to exemplify these types of scorers...?
(or any other way that gives an example of "what is this scorer", because I don't think it's intuitive for the user...
and maybe it shouldn't be documented here cause this isn't really user facing, but was thinking "where to start")

matanper: The same parameters here will be exposed to the user (in suite or check run) so we'll just copy the docstring anyway

deepchecks/base/check_context.py
```diff
@@ -244,36 +249,36 @@ def assert_regression_task(self):
...
             raise ModelValidationError('Check is irrelevant for classification tasks')
 
-    def get_scorers(self, alternative_scorers: Mapping[str, Union[str, Callable]] = None, multiclass_avg=True):
+    def get_scorers(self, alternative_scorers: Mapping[str, Union[str, Callable]] = None, class_avg=True):
         """Return initialized & validated scorers in a given priority.
 
         If receive `alternative_scorers` return them,
         Else if user defined global scorers return them,
         Else return default scorers.
         """
```
ItayGabbay, line 258:
```suggestion
        """Return initialized & validated scorers in a given priority.
        
        If receive `alternative_scorers` return them,
        Else if user defined global scorers return them,
        Else return default scorers.
       
       Parameters
       -----------
       alternative_scorers : Mapping[str, Union[str, Callable]] , default: None
           Alternative scorers that will be calculated, in the sklearn scorers' format
       class_avg : bool , default: True
           Whether the results will be averaged (applies to classification tasks only)

        """
```

ItayGabbay (approved):

PR #737, opened 2022-01-25
formatting/displaying logic

yromanyshyn:
our output formatting/displaying logic become a little bit messy, to make things more ordered suggest adding `Presentation[T]` abstraction (for more details please take a look at `from deepchecks.base.presentation` package)

```python
T = t.TypeVar("T")


class Presentation(t.Generic[T]):

    def __init__(self, value: T, **kwargs):
        self.value = value
    
    def as_html(self, *, **kwargs) -> str:
        raise NotImplementedError()
    
    def as_json(self, *, **kwargs) -> str:
        raise NotImplementedError()
    
    def as_widget(self, *, **kwargs) -> Widget:
        raise NotImplementedError()
```

### DIFF
--- a/deepchecks/base/presentation/abc.py
+++ b/deepchecks/base/presentation/abc.py
@@ -1,0 +1,23 @@
+import typing as t
+from ipywidgets import Widget
+
+
+__all__ = ["Presentation"]
+
+
+T = t.TypeVar("T")
+
+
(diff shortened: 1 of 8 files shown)


matanper:
I don't understand where the notebook html vs widgets happens here, but it looks cleaner than our current code

PR #739, opened 2022-01-25
explicitly set default timeout for fi calculation

noamzbr:
Small fix - feature importance timeout calculation should be available in context defaults.


matanper:
Updated few things

matanper (approved):

PR #746, opened 2022-01-25
Label drift

nirhutnik:
This is the initial version of the label drift test.

This is not yet complete:
1. Functionality here needs to be improved (right now going over all batches for each metric separately)
2. Doctstrings are not updated
3. Tests are non-existent
4. Notebook doesn't show drift - we need to create a "test" version of coco for that.

However, this initial version is working so merging so it can be used as reference.


deepchecks/vision/checks/distribution/train_test_label_drift.py
```diff
@@ -0,0 +1,447 @@
...
+        test_dataset = VisionDataset.validate_dataset(test_dataset)
+        train_dataset.validate_label()
+        test_dataset.validate_label()
+
+        task_type = train_dataset.label_type
+        displays = []
+
+        if task_type == TaskType.CLASSIFICATION.value:
+
+            train_label_distribution = train_dataset.get_samples_per_class()
```
noamzbr, line 101: not really related to your PR, but think this should be stored as it may be costly (currently it's done per call)

deepchecks/vision/checks/distribution/train_test_label_drift.py
```diff
@@ -0,0 +1,447 @@
...
+    delta = 1 / expected_percents.size
+    emd = 0
+    for i in range(dirt.shape[0]-1):
+        dirt_to_pass = dirt[i] - expected_percents[i]
+        dirt[i+1] += dirt_to_pass
+        emd += abs(dirt_to_pass)*delta
+    return emd
+
+
+def calc_drift_and_plot(train_distribution: dict, test_distribution: dict, plot_title: Hashable,
```
noamzbr, line 301: A lot of code from here on should be in base right? So we can use it in both subpackages... at least the plotting part, and maybe the computation past the generation of hists

deepchecks/vision/checks/distribution/train_test_label_drift.py
```diff
@@ -0,0 +1,447 @@
...
+        # In order for the value not to diverge, we cap our min percentage value
+        e_perc = max(expected_percents[i], PSI_MIN_PERCENTAGE)
+        a_perc = max(actual_percents[i], PSI_MIN_PERCENTAGE)
+        value = (e_perc - a_perc) * np.log(e_perc / a_perc)
+        psi_value += value
+
+    return psi_value
+
+
+def earth_movers_distance_by_histogram(expected_percents: np.ndarray, actual_percents: np.ndarray):
```
noamzbr, line 290: As a sanity check, can you test this against the scipy implementation?

deepchecks/vision/checks/distribution/train_test_label_drift.py
```diff
@@ -0,0 +1,447 @@
...
+    for i in range(len(dataset._data)):
+        list_of_arrays = next(iter(dataset._data))[1]
+        calc_res = [transform(arr) for arr in list_of_arrays]
+        if len(calc_res) != 0 and isinstance(calc_res[0], list):
+            calc_res = [x[0] for x in sum(calc_res, [])]
+        counter.update(calc_res)
+    return counter
+
+
+def histogram_in_batch(dataset: VisionDataset, transform: Callable = lambda x: x):
```
noamzbr, line 239: transform is kinda reserved word here in torchvision. Maybe label_property or label_computation or label_derivative?

deepchecks/vision/checks/distribution/train_test_label_drift.py
```diff
@@ -0,0 +1,447 @@
...
+    label_min = np.inf
+    label_max = -np.inf
+    for i in range(len(dataset._data)):
```
noamzbr, line 242: isn't it more common to do 
`for batch in dataset._data:
        list_of_label_arrays = batch[1]`
?

deepchecks/vision/checks/distribution/train_test_label_drift.py
```diff
@@ -0,0 +1,447 @@
...
+        list_of_arrays = next(iter(dataset._data))[1]
+        calc_res = [transform(arr) for arr in list_of_arrays]
```
noamzbr, line 244: feel like we should have a built in defense against failing transforms. Such as a try catch that makes sure we get null if fails so we can remove it and go on rather than crash.

deepchecks/vision/checks/distribution/train_test_label_drift.py
```diff
@@ -0,0 +1,447 @@
...
+        if len(calc_res) != 0 and isinstance(calc_res[0], list):
+            calc_res = [x[0] for x in sum(calc_res, [])]
+        label_min = min(calc_res + [label_min])
+        label_max = max(calc_res + [label_max])
+
+    hist, edges = np.histogram([], bins=100, range=(label_min, label_max))
```
noamzbr, line 250: bins should be hyperparameter

deepchecks/vision/checks/distribution/train_test_label_drift.py
```diff
@@ -0,0 +1,447 @@
...
+def count_num_bboxes(label):
+    num_bboxes = label.shape[0]
+    return num_bboxes
+
+
+def count_custom_transform_on_label(dataset: VisionDataset, transform: Callable = lambda x: x):
+    counter = Counter()
+    for i in range(len(dataset._data)):
+        list_of_arrays = next(iter(dataset._data))[1]
+        calc_res = [transform(arr) for arr in list_of_arrays]
```
noamzbr, line 232: same comments as in histogram_in_batch

noamzbr (commented):
Nice work. Didn't comment on docs as they were not updated yet. 

Point to consider - this check may be a good place to start sampling rather than going over all the data. This can be done by either by creating a copy of dataloader with shuffle=True, or making sure existing one has this property. Then limiting to a set amount of samples. Really there is no need for drift to go over all the data in the dataloader. This will apply for:
1. get_samples_per_class
2. histogram_in_batch
3. count_custom_transform_on_label

deepchecks/vision/checks/distribution/train_test_label_drift.py
nirhutnik: yes, will move later

deepchecks/vision/checks/distribution/train_test_label_drift.py
nirhutnik: Already did and it is a very close approximation. Did not do this blindly :)

deepchecks/vision/checks/distribution/train_test_label_drift.py
nirhutnik: I don't know, can change

noamzbr:
Closed in light of #797

PR #748, opened 2022-01-26
Add CheckRunConfig

matanper:
After @noamzbr suggested to add the parameters to all `run` in the base checks, I decided to try and put everything in a class so we won't copy-paste it in a lot of places. 

pros: no copy-paste
cons: a bit harder to initialize `check.run(ds, model, force_permutation=True)` vs `check.run(ds, model, CheckRunConfig(force_permutation=True)`

### DIFF
--- a/deepchecks/base/__init__.py
+++ b/deepchecks/base/__init__.py
@@ -23,7 +23,7 @@ from .check import (
     ModelComparisonBaseCheck,
     ModelComparisonContext
 )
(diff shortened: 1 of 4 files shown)


ItayGabbay (changes requested):
I don't like this solution. we need to discuss about it

matanper:
Closing due to disagreement

PR #749, opened 2022-01-26
Add more warning for the user when calculating FI

matanper:
I added some warning in the feature importance calculation so the user will know not to be surprised.

1. force_permutation=True + dataframe => warn that can't use permutation on dataframe so using built-in of model
2. permutation timeout => warn that instead of permutation using the built-in of the model (plus before we just returned null on timeout instead of falling to the model's FI)
3. force_permutatation=False + didn't find built-in FI in model => warn that I'm using permutation FI instead of model's


noamzbr (approved):
nice

PR #751, opened 2022-01-26
[issue-650] functionality split

### LINKED ISSUE #650: separate the package to /tabular and /core namespaced packages

yromanyshyn:
resolves #650


deepchecks/tabular/__init__.py
```diff
@@ -0,0 +1,37 @@
...
+# You should have received a copy of the GNU Affero General Public License
+# along with Deepchecks.  If not, see <http://www.gnu.org/licenses/>.
+# ----------------------------------------------------------------------------
+#
+"""Package for tabular functionality."""
+from .dataset import Dataset
+from .base import (
+    TabularContext,
+    TabularCheck,
+    Suite,
```
matanper, line 16: I think if leave it "Suite" and not "TabularSuite" we can do the same for the rest of classes

yromanyshyn: like this?
- `TabularContext` -> `Context`
- `TabularCheck` -> `Check`

do not know/ @ItayGabbay , @noamzbr  what do you think?

ItayGabbay: I think it's better

yromanyshyn: renamed

ItayGabbay (approved):

PR #771, opened 2022-01-27
added tests for single feature data + improved tests

JKL98ISR:


deepchecks/utils/metrics.py
```diff
@@ -118,29 +118,24 @@ def __init__(self, scorer: t.Union[str, t.Callable], name: str):
             raise errors.DeepchecksValueError(message)
 
     @classmethod
-    def filter_nulls(cls, dataset: 'tabular.Dataset'):
+    def filter_nulls(cls, dataset: 'tabular.Dataset') -> 'tabular.Dataset':
         """Return data of dataset without null labels."""
         valid_idx = dataset.data[dataset.label_name].notna()
-        return dataset.data[valid_idx]
+        return dataset.copy(dataset.data[valid_idx])
```
noamzbr, line 124: isn't this the same in pandas?

deepchecks/utils/metrics.py
```diff
@@ -118,29 +118,24 @@ def __init__(self, scorer: t.Union[str, t.Callable], name: str):
...
 
-    def _run_score(self, model, dataframe, dataset):
```
noamzbr, line 126: I seem to remember there was a reason for that... places where the dataframe passes was not the dataset.feature_columns. Could you please check?

deepchecks/tabular/dataset.py
```diff
@@ -711,6 +711,8 @@ def features_columns(self) -> pd.DataFrame:
         pd.DataFrame
         """
         self.assert_features()
+        if len(self.features) == 1:
```
noamzbr, line 714: I feel like this could cause problems - I would expect this to always return a DataFrame. Need to test this out on a full_suite on several models IMO
Also, if we decide this is OK - change the typing of the function return.

ItayGabbay (approved):

PR #780, opened 2022-01-30
updated makefile

yromanyshyn:


ItayGabbay (approved):

yromanyshyn:
@ItayGabbay 
IMO it will be better to use env markers ([pep link](https://www.python.org/dev/peps/pep-0508/#environment-markers)) for requirements instead of
```
@if [ $(OS) = "Linux" ]; \
	then \
		$(PIP) install torch==1.10.2+cpu torchvision==0.11.3+cpu torchaudio==0.10.2+cpu -f https://download.pytorch.org/whl/cpu/torch_stable.html ; \
	else \
		$(PIP) install -U torch torchvision torchaudio; \
	fi;
```
I will update

PR #786, opened 2022-01-31
Add sample data loader & sample predictions and label caching

matanper:
For iterable dataset I had to load all samples into memory in O(2n) complexity


deepchecks/vision/base.py
```diff
@@ -147,90 +130,35 @@ def task_type(self) -> TaskType:
...
 
     def assert_task_type(self, *expected_types: TaskType):
-        """Assert task_type matching given types.
-
-        If task_type is defined, validate it and raise error if needed, else returns True.
-        If task_type is not defined, return False.
-        """
-        # To calculate task type we need model and train. if not exists return False, means we did not validate
-        if self._model is None or self._train is None:
-            return False
```
ItayGabbay, line 162: Why did you remove this?

deepchecks/vision/dataset.py
```diff
@@ -69,8 +69,13 @@ class VisionDataset:
...
 
-    def __init__(self, data_loader: DataLoader, label_type: str, num_classes: Optional[int] = None,
-                 label_transformer: Optional[Callable] = None):
+    def __init__(self,
+                 data_loader: DataLoader,
+                 label_type: str,
+                 num_classes: Optional[int] = None,
+                 label_transformer: Optional[Callable] = None,
+                 sample_size: int = 1000,
+                 seed: int = 0):
```
ItayGabbay, line 78: random_seed is better imo

deepchecks/vision/dataset.py
```diff
@@ -211,3 +235,83 @@ def validate_dataset(cls, obj) -> 'VisionDataset':
...
+    dataset = data_loader.dataset
+    # IterableDataset doesn't work with samplers, so instead we manually copy all samples to memory and create
+    # new dataset that will contain them.
+    if isinstance(dataset, torch.utils.data.IterableDataset):
+        iter_length = 0
+        for _ in dataset:
+            iter_length += 1
+        sample_size = min(sample_size, iter_length)
+        np.random.seed(seed)
+        sample_indices = set(np.random.choice(iter_length, size=(sample_size,), replace=False))
```
ItayGabbay, line 304: Isn't it an anti-pattern to iterate an iterator to find its length?
I understand we can't sample in a uniform distribution otherwise, but maybe it's better to limit the number of samples we are iterating on (and it can be a large number, e.g. 100000 or something). then, we can sample using [Reservoir sampling](https://en.wikipedia.org/wiki/Reservoir_sampling).

I know it might be an overkill solution now, but I'm afraid from use-cases where the data will be enormous..

deepchecks/vision/base.py
matanper: a. it goes nowhere the return, we use this method as "raise error else it's okay"
b. it is already encapsulated when calling `task_type`. I'm not sure about the method there, if we can have task_type without model than the validation there also need to change

deepchecks/vision/dataset.py
matanper: @noamzbr WDYT?

matanper: Okay done that. Added a note in VisionDataset which is pretty important. might be good to put it somewhere else also.

ItayGabbay (approved):

ItayGabbay:
@matanper @noamzbr I really think we should put this note somewhere else. 
+ Reservoir sampling is the perfect solution to our problem and will make sure we are sampling uniformly. But - and more important - it's academic and people will like that we did that. Maybe we should put that on our backlog.

PR #788, opened 2022-01-31
Vision docs placeholder

### LINKED ISSUE #752: [DOCS][CV]Documentation placeholders

ItayGabbay:
#### Reference Issues/PRs

Closes #752 

#### Any other comments?

---


shir22 (approved):
technical note (request for a small addition, if sounds good open an issue?) -
I think we want both for the vision and for the tabular an "index.rst" with a toc also inside the main "checks" folder (and not only inside each of the "types of checks". cause it's nice to have a page with a list, and maybe in the future we'll make it visual (have pics for each).

ItayGabbay:
@shir22 i think these pages are already exist but are not included in any toctree

shir22:
nope, probably deleted at some point (yes they used to exist... )

yromanyshyn:
@shir22 

not sure if I understand you correctly/ but probably I removed those files. Because they were causing problems with the navigation sidebar.

I was not digging deep into that problem but in general, it is better when all `toctree`  form a correct tree, a situation like this could cause some problems

test drawio

PR #789, opened 2022-01-31
[issue-781] moved checks base classes to core

### LINKED ISSUE #781: [FEAT][CV] Move check base classes to core and create base classes for dataset & context
Currently, BaseCheck is in core but it's subclasses (such as TrainTestBaseCheck) are duplicated in the tabular and vision subpackages.

The solution is creating one version of these base checks in core, and creating base classes for vision and tabular dataset and context to make that possible.

yromanyshyn:
resolves #781


deepchecks/core/check.py
```diff
@@ -458,6 +459,79 @@ def name(cls):
...
+    """Parent class for checks that only use a model and no datasets."""
+
+    context_type: ClassVar[Optional[Type[Any]]] = None  # TODO: Base context type
+
+    def __init__(self):
+        super().__init__()
+        # Replace the run_logic function with wrapped run function
+        setattr(self, 'run_logic', wrap_run(getattr(self, 'run_logic'), self))
+
+    def run(self, model) -> CheckResult:
```
JKL98ISR, line 524: why we switched the run and run_logic?

yromanyshyn: 1. take a look at suite impl, it calls run_logic  ` check_result = check.run_logic(context) `
2. I did not change it was like this

yromanyshyn:
> is there a way to move the classes in a way git sees it as renaming? because the last pr that move the tabular removed all the file history

@JKL98ISR 
do not know/ if somebody approves it I will explore this question before merging

noamzbr (approved):
Looks fine good to me. Only question is do we want to avoid the confusion of having both tabular.SingleDatasetCheck and vision.SingleDatasetCheck. For our purposes, it's convenient and elegant but we / a future user may confuse the two...

matanper (approved):
To summarize I'm not sure about the global base checks, but for now we can go with that until our demands will require change

PR #800, opened 2022-02-01
Adding labels and detection encoder classes

### LINKED ISSUE #759: [FEAT][CV]Label formats + transform
Currently, labels are required to match a precise format, and we enable transforming the labels into that format using the label_transformer callable parameter of our dataset. 

In addition to that, we should have an enum linking common formats to our format. Our format is called xywh (when all quantities are in pixel coordinates, and x and y are of the upper left corner of the image), so the enum should enable the user to easily select transformation callables for the following formats:

1. xyxy - upper left and lower right
2. xywhn - same as ours, but quantities are normalized by the image dimensions, so that if x is 100pxl and image width is 400pxl than in this coordinates x will be 0.25
3. xyxyn
4. yolo - yolo has it's own format, which is similar to xywhn except x and y are measured from the image center.

ItayGabbay:
#### Reference Issues/PRs

Closes #759 

Adding encoding classes for classification and detection use-cases

#### Any other comments?

---


deepchecks/vision/utils/detection_encoders.py
```diff
@@ -0,0 +1,64 @@
...
+            - 'cxcywh' - x, y, w, h represent the center of the bounding box and its width and height.
+            - 'xyxyn' - x1, y1, x2, y2, n represent the upper left and lower right corners of the bounding box,
+                        normalized to the image dimensions.
+            - 'xywhn' - x1, y1, w, h, represent the upper left corner of the bounding box and its width and height,
+                        normalized by the image dimensions.
+            - 'cxcywhn' - (YOLO format) x, y, w, h, represent the center of the bounding box and its width and height,
+                          normalized by the image dimensions.
+        If a function, it must follow the signature:
+        Function that takes in a batch of labels and returns the encoded labels in the following format:
+        List of length N containing tensors of shape (B, 5), where N is the number of samples,
```
noamzbr, line 26: I say B used as "number of samples in batch" so consider a different letter.

deepchecks/vision/utils/detection_encoders.py
```diff
@@ -0,0 +1,64 @@
...
+
+class DetectionLabelEncoder:
+    """
+    Class for encoding the detection annotations to the required format.
+
+    Parameters
+    ----------
+    label_encoder : Union[str, Callable]
+        Function or string that specifies the encoding function.
+        If a string, it must be one of the following:
```
noamzbr, line 14: State that in all these, class comes before coordinates.

noamzbr, line 14: In the string option - also worth noting that you are supposed to pass a string only if it represents *exactly* the format outputted as a batch from your dataloader. If the match is not exact - use a callable.

noamzbr: Also need to note that the batch label is a list containing tensors with axis=0 being the axis of bounding boxes and axis=1 being the specified format.

noamzbr (commented):
Currently the validation of the label is defined and performed in the dataset obj, and the definition of the validation of predictions is in metrics.py in validate_prediction. 

I think that these functions should be defined in the Encoder classes, and only used in these locations. In that way we avoid these ifs between the task types

deepchecks/vision/dataset.py
```diff
@@ -81,31 +80,33 @@ class VisionDataset:
...
-        if label_type in valid_label_types:
-            self.task_type = TaskType(label_type)
+        if isinstance(self.label_transformer, ClassificationLabelEncoder):
+            self.task_type = TaskType.CLASSIFICATION
+        elif isinstance(self.label_transformer, DetectionLabelEncoder):
+            self.task_type = TaskType.OBJECT_DETECTION
         else:
-            raise DeepchecksValueError(f'Invalid label type: {label_type}, must be one of {valid_label_types}.')
+            raise ValueError(f'Unsupported label transformer type. {type(self.label_transformer)}, must be one of '
+                             f'ClassificationLabelEncoder or DetectionLabelEncoder.')
```
nirhutnik, line 105: Repeating my comment from Slack - why? We want to enable users to use task types that are not classification or object detection, we just won't have any default OOTB configurations for them. Why not enable them to use a general LabelEncoder?

deepchecks/vision/dataset.py
```diff
@@ -45,11 +46,9 @@ class VisionDataset:
     ----------
     data_loader : DataLoader
         PyTorch DataLoader object. If your data loader is using IterableDataset please see note below.
-    label_type : str
-        Type of label. Must be one of the following: 'classification', 'object_detection'.
     num_classes : int, optional
         Number of classes in the dataset. If not provided, will be inferred from the dataset.
-    label_transformer : Callable, optional
+    label_transformer : Union[ClassificationLabelEncoder, DetectionLabelEncoder]
```
nirhutnik, line 51: Now that I see it, won't it be confusing with scikit-learn's LabelEncoder? @noamzbr WDYT?

deepchecks/vision/utils/classification_encoders.py
```diff
@@ -0,0 +1,139 @@
...
+            DataLoader to get the samples per class from.
+
+        Returns
+        -------
+        Counter
+            Dictionary of the number of samples per class.
+
+        """
+        counter = Counter()
+        for _ in range(len(data_loader)):
```
nirhutnik, line 63:
```suggestion
        for batch in data_loader:
```

deepchecks/vision/utils/classification_encoders.py
```diff
@@ -0,0 +1,139 @@
...
+            counter.update(self(next(iter(data_loader))[1].tolist()))
```
nirhutnik, line 64:
```suggestion
            counter.update(batch[1].tolist())
```

deepchecks/vision/utils/detection_encoders.py
```diff
@@ -0,0 +1,172 @@
...
+            - 'xyxy' - x1, y1, x2, y2 represent the upper left and lower right corners of the bounding box.
+            - 'xywh' - x1, y1, w, h represent the upper left corner of the bounding box and its width and height.
+            - 'cxcywh' - x, y, w, h represent the center of the bounding box and its width and height.
```
nirhutnik, line 36:
```suggestion
            - 'cxcywh' - cx, cy, w, h represent the center of the bounding box and its width and height.
```

deepchecks/vision/utils/detection_encoders.py
```diff
@@ -0,0 +1,172 @@
...
+            - 'xyxyn' - x1, y1, x2, y2, n represent the upper left and lower right corners of the bounding box,
+                        normalized to the image dimensions.
+            - 'xywhn' - x1, y1, w, h, represent the upper left corner of the bounding box and its width and height,
+                        normalized by the image dimensions.
+            - 'cxcywhn' - (YOLO format) x, y, w, h, represent the center of the bounding box and its width and height,
+                          normalized by the image dimensions.
+        In addition, the label shape should be a list of length N containing tensors of shape (M, 5), where N is the
```
nirhutnik, line 43: Why is it "in addition"? this contradicts the explanation of the formats before

deepchecks/vision/utils/detection_encoders.py
```diff
@@ -0,0 +1,172 @@
...
+        number of samples, M is the number of bounding boxes, and each bounding box is represented by 5 values:
+        (class_id, 4 coordinates in the format specified by the `label_encoder` parameter).
+
+        If a function, it must follow the signature:
+        Function that takes in a batch of labels and returns the encoded labels in the following format:
+        List of length N containing tensors of shape (M, 5), where N is the number of samples,
+        M is the number of bounding boxes in the sample and each bounding box is represented by 5 values: (class_id,
+        x, y, w, h). x and y are the coordinates (in pixels) of the upper left corner of the bounding box, w and h are
+        the width and height of the bounding box (in pixels) and class_id is the class id.
```
nirhutnik, line 52: This explanation is quite long. Maybe refer to the format description above, and instead of the lengthy explanation, add an example of a transformed label?

deepchecks/vision/utils/detection_encoders.py
```diff
@@ -0,0 +1,172 @@
...
+        data_loader : DataLoader
+            DataLoader to get the samples per class from.
+
+        Returns
+        -------
+        Counter
+            Counter of the number of samples per class.
+        """
+        counter = Counter()
+        for _ in range(len(data_loader)):
```
nirhutnik, line 84:
```suggestion
        for batch in data_loader:
```

deepchecks/vision/utils/detection_encoders.py
```diff
@@ -0,0 +1,172 @@
...
+            list_of_arrays = self(next(iter(data_loader))[1])
```
nirhutnik, line 85:
```suggestion
            list_of_arrays = batch[1]
```

deepchecks/vision/utils/detection_encoders.py
```diff
@@ -0,0 +1,172 @@
...
+            class_list = sum([arr.reshape((-1, 5))[:, 0].tolist() for arr in list_of_arrays], [])
+            counter.update(class_list)
+
+        return counter
```
nirhutnik, line 89: There are relevant functions in TrainTestLabelDrift that should be moved here, IMO - calculating bbox areas and counting bboxes in images

deepchecks/vision/dataset.py
ItayGabbay: I agree it's confusing. any idea for a better name?

deepchecks/vision/utils/detection_encoders.py
ItayGabbay: I am not sure. Perhaps we should calculate it once using the COCO JSON format. @JKL98ISR is still figuring out how to make this work.

deepchecks/vision/dataset.py
ItayGabbay: I can't see someone who will write a label encoder, prediction encoder, and metrics just to get a performance report.  In my opinion, the code is clearer that way. 
Most of the other checks won't work even with custom configuration as the assumptions of classification/regression are bound into the core logic of the checks.

deepchecks/vision/dataset.py
noamzbr: label formatter
label translator
label molder 
label schema

only first two are real candidates but you never know...

deepchecks/vision/dataset.py
noamzbr: Tend to agree with @ItayGabbay on that

deepchecks/vision/dataset.py
```diff
@@ -81,31 +80,33 @@ class VisionDataset:
...
-                 label_transformer: Optional[Callable] = None,
+                 label_transformer: Union[ClassificationLabelEncoder, DetectionLabelEncoder] = None,
                  sample_size: int = 1000,
                  sample_iteration_limit: int = 1_000_000,
                  random_seed: int = 0):
         self._data = data_loader
 
         if label_transformer is None:
-            self.label_transformer = lambda x: x
+            self.label_transformer = ClassificationLabelEncoder(lambda x: x)
```
noamzbr, line 95: Think it's clearer if label_transformer =None will mean no task and only integrity / data checks can run

deepchecks/vision/datasets/classification/mnist.py
```diff
@@ -93,7 +93,6 @@ def load_dataset(
     elif object_type == 'Dataset':
         return VisionDataset(
             data_loader=loader,
```
noamzbr, line 95: state explicitly (because I'm against the assumption of classification)

deepchecks/vision/utils/detection_encoders.py
noamzbr: I think that we must state this in details, but an example is a good idea.

noamzbr (approved):

deepchecks/vision/dataset.py
nirhutnik: Disagree. Many checks are agnostic to the task (all integrity)

PR #801, opened 2022-02-01
Robustness Report

matanper:
Missing tests and need to further elaborate in the notebook


deepchecks/vision/checks/performance/robustness_report.py
```diff
@@ -0,0 +1,244 @@
...
+                 alternative_metrics: Optional[List[Metric]] = None,
+                 prediction_extract: Optional[Callable] = None,
+                 augmentations=None,
+                 epsilon: float = 10 ** -2,
+                 random_seed: int = 42):
+        super().__init__()
+        self._epsilon = epsilon
+        self.alternative_metrics = alternative_metrics
+        self.prediction_extract = prediction_extract
+        self.random_seed = random_seed
```
noamzbr, line 46: in most tabular checks we used "random_state". Think we want to transition towards using this as our standard.

deepchecks/vision/checks/performance/robustness_report.py
```diff
@@ -0,0 +1,244 @@
...
+        self.augmentations = augmentations
+
+    def run_logic(self, context: Context, dataset_type: str = 'train') -> CheckResult:
```
noamzbr, line 49: This should run on test by default.

deepchecks/vision/metrics_utils/metrics.py
```diff
@@ -129,7 +130,10 @@ def validate_prediction(batch_predictions: t.Any, dataset: VisionDataset, eps: f
         )
 
 
-def calculate_metrics(metrics: t.List[Metric], dataset: VisionDataset, model: nn.Module,
+def calculate_metrics(metrics: t.List[Metric],
+                      dataset: VisionDataset,
+                      data_loader: DataLoader,
```
noamzbr, line 135: why have both dataset and it's dataloader? I get it that it's more convenient for the robustness check where the dataloader is modified, but don't think it's reason enough.

deepchecks/vision/dataset.py
```diff
@@ -89,7 +90,8 @@ def __init__(self,
                  label_transformer: Optional[Callable] = None,
                  sample_size: int = 1000,
                  sample_iteration_limit: int = 1_000_000,
-                 random_seed: int = 0):
+                 random_seed: int = 0,
```
noamzbr, line 93: same comment on random seed.

deepchecks/vision/dataset.py
```diff
@@ -214,6 +217,25 @@ def get_data_loader(self):
...
+        try:
+            transform_field = dataset_ref.__getattribute__(self.transform_field)
+            # Validate in the wrapper that the transform_field is of valid type
+            transform_type = get_transform_type(transform_field)
+            # Validate the field is in use in the dataset
+            dataset_copy = copy(dataset_ref)
+            # Replace transform field with stupid transformation
+            dataset_copy.__setattr__(self.transform_field, get_dummy_transform(transform_type))
+            if dataset_copy[0] == dataset_ref[0]:
+                raise DeepchecksValueError(f'Transform field {self.transform_field} found to not be in use')
```
noamzbr, line 233: Elaborate a bit more - tell the user how you checked and what you found

deepchecks/vision/checks/performance/robustness_report.py
```diff
@@ -0,0 +1,244 @@
...
+                for j in range(3):
+                    print(j + (i * 3))
+                    fig.add_trace(
+                        go.Image(z=aug_images[j + (i * 3)]), row=i + 1, col=j + 1
+                    )
+            fig.update_layout(height=600, width=800)
+            figures.append(fig)
+        return figures
+
+    def _set_seeds(self, seed):
```
noamzbr, line 207: This is nice, maybe should be a member of image check bases...

deepchecks/vision/checks/performance/robustness_report.py
```diff
@@ -0,0 +1,244 @@
...
+        return [
+            albumentations.RandomBrightnessContrast(p=1.0),
+            albumentations.ShiftScaleRotate(p=1.0),
+            albumentations.HueSaturationValue(p=1.0),
+            albumentations.RGBShift(r_shift_limit=15, g_shift_limit=15, b_shift_limit=15, p=1.0),
+        ]
+    # imgaug augmentations works also inside pytorch compose
+    else:
+        return [
+            imgaug.augmenters.MultiplyHueAndSaturation()
```
noamzbr, line 234: Need to expand to match the albumentations version if possible, but can do in a later PR.

deepchecks/vision/checks/performance/robustness_report.py
```diff
@@ -0,0 +1,244 @@
...
+                .for_each_yaxis(lambda yaxis: yaxis.update(showticklabels=True))
+        )
+
+        return CheckResult(
+            results_df,
+            header='Robustness Report',
+            display=fig
+        )
+
+    def _report_inner_loop(self, dataset: VisionDataset, model, scorers):
```
noamzbr, line 104: Think we want to call metrics metrics here, as that is their ignite name. @ItayGabbay WDYT?

deepchecks/vision/checks/performance/robustness_report.py
```diff
@@ -0,0 +1,244 @@
...
+            examples.append(self._get_random_image_pairs_from_dataloder(dataset.get_data_loader(), copy_data_loader))
+
+        # Create grand DataFrame from dictionary of augmentations
+        # This has flat hierarchy for augmentations and baseline-vs-augmentations
+        results_df = pd.concat(results, axis=0). \
+            reset_index(). \
+            rename({"level_0": "Status"}, axis=1). \
+            drop(labels="level_1", axis=1)
+
+        def _is_robust_to_augmentation(p, epsilon=10 ** -2):
```
noamzbr, line 150: should be a condition.

deepchecks/vision/checks/performance/robustness_report.py
```diff
@@ -0,0 +1,244 @@
...
+        """
+        We iterate the internal dataset object directly to avoid randomness
+        Dataset returns data points as processed images, making this currently not really usable
+        To avoid making more assumptions this currently stays as-is
+        Note that images return in RGB format, ond to visualize them using OpenCV the final dimesion should be
+        transposed;
+        can be done via image = image[:, :, ::-1] or cv2.cvtColor(img, cv2.COLOR_RGB2BGR)
+        """
+        # We definitely make the assumption that the underlying structure is torch.utils.Dataset
+        baseline_sampler = iter(baseline_data_loader.dataset)
```
noamzbr, line 179: in most datasets (or at least datasets which have `__getitem__`, and I suspet that even for Iterables) this would most certainly lead to sampling along the image order, which means it's likely to get images of the same class. We must have a method to select here, whether it is a way of sampling randomly, selecting samples of different classes or selecting the images for which the "loss" is greatest (from different classes).

deepchecks/vision/checks/performance/robustness_report.py
matanper: This doesn't work like that. this is way SingleDatasetCheck behave

deepchecks/vision/metrics_utils/metrics.py
matanper: A. we might have 2 data loaders in VIsionDataset (normal one and sample one)
B. Feel free to share another solution, I couldn't find

deepchecks/vision/checks/performance/robustness_report.py
```diff
@@ -0,0 +1,244 @@
...
+            curr_results_aug = evaluate(copy_data_loader)
+            results_aug_df = pd.DataFrame(curr_results_aug, columns=['Class', 'Metric', 'Value', 'N']
+                                          ).sort_values(by=['Class'])
+            # This modifies the augmentation name in-place
+            results_base_df["Augmentation"] = aug_name
+            # This adds the augmentation
+            results_aug_df["Augmentation"] = aug_name
+            results.append(pd.concat([results_base_df, results_aug_df], keys=["Baseline", "Augmented"]))
+
+            examples.append(self._get_random_image_pairs_from_dataloder(dataset.get_data_loader(), copy_data_loader))
```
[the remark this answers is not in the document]
matanper: They are

deepchecks/vision/checks/performance/robustness_report.py
matanper: It depends if the data is shuffled or not. Anyway I'm not sure how to implement it without going over the data a lot of times

deepchecks/vision/checks/performance/robustness_report.py
matanper: We have a problem with the condition, how do we decide what augs to show to the user without this?

deepchecks/vision/checks/performance/robustness_report.py
```diff
@@ -0,0 +1,244 @@
...
+    ----------
+        alternative_metrics (List[Metric], default None):
+            A list of ignite.Metric objects whose score should be used. If None are given, use the default metrics.
+    """
+
+    def __init__(self,
+                 alternative_metrics: Optional[List[Metric]] = None,
+                 prediction_extract: Optional[Callable] = None,
+                 augmentations=None,
+                 epsilon: float = 10 ** -2,
```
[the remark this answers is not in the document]
nirbenz: This is a single epsilon for all metrics, optimally Epislon can be a list or a diciontary mapping metric to epsilon.

deepchecks/vision/checks/performance/robustness_report.py
nirbenz: I'd think seeding would happen in a single place in the whole repo and outside of a specific class. Ala;
from deepchecks import seed_all
seed_all(42)

matanper: If we want to make the checks consistent we need to manually reset the seed at the start of each check, no?

deepchecks/vision/checks/performance/robustness_report.py
noamzbr: so what's the role of dataset_type: str = 'train' ?

matanper: I mean it is the role, but it's fixed for SingleDatasetCheck, it's not per-check configuration

deepchecks/vision/checks/performance/robustness_report.py
nirbenz: Yep. But I'd call a method of the entire deepcheck suite. A lot of python libraries have that. That's basically like calling np.seed. I'd do that for all of DeepCheck, which the use can use to set a seed for all underlying used frameworks.

deepchecks/vision/checks/performance/robustness_report.py
```diff
@@ -271,31 +299,6 @@ def _create_top_affected_graph(self, top_affected_dict):
         return fig
 
 
-def get_robustness_augmentations(dataset):
```
nirbenz: I'm actually not a fan of having big if-albumentation elif imgaug across the code as it can go wrong fast. I'd like to this this abstracted away in the inialization function of a wrapping class. Initialize the augmentation classes there, self-assign class methods and from that point onwards simply calling them will be agnostic to the library used by the user.

deepchecks/vision/checks/performance/robustness_report.py
```diff
@@ -271,31 +299,6 @@ def _create_top_affected_graph(self, top_affected_dict):
...
-    transform_type = dataset.get_transform_type()
-    multi_channel = dataset.data_dimension > 1
-    if transform_type == 'albumentations':
-        # Note that p=1.0 since we want to apply those to entire dataset
-        augmentations = [
-            albumentations.RandomBrightnessContrast(p=1.0),
-            albumentations.ShiftScaleRotate(p=1.0),
-        ]
-        if multi_channel:
```
nirbenz: What happens here?

deepchecks/vision/utils/transformations.py
```diff
@@ -8,36 +8,74 @@
...
-    else:
-        raise DeepchecksNotSupportedError(f'Not implemented for type {transform_type}')
+
+    @classmethod
+    def get_test_transformation(cls):
+        return imgaug.augmenters.Rotate(rotate=(20, 30))
+
+    @classmethod
+    def get_robustness_augmentations(cls, data_dim):
+        augmentations = [
```
nirbenz, line 48: So, are the actual augmentations hard coded here? Can't the user input them in the constructor?
By the way- I'd assert that the given list of robustness- augmentations is the same type of the inner-attribute given by the user in `__getattr__`

deepchecks/vision/checks/performance/robustness_report.py
matanper: There are augmentations that doesn't work on black-white images

deepchecks/vision/utils/transformations.py
matanper: This are default, user can pass his own. (It's not here, it's in the check)
And we do asserting, we return the appropriate type by the type the user have in his field

deepchecks/vision/dataset.py
```diff
@@ -188,7 +179,36 @@ def get_data_loader(self):
         """Return the data loader."""
         return self._data
 
-    def validate_shared_label(self, other):
+    def get_transform_type(self):
+        """Return transforms handler created from the transform field."""
+        dataset_ref = self.get_data_loader().dataset
+        # If no field exists raise error
+        if not hasattr(dataset_ref, self.transform_field):
+            raise DeepchecksValueError(f'Underlying Dataset instance must have a {self.transform_field} attribute')
```
noamzbr, line 187: Maybe not it should be the one containing the aug?

deepchecks/vision/datasets/classification/mnist.py
```diff
@@ -198,4 +206,4 @@ def forward(self, x):
         x = F.relu(self.fc1(x))
         x = F.dropout(x, training=self.training)
         x = self.fc2(x)
-        return F.log_softmax(x)
+        return F.log_softmax(x, dim=1)
```
noamzbr, line 209: How did the performance report work pre-this?

deepchecks/vision/utils/image_functions.py
```diff
@@ -0,0 +1,113 @@
...
+import plotly.graph_objects as go
+import torch
+
+from deepchecks.core.errors import DeepchecksValueError
+
+
+__all__ = ['get_image_info', 'numpy_to_image_figure', 'apply_heatmap_image_properties', 'label_bbox_add_to_figure']
+
+
+class ImageInfo:
```
noamzbr, line 27: @nirhutnik please have a look here, you may have relevant feedback for #839

deepchecks/vision/utils/transformations.py
```diff
@@ -0,0 +1,106 @@
...
+def un_normalize_batch(tensor, mean: Sized, std: Sized, max_pixel_value: int = 255):
+    """Apply un-normalization on a tensor in order to display an image."""
+    dim = len(mean)
+    reshape_shape = (1, 1, 1, dim)
+    max_pixel_value = [max_pixel_value] * dim
+    mean = torch.tensor(mean).reshape(reshape_shape)
+    std = torch.tensor(std).reshape(reshape_shape)
+    tensor = (tensor * std) + mean
+    tensor = tensor * torch.tensor(max_pixel_value).reshape(reshape_shape)
+    return tensor.numpy()
```
noamzbr, line 106:
```suggestion
    return tensor.cpu().numpy()
```

deepchecks/vision/checks/performance/robustness_report.py
```diff
@@ -0,0 +1,406 @@
...
+    def __init__(self,
+                 prediction_formatter: BasePredictionFormatter,
+                 alternative_metrics: Optional[List[Metric]] = None,
+                 augmentations=None,
+                 random_state: int = 42):
+        super().__init__()
+        self.alternative_metrics = alternative_metrics
+        self.prediction_formatter = prediction_formatter
+        self.random_state = random_state
+        self.augmentations = augmentations
```
noamzbr, line 51: probably should be validated here?

deepchecks/vision/checks/performance/robustness_report.py
```diff
@@ -0,0 +1,406 @@
...
+        # Save as result only the metrics diff per augmentation
+        result = {aug: data['metrics_diff'] for aug, data in aug_all_data.items()}
+
+        return CheckResult(
+            result,
+            header='Robustness Report',
+            display=figures
+        )
+
+    def add_condition_degradation_not_more_than(self, ratio: 0.01):
```
noamzbr, line 128:
```suggestion
    def add_condition_degradation_not_greater_than(self, ratio: 0.01):
```

deepchecks/vision/checks/performance/robustness_report.py
```diff
@@ -0,0 +1,406 @@
...
+            for index_class, diff_value in diff.sort_values().iloc[:n_classes_to_show].iteritems():
+                aug_top_affected[metric].append({'class': index_class,
+                                                 'value': single_metric_scores.at[index_class, 'Value'],
+                                                 'diff': diff_value,
+                                                 'samples': dataset.get_samples_per_class()[index_class]})
+        return aug_top_affected
+
+    def _calc_performance_diff(self, mean_base, augmented_metrics):
+        def difference(aug_score, base_score):
+            # TODO how to handle zero base score
```
noamzbr, line 204: return max value? Not 0. Quick hack would be return 1. Better would be to look at how it's done in tabular.

deepchecks/vision/checks/performance/robustness_report.py
```diff
@@ -0,0 +1,406 @@
...
+    transposed;
+    can be done via image = image[:, :, ::-1] or cv2.cvtColor(img, cv2.COLOR_RGB2BGR)
+    """
+    classes_to_show = {
+        class_info['class']: class_info['diff']
+        for classes_list in top_affected_classes.values()
+        for class_info in classes_list
+    }
+
+    baseline_sampler = iter(original_dataset.get_data_loader().dataset)
```
noamzbr, line 364: Maybe we should check that iter on two copies of the original returns the same image, before we rely on this here to produce the pairs.

deepchecks/vision/dataset.py
```diff
@@ -146,12 +132,16 @@ def to_display_data(self, batch):
         self.image_transformer.validate_data(batch)
         return self.image_transformer(batch)
 
+    @property
+    def data_dimension(self):
```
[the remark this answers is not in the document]
matanper: I don't know, can do that

deepchecks/vision/utils/classification_formatters.py
```diff
@@ -36,7 +36,7 @@ class ClassificationLabelFormatter(BaseLabelFormatter):
         of predictions returned by the model.
     """
 
-    def __init__(self, label_formatter: Callable):
+    def __init__(self, label_formatter: Callable = lambda x: x):
```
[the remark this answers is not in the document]
matanper: what's the problematic part? it's immutable to it's valid, but not understandable?

deepchecks/vision/checks/performance/robustness_report.py
matanper: what should be? the type of augmentations? anyway there is validation needed that is based on the dataset that passed, so felt it is too much to have separate validations

deepchecks/vision/checks/performance/robustness_report.py
matanper: note it's not the metric value, it's the difference. so 0 ratio means no change from base to aug. the other option is base is 0 and aug is not, and then it means aug was better, which is less interesting probably just random luck

deepchecks/vision/dataset.py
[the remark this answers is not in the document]
noamzbr: I was meaning - it's easier because you know exactly the structure of the data post-image_transformer

deepchecks/vision/dataset.py
noamzbr: updated original comment, meant maybe we should say here that this field should contain the aug

deepchecks/vision/checks/performance/robustness_report.py
matanper: I think it doesn't matter. if it doesn't you will still want to show examples, so the only difference might be some more space between original and augmented examples or something which is not that critical

requirements/vision-requirements.txt
```diff
@@ -1,3 +1,5 @@
 pytorch-ignite>=0.4.8
 opencv-python>=4.5.5.62
 albumentations>=1.1.0
+kaleido>=0.2.1
+imgaug
```
noamzbr, line 5: Version?

deepchecks/vision/checks/performance/robustness_report.py
```diff
@@ -0,0 +1,428 @@
...
+                return 0 if aug_score == 0 else np.inf if aug_score > base_score else -np.inf
+            return (aug_score - base_score) / base_score
+
+        diff_dict = {}
+        for metric, score in self._calc_mean_metrics(augmented_metrics).items():
+            diff_dict[metric] = {'score': score, 'diff': difference(score, mean_base[metric])}
+
+        return diff_dict
+
+    def _calc_mean_metrics(self, metrics_df) -> dict:
```
noamzbr, line 237:
```suggestion
    def _calc_median_metrics(self, metrics_df) -> dict:
```

deepchecks/vision/checks/performance/robustness_report.py
```diff
@@ -0,0 +1,428 @@
...
+        model = context.model
+
+        # Validate the transformations works
+        transforms_handler = dataset.get_transform_type()
+        self._validate_augmenting_affects(transforms_handler, dataset)
+        # Return dataframe of (Class, Metric, Value)
+        base_results: pd.DataFrame = metric_results_to_df(
+            {k: m.compute() for k, m in self._state['metrics'].items()}, dataset
+        )
+        # TODO: update later the way we handle average metrics
```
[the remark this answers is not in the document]
matanper: you told me to add that, I don't know to elaborate

deepchecks/vision/utils/transformations.py
```diff
@@ -0,0 +1,114 @@
...
+    @classmethod
+    def get_test_transformation(cls):
+        """Get transformation which is affecting both image data and bbox."""
+        return imgaug.augmenters.Rotate(rotate=(20, 30))
+
+    @classmethod
+    def get_robustness_augmentations(cls, data_dim):
+        """Get default augmentations to use in robustness report check."""
+        augmentations = [
+            imgaug.augmenters.MultiplyHueAndSaturation()
```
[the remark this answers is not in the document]
matanper: it's ben zvi task

noamzbr (approved):

PR #805, opened 2022-02-02
[issue-799] coco split into train/test; updated coco dataset module api

### LINKED ISSUE #799: [FEAT][CV] Create train-test COCO object detection dataset
Current coco built-in dataloader just uses all data.
For many checks and examples, it'll be better to split it to train and test dataloaders.

yromanyshyn:
resolves #799


deepchecks/vision/datasets/detection/coco.py
```diff
@@ -9,28 +9,94 @@
...
+    num_workers: int = 0,
+    shuffle: bool = False,
+    pin_memory: bool = True,
+    object_type: Literal['Dataset', 'DataLoader'] = 'DataLoader'
+) -> t.Union[DataLoader, vision.VisionDataset]:
+    """Get the COCO dataset and return a dataloader.
+
+    Parameters
+    ----------
+    batch_size : int, default: 64
```
nirhutnik, line 54: Make 32 the default, so if we have train/test defaults, they will both have 2 batches

yromanyshyn:
update:

@ItayGabbay @nirhutnik 
please take a look at the `deepchecks/vision/metrics_utils/detection_precision_recall.py`
I had to make a small fix to it, not sure if it is correct
The reason why I added it is because it was failing at this point

In general `_compute_ap_recall` was returning a dict with all values set to None, but 
tensor cannot be created with none values, you will get an error if you try

```
RuntimeError: Could not infer dtype of NoneType
```

deepchecks/vision/datasets/detection/coco.py
yromanyshyn: updated

nirhutnik:
Seems good to me, but would prefer Gabbay take a look as its his code. If I understand correctly, this means that sometimes we get average precision = None (why?), and in those cases you want it to count as 0. Not sure what is done later with that information - if its summed up, then ok, but I guess it is averaged again and in that case maybe should be just ignored and not counted in the denominator as well.

But didn't dive into the code yet.

yromanyshyn:
also, are those two lines within `VisionContext` correct?
(that is a reason why notebook CI/CD check fails)

```python
80: if train and test:
81:      train.validate_shared_label(test)
```

if I got it right, in object-detection task label is a 2d array each row of which 
represent a separate object (bbox), and each image(label) could contain different 
number of objects

we probably should only verify that number of columns is the same?

nirhutnik:
@yromanyshyn I think that Gabbay's PR solves that anyway as he transfers label validation to the LabelEncoder classes

deepchecks/vision/datasets/detection/coco.py
```diff
@@ -9,28 +9,103 @@
...
 
+def load_dataset(
+    train: bool = True,
+    batch_size: int = 32,
+    num_workers: int = 0,
+    shuffle: bool = False,
+    pin_memory: bool = True,
+    object_type: Literal['Dataset', 'DataLoader'] = 'DataLoader'
+) -> t.Union[DataLoader, vision.VisionDataset]:
+    """Get the COCO dataset and return a dataloader.
```
ItayGabbay, line 50: I would call it the COCO 128 dataset, as the COCO dataset is much larger than that.

ItayGabbay: Also, did we validate the license of this dataset? and model?

makefile
```diff
@@ -229,9 +229,12 @@ notebook: requirements dev-requirements
 	@$(JUPYTER) nbextension enable --py widgetsnbextension
 	@echo "+++ Number of notebooks to execute: $$(find ./docs/source/examples -name "*.ipynb" | wc -l) +++"
 	@echo "+++ Executing notebooks in $(PWD) +++"
-	@find ./docs/source/examples -name "*.ipynb" | \
-	 xargs -P4 -I'{}' $(JUPYTER) nbconvert --execute '{}' \
-	  --to notebook --stdout > /dev/null
+
+	$(JUPYTER) nbconvert --execute $$(find ./docs/source/examples -name "*.ipynb") --to notebook --stdout > /dev/null
```
ItayGabbay, line 233: why?

deepchecks/vision/metrics_utils/detection_precision_recall.py
```diff
@@ -79,7 +79,10 @@ def compute(self):
                 **self._compute_ap_recall(ev["scores"], ev["matched"], ev["NP"])
             }
         if self.return_ap_only:
-            res = torch.tensor([res[k]["AP"] for k in sorted(res.keys())])
+            res = torch.tensor([
```
ItayGabbay, line 82: I think that's OK. This may happen in the case the model detected a class that doesn't exist in the test set.
Anyway, we are planning to replace this module in the near future.

ItayGabbay (changes requested):

makefile
yromanyshyn: weird but I initially thought that this was causing 'notebook check' failure. I have undone it

deepchecks/vision/datasets/detection/coco.py
yromanyshyn: changed, 
it is GNU Version 3

makefile
ItayGabbay: Maybe you are correct. I'm now checking this

ItayGabbay (approved):

PR #807, opened 2022-02-03
Fix widget plotly bug

### LINKED ISSUE #762: [BUG] In some cases to plotly widget creation fails
**Describe the bug**
In some cases to plotly widget creation fails

**Screenshots**
Screen Shot 2022-01-26 at 16 14 24

 - OS: mac
 - Python Version: 3.8
 - Deepchecks Version: 0.3.2

### LINKED ISSUE #804: [BUG] Calling `save_as_html` can raise an error if the label column is imbalanced and has the type `np.bool_`
**Describe the bug**
The error comes from this chart (section "Train Test Label Drift"):
image

This bar chart is created whenever the label is imbalanced and drifts between train and test (which is the case with my data).

The x axis has the values `False` and `True` with the dtype `np.bool_`. Before saving as HTML, somewhere in the code, this diagram is encoded as JSON. The call to `json.dumps` causes the error. The solution is to convert the problematic data to the regular Python `bool` type.

**To Reproduce**
1. Run the code below
2. See error: "TypeError: Object of type bool_ is not JSON serializable"

```python
import pandas as pd
import numpy as np

from deepchecks import Dataset
from deepchecks.suites import train_test_validation


if __name__ == "__main__":
    # Generate some random data
    # Note that the data type of the generated label is numpy.bool_
    # (in my project, the label column has this type)
    np.random.seed(42)
    training_size = 1000
    training_positive_proportion = 0.01  # Very low proportion to simulate an imbalanced dataset
    testing_size = 100
    testing_positive_proportion = 0.05  # This proportion is different from training to simulate a drift on the label
    training_df = pd.DataFrame({
        "feature_1": np.ones(training_size),
        "label": np.random.choice(a=[False, True], size=(training_size,), p=[training_positive_proportion, 1 - training_positive_proportion])
    })
    testing_df = pd.DataFrame({
        "feature_1": np.ones(testing_size),
        "label": np.random.choice(a=[False, True], size=(testing_size,), p=[testing_positive_proportion, 1 - testing_positive_proportion])
    })

    # Run the train_test_validation suite
    training_ds = Dataset(df=training_df, label="label", features=["feature_1"])
    testing_ds = Dataset(df=testing_df, label="label", features=["feature_1"])
    suite = train_test_validation()
    results = suite.run(train_dataset=training_ds, test_dataset=testing_ds)

    # Save as HTML, this will raise the following error:
    # "TypeError: Object of type bool_ is not JSON serializable"
    results.save_as_html("output.html")
```

**Expected behavior**
I would expect deepchecks to deal with this internally by converting the legend values to `bool`.

 - OS: Ubuntu 20.04 (running on Windows 10 with WSL)
 - Python Version: 3.8.10
 - Deepchecks Version: 0.4.1

JKL98ISR:
fix #762 
fix #804


deepchecks/utils/distribution/plot.py
```diff
@@ -108,6 +108,16 @@ def drift_score_bar_traces(drift_score: float, bar_max: float = None) -> Tuple[L
     return bars, xaxis, yaxis
 
 
+def _un_numpy(val):
```
ItayGabbay, line 111: Can you please add a comment with a link to the issue and some elaborations?

JKL98ISR: did a comment

nirhutnik (approved):

ItayGabbay (approved):

PR #813, opened 2022-02-03
📓 Updated main page and when should you use page

shir22:

Main page: added images and removed link to use case.
When should you use page: added more elaborate description and direct links to suites.
Readme: fixed typo in conda installation


shir22:
Merged to 0.4.x in separate PR (PR to main will be different)

PR #826, opened 2022-02-07
change checks to init - update - compute structure 

noamzbr:
This PR is intended to change the structure of our vision checks to use the following methods:
initialize_run - everything that should be done before we start running on the batches returned by dataloader
update - accumulate results for a specific batch
compute - return the final CheckResult using accumulated values.

Updated also the base checks in vision to match, where now run_logic handels running all these functions for a call of a single check. 

This is a draft, please let me know what you think. If it's approved I'll expand these changes to suite and to the label drift check.


deepchecks/vision/base.py
```diff
@@ -158,6 +158,33 @@ class SingleDatasetCheck(SingleDatasetBaseCheck):
 
     context_type = Context
 
+    def run_logic(self, context, dataset_type: str = 'train') -> CheckResult:
+        """Run check."""
+
+        self.initialize_run(context)
```
matanper, line 164: isn't the initialize run should be called once from the suite?

matanper, line 164: isn't the initialize run should be called once from the suite?
what the purpose of having it here?

matanper, line 164: isn't the initialize run should be called once from the suite?
what the purpose of having it here?

noamzbr:
> so all checks are built for batches in their design like the ignite metrics?

Basically yes, I just named our "reset" to initialize_run

deepchecks/vision/checks/performance/performance_report.py
```diff
@@ -45,44 +45,56 @@ def __init__(self,
...
+        self._state = {}
         context.assert_task_type(TaskType.CLASSIFICATION, TaskType.OBJECT_DETECTION)
-
         # Get default scorers if no alternative, or validate alternatives
-        scorers = get_scorers_list(test_dataset, self.alternative_metrics)
-        datasets = {'Train': train_dataset, 'Test': test_dataset}
-
-        classes = train_dataset.get_samples_per_class().keys()
-        plot_x_axis = 'Class'
+        self._state['train'] = {}
```
matanper, line 55: why do you need train and test state? the check shouldn't care about it.
for each check it should do init, then run then reset

deepchecks/vision/checks/performance/performance_report.py
```diff
@@ -45,44 +45,56 @@ def __init__(self,
...
+        self._state['test'] = {}
+        self._state['train']['scorers'] = get_scorers_list(context.train, self.alternative_metrics)
```
matanper, line 57: why do you need to save all this properties (scorers, samples, classes) in the state?
They will be available in the context and they are cached anyway.

deepchecks/vision/checks/performance/performance_report.py
```diff
@@ -45,44 +45,56 @@ def __init__(self,
...
+        """Compute final check result based on accumulated internal state."""
         results = []
-
-        for dataset_name, dataset in datasets.items():
-            n_samples = dataset.get_samples_per_class()
+        for dataset_name in ['train', 'test']:
+            n_samples = self._state[dataset_name]['n_samples']
             results.extend(
                 [dataset_name, class_name, name, class_score, n_samples[class_name]]
-                for name, score in calculate_metrics(list(scorers.values()), dataset, model,
```
JKL98ISR, line 74: do we need the calculate_metrics then? maybe we should delete it?

JKL98ISR (changes requested):

deepchecks/vision/checks/performance/performance_report.py
noamzbr: Done

deepchecks/vision/base.py
noamzbr: done

JKL98ISR (approved):

deepchecks/vision/base.py
```diff
@@ -152,12 +135,60 @@ def assert_task_type(self, *expected_types: TaskType):
                 f'Check is irrelevant for task of type {self.train.task_type}')
         return True
 
+    def infer(self, batch: Any) -> Any:
```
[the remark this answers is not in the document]
noamzbr: to cache the predictions for a given batch. Used for e.g. in performance report.

deepchecks/vision/base.py
```diff
@@ -215,60 +295,104 @@ def run(
...
-            except Exception as exp:
-                results.append(CheckFailure(check, exp))
+
+        # Create instances of SingleDatasetCheck for train and test if train and test exist.
+        # This is needed because in the vision package checks update their internal state with update, so it will be
+        # easier to iterate and keep the check order if we have an instance for each dataset.
+        checks_list = []
+        for check_idx, check in list(self.checks.items()):
+            if isinstance(check, (TrainTestCheck, ModelOnlyCheck)):
+                checks_list.append((check_idx, copy.deepcopy(check)))
```
matanper, line 305: why do you need to copy here?

noamzbr: Because checks have state, seems cleaner to me to copy them when doing a specific .run() call rather than use the originals used in the constructor

deepchecks/vision/base.py
```diff
@@ -215,60 +295,104 @@ def run(
...
+
+        # Initialize all checks
+        for check in check_dict.values():
+            check.initialize_run(context)
+
+        # Create result dict to hold results by check name
+        results = OrderedDict({check_idx: None for check_idx, check in check_dict.items()})
+
+        # Loop over training batches
+        n_batches = len(context.train.get_data_loader())
```
matanper, line 323: so we decided data loader must be mapped?

PR #855, opened 2022-02-13
Added properties in image formatter

nirhutnik:
Added image properties.
Added it to ImageFormatter as I expect in the future (or maybe in this PR we will cache the data in batch, like we do in context.infer.

Still missing tests


deepchecks/vision/utils/image_formatters.py
```diff
@@ -32,14 +31,21 @@ class ImageFormatter:
         grayscale images should have 1 channel. The batch itself can be any iterable - a list of arrays, a tuple of
         arrays or simply a 4D numpy array, when the first dimension is the batch dimension.
         If None, the identity function will be used (the data will be assumed to be in format as-is).
+    sample_size_for_image_properties: int, optional. default: 1000
+        sampling size used for image property estimation.
+        If defined, samples only N pixels (uniformly) out of image and calculates property on those.
+        If None, calculates on whole image.
```
noamzbr, line 37:
```suggestion
        If None, calculates on the whole image.
```

deepchecks/vision/utils/image_formatters.py
```diff
@@ -73,3 +79,91 @@ def validate_data(self, batch_data):
             raise DeepchecksValueError('The data inside the iterable must have 1 or 3 channels.')
         if sample.min() < 0 or sample.max() > 255:
             raise DeepchecksValueError('The data inside the iterable must be in the range [0, 255].')
+
+    def sizes(self, batch: List[np.array]):
+        """Return list of tuples of image height and width."""
+        return [self.get_size(img) for img in batch]
+
+    def ratio(self, batch: List[np.array]):
```
noamzbr, line 87:
```suggestion
    def aspect_ratio(self, batch: List[np.array]):
```

deepchecks/vision/utils/image_formatters.py
```diff
@@ -73,3 +79,91 @@ def validate_data(self, batch_data):
...
+        """Return list of integers of image area (height multiplied by width)."""
+        return [np.prod(self.get_size(img)) for img in batch]
+
+    def brightness(self, batch: List[np.array]):
+        """Calculate brightness on each image in the batch."""
+        if self._is_grayscale(batch) is True:
+            return [img.mean() for img in batch]
+        else:
+            flattened_batch = self._flatten_batch(batch)
+            return [(0.299*img[:, 0] + 0.587*img[:, 1] + 0.114 * img[:, 2]).mean() for img in flattened_batch]
```
noamzbr, line 101: no existing to_greyscale?

deepchecks/vision/utils/image_formatters.py
```diff
@@ -73,3 +79,91 @@ def validate_data(self, batch_data):
...
+
+    def contrast(self,  batch: List[np.array]):
+        pass
```
noamzbr, line 104: NotImplementedError

noamzbr (commented):
Makes sense to put these funcs here, we just need to have a clear example of how the user can make his own. 

Regarding the function outputs - in order for these properties to be properties (for which we can calc drift etc), they need to return a float per image. So:
1. We need a clear distinction of what functions are "helpers" (such as normalized_rgb_mean or sizes) that produce complex outputs like lists of arrays and are used by other functions, and actual "properties" that return a float per image.
2. We need to add some properties here, such as "green_relative_intensity" etc.

deepchecks/vision/utils/image_formatters.py
nirhutnik: Surprisingly not if your not reading from a file or using another class.
Also, as this is done on the flattened version, as I want this to work on sampling, this is better in my opinion.
Thought about extracting to another method, WDYT?

nirhutnik:
> 

Regarding custom usage, they can create their own formatter anyway. But maybe this requires more design. 

Regarding the rest:
1. Agree, was wondering about that as well, WDYT about adding "property" to the function? Weird, but... IDK.
2. Agree

noamzbr (approved):

PR #857, opened 2022-02-14
suite fixes

yromanyshyn:


noamzbr (approved):
Looks good, but maybe we should make it so that trying to access context.train if not set will return None rather than raise an exception?

yromanyshyn:
@noamzbr 

> Looks good, but maybe we should make it so that trying to access context.train if not set will return None rather than raise an exception?

does not this contradict the initial purpose of the context/
@matanper

matanper:
> @noamzbr
> 
> > Looks good, but maybe we should make it so that trying to access context.train if not set will return None rather than raise an exception?
> 
> does not this contradict the initial purpose of the context/ @matanper

Right because then in every check we will have to test it instead of just in context

deepchecks/vision/base.py
```diff
@@ -362,102 +363,160 @@ def run(
...
+                results=results
+            )
+
+        for check_idx, check in checks.items():
+            if check_idx not in results:
+                try:
+                    if isinstance(check, SingleDatasetCheck):
+                        results[check_idx] = check.compute(
+                            context,
+                            dataset_kind=None # TODO: when this method should be called?
```
matanper, line 420: This should be the appropriate dataset

yromanyshyn: I understand this) but this code is outside training/validation loops, which dataset is appropriate?

yromanyshyn: should it be called twice if train and test were passed?

yromanyshyn: fixed, resolving

PR #869, opened 2022-02-15
Added imgaug-equivalent augmentations.

nirbenz:
Let's discuss them:
- albumentations.RandomBrightnessContrast
To replicate this I used a sequential composition of LinearContrast (which I think does the same thing to contrast) and MultiplyBrightness, which *doesn't* do the same thing and results are a bit different. The LUT formulization of albumentations is elegant but hard to replicated (direct implementation is easy however, should we want to go there).
- albumentations.ShiftScaleRotate -
to replciate this I used Affine with proper params.
- albumentations.HueSaturationValue -
imgaug WithColorspace and channel-specific Add arithmetic does the trick.
- albumentations.RGBShift - likewise, only working in RGB space.

#### Reference Issues/PRs

#### Any other comments?

---


nirbenz:
Forgot to mention. I also added a default argument of 3 to `data_dim`. Most images are RGB so I think this is alright, but maybe you intetionally wanted that to be explicitly added?

noamzbr:
> Forgot to mention. I also added a default argument of 3 to `data_dim`. Most images are RGB so I think this is alright, but maybe you intetionally wanted that to be explicitly added?

Currently this is only used in a setup where data_dim is passed, but in general yes I expect RGB to be the norm

noamzbr (approved):

PR #876, opened 2022-02-16
Documentation fixes (refactoring)

yromanyshyn:
problems which this PR solves:

- thumbnails at the bottom of the check API references
- sidebar rendering

additionally solves #753 

----------------------

list of important and less obvious changes:

**point 1.**

I removed the next files:
+ examples/tabular/index.rst
+ examples/vision/index.rst

it is because with their presence toctree was incorrect and as 
result navigation sidebar was rendered incorrectly.

It is not said in the sphinx documentation but all toctrees should 
form a directed rooted tree. Take a look at the diagram:

diagram2

**point 2.**

I restructured(refactored) content within 'api/' folder, the reason is that 
it is easier to work with `autosummary` extension in this way (IMO) 
especially when we mix-up explicit hidden toctree directives with 
autosummary directive without `:toctree:` option. Take a look at the example
to understand what I mean

```
// some-doc-file.rst

.. currentmodule:: deepchecks.core

.. autosummary::

    check
    condition

// during the docs build autosummary directive above will be expanded 
// into a table with two columns
//
// .. table::
//
//   ===================================  ======================================
//                link                                description
//   ===================================  ======================================
//     :obj:`deepchecks.core.check`         {{deepchecks.core.check.__doc__}}
//     :obj:`deepchecks.core.condition`     {{deepchecks.core.condition.__doc__}}
//   ===================================  ======================================
//
// note:
//  1. autosummary directive above does not contain :toctree: option as result toctree will not be added nor stub-file 
//     generation will happen
//  2. notice how links (:obj:`<routine>`) are formed

```

for more info about autosummary ext please refer to the [documentation](https://www.sphinx-doc.org/en/master/usage/extensions/autosummary.html?highlight=autosummary).

**point 3.**

I use `autosummary_generate_overwrite = False` + `autosummary_filename_map = {...}` hack (take a look at conf.py)
to be able to mixup within one `autosummary` directive two kinds of entries:
+ entries for which stub-files should be generated automaticly
+ entries for which stub-files were written manually

this means that each time you want to rebuild docs you should delete
'docs/source/api/generated' directive manually.

for more info about autosummary configuration options please refer to [autosummary documentation](https://www.sphinx-doc.org/en/master/usage/extensions/autosummary.html?highlight=autosummary).

**point 4.**

I renamed next 'API Reference' navigation entries:
+ from 'Core' to 'deepchecks.core'
+ from 'Tabular' to 'deepchecks.tabular'
+ from 'Vision' to 'deepchecks.vision'
+ from 'Utils' to 'deepchecks.utils'
+ from 'Checks' to 'checks' (changed first char to lowercase)

reasons:
- in python class names should start with uppercase char, not module names.
- it becomes more obvious how to import those routines

img


docs/source/_templates/autosummary/check.rst
```diff
@@ -3,7 +3,12 @@
 {% block footer %}
 
 {% set splitname = fullname.split('.') %}
-{% set checkname = splitname[-3:-1] %}
+
+{# checktype = distribution|integrity|methodology... #}
```
ItayGabbay, line 7: We have some more types, no?

yromanyshyn: it is just a comment

ItayGabbay (approved):

PR #902, opened 2022-02-21
Make image drift tests consistent

matanper:
1. changed the loading of images in coco from Pillow to cv2
2. Added set_seed method on VisionData. we technically can just load without shuffle, but for MNIST I'm not sure if it's sorted, and it's better to be consistent in the way we work with the 2 datasets (mnist and coco)

### DIFF
--- a/makefile
+++ b/makefile
@@ -208,7 +208,7 @@ docstring: dev-requirements
 
 
 test: requirements dev-requirements
-	$(PYTEST) $(args) $(TESTDIR)
+	$(PYTEST) $(args) $(TESTDIR)/vision/checks/distribution/image_dataset_drift_test.py
 
 
 test-win:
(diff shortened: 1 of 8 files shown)


makefile
```diff
@@ -208,7 +208,7 @@ docstring: dev-requirements
 
 
 test: requirements dev-requirements
-	$(PYTEST) $(args) $(TESTDIR)
+	$(PYTEST) $(args) $(TESTDIR)/vision/checks/distribution/image_dataset_drift_test.py
```
noamzbr, line 211: what's not supposed to be here right?

matanper: no, will delete. just to make me testing iterations faster

noamzbr (approved):
As you suggested, we may also want the seed to be set as part of the run arguments of vision checks, to ensure that train and test datasets are seeded before iteration.

PR #906, opened 2022-02-22
Wandb integration

JKL98ISR:


deepchecks/core/check.py
```diff
@@ -206,6 +212,79 @@ def _display_to_json(self) -> List[Tuple[str, str]]:
...
         return displays
 
+    def to_wandb(self, wandb_init: bool = True, **kwargs: Any):
+        """Export check result to wandb.
+
+        Parameters
+        ----------
+        wandb_init : bool , default: True
+            if to initiate a new wandb run
+        kwargs: Keyword arguments to pass to wandb.init - relevent if wandb_init is True.
```
ItayGabbay, line 222: Please add an Examples section that demonstrates an init with the important arguments of wandb

deepchecks/core/check.py
```diff
@@ -206,6 +212,79 @@ def _display_to_json(self) -> List[Tuple[str, str]]:
...
+            cond_table = wandb.Table(dataframe=cond_df.data, allow_mixed_types=True)
+            wandb.log({f'{section_suffix}conditions_table': cond_table}, commit=False)
+        if isinstance(self.value, pd.DataFrame):
+            value = self.value.to_json()
+        elif isinstance(self.value, Styler):
+            value = self.value.data.to_json()
+        elif isinstance(self.value, np.ndarray):
+            value = self.value.tolist()
+        else:
+            value = jsonpickle.dumps(self.value)
```
ItayGabbay, line 244: Isn't it just the same as the to_json method? maybe we should reuse the code from there..?

ItayGabbay (commented):
It looks good.
I suggest doing some code re-use in terms of the display manipulation. Right now we have duplicate code.

deepchecks/core/check.py
```diff
@@ -206,6 +212,79 @@ def _display_to_json(self) -> List[Tuple[str, str]]:
...
+        table_i = 0
+        plot_i = 0
+        old_backend = matplotlib.get_backend()
+        for item in self.display:
+            if isinstance(item, Styler):
+                # print(check_metadata['header'])
```
noamzbr, line 250: remove these?

deepchecks/core/check.py
```diff
@@ -206,6 +212,79 @@ def _display_to_json(self) -> List[Tuple[str, str]]:
...
+        data = [check_metadata['header'],
+                str(check_metadata['params']),
+                check_metadata['summary'],
+                value]
+        final_table = wandb.Table(columns=['header', 'params', 'summary', 'value'])
+        final_table.add_data(*data)
+        wandb.log({f'{section_suffix}results': final_table}, commit=False)
+        print('logged: ' + check_metadata['header'])
+        if wandb_init:
+            wandb.finish()
```
noamzbr, line 286: If we're gonna use it like this, we need to make it clear that the flag wandb_init actually means we're gonna create a run that will only exist for deepchecks, and will exit when this function exits. If so, we need to change the name of this flag, e.g. to "output_to_dedicated_run" or something.

tests/base/check_test.py
```diff
@@ -156,4 +156,5 @@ def test_params():
     assert_that(default_check.params(), equal_to({}))
     assert_that(parameter_check.params(), equal_to({'param2': 5}))
     assert_that(all_param_check.params(), equal_to({'param1': 8, 'param2': 9}))
+    assert_that(default_check.params(show_defaults=True), equal_to({'param1': 1, 'param2': 2}))
```
noamzbr, line 159: I think we can add test for the wandb functionality with a mock wandb object

deepchecks/core/check.py
```diff
@@ -206,6 +212,79 @@ def _display_to_json(self) -> List[Tuple[str, str]]:
         matplotlib.use(old_backend)
         return displays
 
+    def to_wandb(self, wandb_init: bool = True, **kwargs: Any):
```
noamzbr, line 215: It's a bit confusing that if wandb_init = True then there are args we *must* give, such as project. 
Not sure how to solve because it's indeed not mandatory if wandb_init = False...

noamzbr (commented):
In genera, would like to see more comments in the code, especially of the CheckResult.to_wandb() method. For me, especially the part with mateplotlib backend was not self explanatory, and also this function should be broken into sections (log conditions, log value, log display...)

deepchecks/core/check.py
JKL98ISR: not all the lines

deepchecks/core/check.py
```diff
@@ -206,6 +212,82 @@ def _display_to_json(self) -> List[Tuple[str, str]]:
...
+        kwargs: Keyword arguments to pass to wandb.init - relevent if wandb_init is True.
+                Default project name is check name.
+                Default config is the check metadata (params, train/test/ name etc.).
+        """
+        assert wandb, 'Missing wandb dependency, please install wandb'
+        check_metadata = self._get_metadata()
+        if dedicated_run is None:
+            dedicated_run = wandb.run is None
+        if dedicated_run:
+            kwargs['project'] = kwargs.get('project', check_metadata['name'])
```
ItayGabbay, line 232: default project name should be deepchecks

deepchecks/core/check.py
```diff
@@ -206,6 +212,82 @@ def _display_to_json(self) -> List[Tuple[str, str]]:
...
         return displays
 
+    def to_wandb(self, dedicated_run: bool = True, **kwargs: Any):
+        """Export check result to wandb.
+
+        Parameters
+        ----------
+        dedicated_run : bool , default: None
+            If to initiate and finish a new wandb run.
+            If None it will be deticated if wandb.run is None.
```
ItayGabbay, line 222: dedicated

deepchecks/utils/wandb_utils.py
```diff
@@ -0,0 +1,50 @@
...
+    import wandb
+
+    assert hasattr(wandb, '__version__')  # verify package import not local dir
+except (ImportError, AssertionError):
+    wandb = None
+
+__all__ = ['wandb_init_helper']
+
+
+def wandb_init_helper(dedicated_run: bool, default_config: dict, **kwargs: Any):
```
benisraeldan, line 24: default values for default_config and dedicated_run, looks like they aren't mandatory even if you always set them.

benisraeldan: and maybe rename to something else? it's confusing what is an `init_helper`? 
suggestion: `set_wandb_run_state`

benisraeldan:
Also how does wandb fail? is it something we want to catch and add an error of our own?

deepchecks/utils/wandb_utils.py
JKL98ISR: wdym the default values?

JKL98ISR:
> Also how does wandb fail? is it something we want to catch and add an error of our own?

I tested every type wandb can receive in the output, only way it can fail is api errors (or if I missed an edge case)

deepchecks/core/check.py
```diff
@@ -519,13 +601,43 @@ def to_json(self, with_display: bool = True):
...
 
+    def to_wandb(self, dedicated_run: bool = True, **kwargs: Any):
+        """Export check result to wandb.
+
+        Parameters
+        ----------
+        dedicated_run : bool , default: None
+            If to initiate and finish a new wandb run.
+            If None it will be dedicated if wandb.run is None.
+        kwargs: Keyword arguments to pass to wandb.init - relevent if wandb_init is True.
```
ItayGabbay, line 617: Same comment here

deepchecks/utils/wandb_utils.py
ItayGabbay: I agree with @benisraeldan about the function name

deepchecks/utils/wandb_utils.py
```diff
@@ -0,0 +1,50 @@
...
+    """Help keep a standart for wandb initialization.
+
+    Parameters
+    ----------
+    dedicated_run : bool
+        If to initiate and finish a new wandb run.
+        If None it will be dedicated if wandb.run is None.
+    default_config : dict
+        Default config dict.
+    kwargs: Keyword arguments to pass to wandb.init - relevent if wandb_init is True.
```
ItayGabbay, line 34: Same comment here

deepchecks/core/check.py
JKL98ISR: it was resolved, I did add an example

ItayGabbay (approved):

PR #909, opened 2022-02-22
set default prediction formatters

noamzbr:

In case the user does not define prediction formatters, we should provide default formatters.


deepchecks/vision/base.py
```diff
@@ -91,10 +92,33 @@ def __init__(self,
...
+                    task_type = None
+
+                if task_type == TaskType.CLASSIFICATION:
+                    prediction_formatter = ClassificationPredictionFormatter()
+                elif task_type == TaskType.OBJECT_DETECTION:
+                    prediction_formatter = DetectionPredictionFormatter()
+                else:
+                    raise DeepchecksNotSupportedError(f'No scorers match task_type {task_type}')
+
+            if prediction_formatter is not None:
```
matanper, line 115: why would it be None here?

matanper (approved):

deepchecks/vision/base.py
noamzbr: if both it and model were not defined

matanper: but you are insde "if model is not None"

PR #929, opened 2022-02-24
Update vision suite

noamzbr:
Add new checks to builtin vision suite

### DIFF
--- a/deepchecks/vision/suites/default_suites.py
+++ b/deepchecks/vision/suites/default_suites.py
@@ -14,11 +14,14 @@ Each function returns a new suite that is initialized with a list of checks and 
 It is possible to customize these suites by editing the checks and conditions inside it after the suites' creation.
 """
 from deepchecks.vision.checks import ClassPerformance, TrainTestLabelDrift, MeanAveragePrecisionReport, \
-                                     MeanAverageRecallReport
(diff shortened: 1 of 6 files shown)


ItayGabbay (approved):

deepchecks/vision/suites/default_suites.py
```diff
@@ -14,29 +14,38 @@
...
     )
 
 
 def model_evaluation() -> Suite:
     """Create a suite that is meant to test model performance and overfit."""
     return Suite(
         'Model Evaluation Suite',
-        ClassPerformance().add_condition_test_performance_not_less_than(0.5),
-        MeanAveragePrecisionReport().add_condition_test_average_precision_not_less_than(0.5),
-        MeanAverageRecallReport().add_condition_test_average_recall_not_less_than(0.5),
```
JKL98ISR, line 39: why not have a condition per check like in tabular?

noamzbr: Currently these conditions are "too conservative" and will raise a lot of failed conditions, and I'm not sure what magic number we define here will work for different classes. I think we'll need to redefine the condition (perhaps have the threshold be dependent on the number of classes) and I prefer to have this hotfix out for now till we do that.

noamzbr: Opened #930 for that

PR #935, opened 2022-02-28
Add segment performance check

### LINKED ISSUE #917: [FEAT][CV] Segment performance according to image / label properties.
Currently, the class performance check shows performance per class.

We want the option to segment the performance according to additional properties, such as image or label properties. 
Implementation can be either enabling to select some other property to segment the performance, or alternatively letting the user select two such properties and showing the resulting performance in a grid, similar to the segment performance check in tabular.

matanper:
resolve #917 

Open problems:
1. 'None' handling of properties
2. Bins size comparison to detect unshuffled data

TBD:
1. tests
3. complete notebook


deepchecks/vision/checks/performance/segment_performance.py
```diff
@@ -0,0 +1,240 @@
...
+
+        batch_data = zip(labels, predictions, batch_properties)
+        # If we already defined bins, add the current data to them
+        if bins is not None:
+            _divide_to_bins(bins, batch_data)
+        else:
+            # Add the current data to the samples list
+            samples_for_bin.extend(batch_data)
+            # Check if enough data to infer bins
+            if len(samples_for_bin) < self.number_of_samples_to_infer_bins:
```
noamzbr:
```suggestion
            if len(samples_for_bin) >= self.number_of_samples_to_infer_bins:
```
We want more sample then that threshold right?

deepchecks/vision/checks/performance/segment_performance.py
```diff
@@ -0,0 +1,240 @@
...
+        (fig.update_xaxes(title=None, type='category', matches=None)
+            .update_yaxes(title=None)
+            .for_each_annotation(lambda a: a.update(text=a.text.split('=')[-1]))
+            .for_each_yaxis(lambda yaxis: yaxis.update(showticklabels=True)))
+
+        return CheckResult(value=bins, display=fig)
+
+    def _create_bins_and_metrics(self, batch_data: t.List[t.Tuple], dataset):
+        """Return dict of bins for each property in format
+        {property_name: [{start: val, stop: val, count: x, metrics: {name: metric...}}, ...], ...}"""
```
noamzbr, line 182:
```suggestion
        {property_name: [{start: val, stop: val, count: x, metrics: {name: metric...}}], ...}"""
```

matanper: this is not true, since this is a list with multiple items

deepchecks/vision/checks/performance/segment_performance.py
matanper: fixed

deepchecks/vision/checks/performance/segment_performance.py
noamzbr: what other item is there after the metric?

ItayGabbay (approved):

PR #936, opened 2022-02-28
Adding explanations to label drift notebook.

### LINKED ISSUE #916: [FEAT][CV] Write a first "usage example" notebook
In the tabular package we often included in the check notebook an example of a usecase in which the check brings some distinct value. For example, in the trust score notebook we create concept drift that is detected in way of decrease in average confidence.

We want to have similar examples for the vision package - for example, create a case in which the property drift check finds a distinct drift along some property between train and test (preferable, one that affects the performance). This task is about creating the first of these examples and setting the standard.

ItayGabbay:
#### Reference Issues/PRs

Closes #916 

Adding examples to the train test label drift nb.

#### Any other comments?

---


ItayGabbay:
<p>Drop the whole object detection part? </p><p>Or maybe let's just drop the performance part of COCO? So we'll show what the label drift looks like for obj det</p>

---

 <a href='https://app.reviewnb.com/deepchecks/deepchecks/pull/936/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

PR #956, opened 2022-03-02
Add latest version check to package init

### LINKED ISSUE #920: [FEAT][GENERAL] check for the latest version and notify if using an older version
Check that the latest version is installed and raise a warning when importing the package if it's not the latest version.

matanper:
resolve #920

### DIFF
--- a/deepchecks/__init__.py
+++ b/deepchecks/__init__.py
@@ -11,10 +11,11 @@
 """Deepchecks."""
 import os
 import pathlib
+import http.client
 import warnings
 import matplotlib
 import plotly.io as pio
-
+import requests as requests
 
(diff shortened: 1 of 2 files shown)


deepchecks/__init__.py
```diff
@@ -98,3 +98,15 @@
 except Exception:  # pylint: disable=broad-except
     # If version file can't be found, leave version empty
     __version__ = ''
+
+# Check for latest version
+try:
+    conn = http.client.HTTPConnection('api.deepchecks.com', timeout=3)
```
ItayGabbay, line 104: Maybe add an ENV var to disable the check?

ItayGabbay (approved):

ItayGabbay (approved):

PR #966, opened 2022-03-06
Split VisionData and remove Formatters

### LINKED ISSUE #931: [FEAT] Formatter as classes
The current problem is that the formatters object are not intuitive and hard to understand. we need to consider maybe to change it to class implementation instead of passing callbacks.

The solution will be as follows:

- Remove the VisionData class and all of its usages
- Create 3 classes: `VisionTask`, `ClassificationTask`, and `DetectionTask`. The VisionTask class will contain base functionality that currently exists in the VisionData, for example  properties like `n_of_samples_per_class`, `to_display_data`, etc. In addition, it will contain the abstract function: `batch_to_images(batch)` that will replace the ImageFormatter. 
  - The `ClassificationTask` will contain classification-specific functionality, in addition to default formatter for classification. (`batch_to_labels`, `infer_on_batch`).
  - The `DetectionTask` will contain detection-specific functionality, in addition to utilities for converting bboxes format to our required format (Currently this logic exists in the DetectionFormatters). It will also contain the default formatters (`batch_to_labels`, `infer_on_batch`).

In order to run a suite/check, the user will HAVE to implement a class that inherits from the relevant task, and override the 3 "formatter" functions: `batch_to_images`, `batch_to_labels`, `infer_on_batch`.

JKL98ISR:
close #931 

Made all formatters to be "extractor" function inside the class.
A separate class for detection and classification. 
Also deleted unused code.


deepchecks/vision/utils/__init__.py
```diff
@@ -9,14 +9,3 @@
 # ----------------------------------------------------------------------------
 #
 """Package for vision utilities."""
-from .classification_formatters import ClassificationLabelFormatter, ClassificationPredictionFormatter
-from .detection_formatters import DetectionLabelFormatter, DetectionPredictionFormatter
-from .image_formatters import ImageFormatter
-
-__all__ = [
```
noamzbr, line 16: why delete all these?

deepchecks/vision/utils/detection_formatters.py
```diff
@@ -10,195 +10,14 @@
...
-            raise DeepchecksValueError('Check requires detection predictions to be a non-empty list')
-        if not isinstance(batch_predictions[0], (torch.Tensor, np.ndarray)):
-            raise DeepchecksValueError('Check requires detection predictions to be a list of torch.Tensor or'
-                                       ' numpy array')
-        if len(batch_predictions[0].shape) != 2:
-            raise DeepchecksValueError('Check requires detection predictions to be a list of 2D tensors')
-        if batch_predictions[0].shape[1] != 6:
-            raise DeepchecksValueError('Check requires detection predictions to be a list of 2D tensors, when '
-                                       'each row has 6 columns: [x, y, width, height, class_probability, class_id]')
+__all__ = ['verify_bbox_format_notation', 'convert_batch_of_bboxes', 'convert_bbox', ]
```
noamzbr, line 20: Where are we making use of this?

deepchecks/vision/utils/image_formatters.py
```diff
@@ -9,189 +9,127 @@
...
 #
 """Module containing the image formatter class for the vision module."""
-from typing import Callable, Optional, Tuple, List
-
+from typing import Tuple, List
 import numpy as np
 
-from deepchecks.core.errors import DeepchecksValueError
+__all__ = ['IMAGE_PROPERTIES',
+           'aspect_ratio',
```
noamzbr, line 16: File should be renamed to image_properties

docs/source/user-guide/vision/classification_tutorial.rst
```diff
@@ -0,0 +1,238 @@
...
+==============================================
+
+In this tutorial, you will learn how to validate your **classification model** using deepchecks test suites.
+You can read more about the different checks and suites for computer vision use cases at the
+:doc:`examples section  </examples/vision/checks/index>`
+
+A classification model is usually used to classify an image into one of a number of classes. Although there are
+multi label use-cases, in which the model is used to classify an image into multiple classes, most use-cases
+require the model to classify images into a single class.
+Currently deepchecks supports only single label classification.
```
noamzbr, line 12:
```suggestion
Currently deepchecks supports only single label classification (either binary or multi-class).
```

noamzbr (commented):
Like we have built in suites, we should have built in Data classes that you can start from.
For example, I think a large portion of classification tasks (those built on Imagenet) will be handled perfectly by the AntsBeesData. So it should be present in our package as a default option. You can use it, or you can copy and modify it. 

currently we're not using the detection label format notation anywhere, and it's a shame because it's really useful. One such place we can use it is in the "default" (builtin) detection task class.

deepchecks/vision/classification_data.py
```diff
@@ -0,0 +1,136 @@
...
+
+    def validate_label(self, batch):
+        """
+        Validate the label.
+
+        Parameters
+        ----------
+        batch
+        """
+        labels = self.batch_to_labels(batch)
```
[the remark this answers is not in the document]
JKL98ISR: It is the same behavior as before

noamzbr (commented):
On quickstart: the title "Define a VisionData Object" is a bit confusing because we then proceed to define a DetectionData. Need to explain it inherits it. Also need to explain that it already has a lot of it's methods implemented so we only need to add the 3 (and name them explicitly)

deepchecks/vision/vision_data.py
```diff
@@ -0,0 +1,344 @@
...
+    or else the callable label_formatter should be able to transform the labels to the desired format.
+    """
+
+    def __init__(self,
+                 data_loader: DataLoader,
+                 num_classes: Optional[int] = None,
+                 label_map: Optional[Dict[int, str]] = None,
+                 transform_field: Optional[str] = 'transforms'):
+
+        self._data_loader = self._get_data_loader_copy(data_loader)
```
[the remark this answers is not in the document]
JKL98ISR: There's no reason not to, makes copy clearer too, will also help in a future change we discussed today.

deepchecks/vision/utils/__init__.py
JKL98ISR: They are no formatters now

deepchecks/vision/base.py
```diff
@@ -99,32 +97,25 @@ def __init__(self,
...
-                raise DeepchecksValueError('Can\'t pass prediction formatter without model and train data')
-            prediction_formatter.validate_prediction(next(iter(train)), model, self._device)
+        if model is not None:
+            for dataset, dataset_type in zip([train, test], ['train', 'test']):
+                if dataset is not None:
+                    try:
+                        dataset.validate_prediction(next(iter(dataset.data_loader)), model, self._device)
+                        print('ahh')
+                    except DeepchecksValueError:
+                        logger.warning('validate_prediction() was not implemented in %s dataset, '
```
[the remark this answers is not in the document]
matanper: what? you are now inside init and you have the name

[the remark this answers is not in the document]
matanper: The question is what to do when there is exception in validate_prediction. Only on not_implemented we keep going? on the rest we raise exception and don't run?

deepchecks/vision/base.py
```diff
@@ -177,15 +163,16 @@ def assert_task_type(self, *expected_types: TaskType):
...
-    def infer(self, batch: Any) -> Any:
+    def infer(self, batch: Any, dataset_kind: DatasetKind) -> Any:
         """Return the predictions on the given batch, and cache them for later."""
-        if self._batch_prediction_cache is None:
-            self._batch_prediction_cache = self.prediction_formatter(batch, self.model, self.device)
-        return self._batch_prediction_cache
+        if self._batch_prediction_cache.get(dataset_kind) is None:
+            dataset = self.get_data_by_kind(dataset_kind)
+            self._batch_prediction_cache[dataset_kind] = dataset.infer_on_batch(batch, self.model, self.device)
+        return self._batch_prediction_cache[dataset_kind]
```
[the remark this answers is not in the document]
ItayGabbay: I didn't want that the check will be under the assumption that each dataset kind run in it's turn. that's why.
In addtion, this is temprary until we cache the Batch object

deepchecks/vision/dataset.py
```diff
@@ -9,25 +9,12 @@
 # ----------------------------------------------------------------------------
 #
 """The vision/dataset module containing the vision Dataset class and its functions."""
```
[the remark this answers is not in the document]
ItayGabbay: will be deleted

noamzbr (commented):
Issues that were raised during a session with @arsenyinfo:
1. Documentation of batch_to_image says it should return `List[array]`, but is actually `Iterable[array]`.
2. The example of instantiating a ClassificationData object does not include a label_map
3. The example is missing a .save_as_html() example
4. Need to make sure that infer_on_batch() is either required to returned a cpu tensor (and then it should be in the docs and example) or alternatively transfer to cpu should be handled externally by the package code.

deepchecks/vision/utils/detection_formatters.py
JKL98ISR: maybe we should use it in the example

deepchecks/vision/base.py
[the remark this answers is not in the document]
JKL98ISR: changed

deepchecks/vision/checks/performance/robustness_report.py
```diff
@@ -95,19 +94,22 @@ def compute(self, context: Context, dataset_kind) -> CheckResult:
...
         # Get augmentations
         augmentations = self.augmentations or transforms_handler.get_robustness_augmentations(dataset.data_dimension)
         aug_all_data = {}
         for augmentation_func in augmentations:
             augmentation = augmentation_name(augmentation_func)
-            aug_dataset = self._create_augmented_dataset(dataset, augmentation_func, context.random_state)
+            aug_dataset = dataset.get_augmented_dataset(augmentation_func)
             # The metrics have saved state, but they are reset inside `calculate_metrics`
             metrics = self._state['metrics']
+            # we set the seed just before the metric to keep seed state
```
matanper, line 108:
```suggestion
            # The augmentations are pseudo-random and affected by the seeds. Setting it here to have fixed state just before the augmentations are run
```

tests/vision/checks/distribution/image_dataset_drift_test.py
```diff
@@ -107,26 +96,28 @@ def test_no_drift_rgb(coco_train_dataloader, coco_test_dataloader, device):
...
+
+    train = DriftCoco(coco_train_dataloader)
+    test = COCOData(coco_test_dataloader)
 
     check = ImageDatasetDrift()
 
     # Act
     result = check.run(train, test, random_state=42, device=device)
-
+    print(result.value)
```
matanper, line 110: remove

tests/vision/checks/performance/robustness_report_test.py
```diff
@@ -37,18 +35,17 @@ def test_mnist(mnist_dataset_train, trained_mnist, device):
     ]
     check = RobustnessReport(augmentations=augmentations)
     # Act
-    result = check.run(mnist_dataset_train, trained_mnist,
-                       prediction_formatter=ClassificationPredictionFormatter(mnist_prediction_formatter),
-                       device=device)
+    result = check.run(mnist_dataset_train, trained_mnist, device=device)
     # Assert
+    print(result.value)
```
matanper, line 40: remove.

tests/vision/checks/performance/robustness_report_test.py
```diff
@@ -37,18 +35,17 @@ def test_mnist(mnist_dataset_train, trained_mnist, device):
...
     assert_that(result.value, has_entries({
         'RandomBrightnessContrast': has_entries({
             'Precision': has_entries(score=close_to(0.984, 0.001), diff=close_to(-0.001, 0.001)),
-            'Recall': has_entries(score=close_to(0.986, 0.001), diff=close_to(-0.001, 0.001))
+            'Recall': has_entries(score=close_to(0.986, 0.001), diff=close_to(-0.0003, 0.001))
```
matanper, line 44: if we moved the set_seeds before the metrics, doesn't the results should remain the same?

JKL98ISR: no, because it wasn't changed just before before the refactor

deepchecks/vision/classification_data.py
```diff
@@ -0,0 +1,136 @@
...
+from torch.utils.data import DataLoader
+
+from deepchecks.core.errors import DeepchecksValueError, ValidationError
+from deepchecks.vision.dataset import TaskType
+from deepchecks.vision.vision_data import VisionData
+
+logger = logging.getLogger('deepchecks')
+
+
+class ClassificationData(VisionData):
```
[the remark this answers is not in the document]
JKL98ISR: we don't want to inherit from ABC because we want the user to be able to not implementing all the functions

deepchecks/vision/vision_data.py
```diff
@@ -0,0 +1,353 @@
...
+        -------
+        Iterable[np.ndarray]
+            List of images in the accepted format. Each image in the iterable must be a [H, W, C] 3D numpy array.
+            See notes for more details.
+            :func: `batch_to_images` must be implemented in a subclass.
+
+        Examples
+        --------
+        >>> def batch_to_images(self, batch):
+        ...     return batch[0]
```
noamzbr, line 133: I think the example we need here is the one that applies to imagenet images - the one we have in the example

noamzbr (approved):
approved with one last comment

PR #969, opened 2022-03-07
Change exmaples in robustness report to use html template

### LINKED ISSUE #678: Robustness check
Input: dataset & model

Check:
Add corruption (eg. short noise, impulse noise, rotation) to images and compare how much the model predictions changes before & after. Also, compare the model performance.

Output:
Value: dict[corruption, scorer(data)-scorer(corrupt_data)]
Display: 
Table with summary of corruptions and decrease in metric
For each corruption, show 2 sets of images - one where the model was confused by the corruption and one where it was successful.

matanper:
#### Reference Issues/PRs

resolve #678 

1. Using HTML template instead of plotly to create the examples figure
2. Add explanation on the graphs


deepchecks/vision/checks/performance/robustness_report.py
```diff
@@ -419,3 +396,57 @@ def get_random_image_pairs_from_dataset(original_dataset: VisionData,
...
+    <div class="row">
+        <h4 class="item">Base Image</h4>
+        {base_images}
+    </div>
+    <div class="row">
+        <h4 class="item">Augmented Image</h4>
+        {aug_images}
+    </div>
+</div>
+"""
```
ItayGabbay, line 452: Maybe in a different file?

deepchecks/vision/utils/image_functions.py
```diff
@@ -55,6 +56,28 @@ def numpy_to_image_figure(data: np.ndarray):
...
+    else:
+        raise DeepchecksValueError(f'Don\'t know to plot images with {dimension} dimensions')
+
+    if labels:
+        label_bbox_add_to_figure(labels, fig)
+
+    fig.update_yaxes(showticklabels=False, visible=True, fixedrange=True, automargin=True)
+    fig.update_xaxes(showticklabels=False, visible=True, fixedrange=True, automargin=True)
+    fig.update_layout(margin={'l': 0, 'r': 0, 't': 0, 'b': 0})
+    png = base64.b64encode(fig.to_image('png')).decode('ascii')
```
ItayGabbay, line 76: 🤢

ItayGabbay: Does it mean we don't need kaleido anymore?

matanper: what's the problem?

deepchecks/vision/checks/performance/robustness_report.py
matanper: I don't know it's kinda annoying to handle files, and the check is not alone in folder, it's with more performance checks

deepchecks/vision/utils/image_functions.py
matanper: no, plotly uses kaleido for the export

ItayGabbay (approved):

PR #977, opened 2022-03-08
Fix order of results in vision suite

### LINKED ISSUE #945: [FEAT][CV] Group train and test check results together for SingleDatasetCheck
Currently the vision suite first displays all results for train and then all results for test.

DoD:
Results for same check for train and test should show one after the other

matanper:
#### Reference Issues/PRs

resolve #945

Results in vision suite were in the order the checks ran instead of the order they were defined.

### DIFF
--- a/deepchecks/vision/checks/__init__.py
+++ b/deepchecks/vision/checks/__init__.py
@@ -10,7 +10,7 @@
 #
 """Module importing all vision checks."""
 from .performance import ClassPerformance, MeanAveragePrecisionReport, MeanAverageRecallReport, \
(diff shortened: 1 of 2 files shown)


deepchecks/vision/checks/__init__.py
```diff
@@ -23,5 +23,6 @@
     'TrainTestLabelDrift',
     'ImageDatasetDrift',
     'ImagePropertyDrift',
-    'TrainTestPredictionDrift'
+    'TrainTestPredictionDrift',
```
noamzbr, line 26: Good catch, please add it also to the performance suite

noamzbr (approved):

PR #983, opened 2022-03-09
Confusion matrix bug

matanper:
Currently, we rely in the confusion matrix on detections to have consecutive classes. 
If we will have 100 classes but not consecutive (for example we have class id 110) we will try to access the matrix at index 110 and will get out of bounds exception.

Another problem is the model might return classes which we haven't seen before. For example our dataset have class ids between 0-100, and the model can return suddenly 120. For now, I just put all the "unseen" under the same bucket, since this requires a much larger refactor in order to fix.


deepchecks/vision/checks/performance/confusion_matrix.py
```diff
@@ -107,22 +119,39 @@ def compute(self, context: Context, dataset_kind: DatasetKind = None) -> CheckRe
...
+                add_unseen_category = True
             else:
-                display_categories.append(dataset.label_id_to_name(category))
+                display_categories.append(self.matrix_id_to_class_name(category, dataset))
+
+        x = display_categories
+        y = x.copy()
+
+        if add_not_found_category:
+            description += ['No overlapping categories are for labels and prediction which did not have a matching '
```
noamzbr, line 139:
```suggestion
            description += ['"No overlapping" categories are labels and prediction which did not have a matching '
```

deepchecks/vision/checks/performance/confusion_matrix.py
```diff
@@ -107,22 +119,39 @@ def compute(self, context: Context, dataset_kind: DatasetKind = None) -> CheckRe
...
+                            'label/prediction. For example a prediction which did not have a label with sufficient '
```
noamzbr, line 140:
```suggestion
                            'label/prediction. For example a predictions that did not have a sufficiently overlapping label bounding box '
```

deepchecks/vision/checks/performance/confusion_matrix.py
```diff
@@ -107,22 +119,39 @@ def compute(self, context: Context, dataset_kind: DatasetKind = None) -> CheckRe
...
+                            'IOU will be under the "No overlapping label" category']
```
noamzbr, line 141:
```suggestion
                            'will appear under the "No overlapping label" category']
```

deepchecks/vision/checks/performance/confusion_matrix.py
```diff
@@ -107,22 +119,39 @@ def compute(self, context: Context, dataset_kind: DatasetKind = None) -> CheckRe
...
+            x += ['No overlapping prediction']
+            y += ['No overlapping label']
+        if add_unseen_category:
+            description += ['Unseen classes are for classes that did not appear in the dataset, but did had predictions'
```
noamzbr, line 145:
```suggestion
            description += ['Unseen classes are classes that did not appear in the dataset labels, but did where predicted to be present.
```

noamzbr (approved):

PR #984, opened 2022-03-09
Add progress bars to vision suite run

### LINKED ISSUE #950: [FEAT][CV] Display live status when running check of suite
Both checks and suites have three stages - init, update, compute. When running, we should display what stage are we in (and when in update, what batch). We have an existing mechanism for that, just need to check that it works when run from native python (not IPython).

matanper:
#### Reference Issues/PRs

resolve #950 

Add progress bar fo each step to always indicate what's the live status


matanper:
> Really cool! I wonder if we should remove the previous step, when advancing for the next one, or show all steps that happened in the run... wdyt?

I don't have any preference, we can leave them on while there is run, and then close them all at once

ItayGabbay (approved):

PR #991, opened 2022-03-10
Add sample option to vision

### LINKED ISSUE #943: [FEAT][CV] Optional - run on sample of all batches
A simple way to reduce check / suite run time would be doing the .update() step only on a subset of batches. This should be controlled by a n_samples parameter supplied to the run() method of checks and suites.

matanper:
#### Reference Issues/PRs

resolve #943

Add option to create sampled vision data


deepchecks/vision/vision_data.py
```diff
@@ -343,21 +348,49 @@ def _get_data_loader_copy(data_loader: DataLoader):
...
-            props['sampler'] = sampler
-        props['dataset'] = copy(data_loader.dataset)
+        # Using the batch sampler to get all indices
+        # First set generator seed to make it reproducible
+        if data_loader.generator and random_state is not None:
+            data_loader.generator.set_state(torch.Generator().manual_seed(random_state).get_state())
+
+        batch_sampler: BatchSampler = data_loader.batch_sampler
+        indices = []
+        for batch in batch_sampler:
```
noamzbr, line 361: need to set the seed for that right?

matanper: look 3 rows above

deepchecks/vision/context.py
```diff
@@ -128,11 +130,14 @@ def __init__(self,
                         logger.warning('validate_prediction() was not implemented in %s dataset, '
                                        'some checks will not run', dataset_type)
 
-        # Set seeds to if possible (we set it after the validation to keep the seed state)
-        if train and random_state:
-            train.set_seed(random_state)
-        if test and random_state:
-            test.set_seed(random_state)
+        # The copy does 2 things: Sample n_samples if parameter exists, and shuffle the data.
+        # we shuffle because the data in VisionData is in fixed order, so if the user wants to run without random_state
```
noamzbr, line 134:
```suggestion
        # we shuffle because the data in VisionData is set to be sampled in a fixed order, so if the user wants to run without random_state
```

noamzbr (approved):

PR #993, opened 2022-03-10
Validating import order

ItayGabbay:


tests/vision/checks/distribution/image_properties_drift_test.py
```diff
@@ -28,6 +28,7 @@
 from deepchecks.vision.utils.image_formatters import default_image_properties
 from deepchecks.vision.checks.distribution import ImagePropertyDrift
 from deepchecks.vision.datasets.detection import coco
+from deepchecks.vision.utils.image_formatters import image_properties
```
JKL98ISR, line 31: needs to be removed

Issue #1022, opened 2022-03-15
[BUG][CV] Refactor usage of `n_samples_per_class` out of `initalize_run`


matanper:
**Describe the bug**
When calling `n_samples_per_class` we generate a second iteration over the data. in order to combine the big update loop with the n_samples_per_class aggergation we need to remove the usage of it from `initalize_run` and only use it in `compute` after the update loop ended.

1. Refactor `n_samples_per_class` out of the following checks initalization:
- [x] HeatmapComparison
- [x] ConfusionMatrixReport
- [x] SimpleModelComparison

2. 
- [x] Add method "update_cache" on the VisionData which accepts batch and updates the `n_samples_per_class` cache, and call it in the update_loop of suite and in run loops of checks.

noamzbr:
Regarding the second task, @yromanyshyn please update if this is resolved by the solution to #985.

PR #1024, opened 2022-03-15
Add map of class to list of dataset index

matanper:
#### Reference Issues/PRs

solves num. 2 in #1022 
and also num. 1 of Heatmap Comparison and Simple Model Comparison

Instead of caching the number of samples per class, caching for each class the batch indices of samples that have this class (in the label).
Also added the the cache update to the larger update loops, so only remains to remove the ad-hoc calculation after removing the usage of it from the init_run of checks


deepchecks/vision/vision_data.py
```diff
@@ -255,9 +278,11 @@ def copy(self, n_samples: int = None, shuffle: bool = False, random_state: int =
                                                                   shuffle=shuffle,
                                                                   random_state=random_state,
                                                                   n_samples=n_samples)
-        # If new data is sampled, then needs to remove cached info
+        # If new data is sampled, then needs to re-calculate cache
         if n_samples:
-            new_vision_data._n_of_samples_per_class = None
+            new_vision_data.init_cache()
+            for batch in new_vision_data:
+                new_vision_data.update_cache(self.batch_to_labels(batch))
```
noamzbr, line 285: When does this happen? because we use copy to create the sampled data obj?

matanper: yes

deepchecks/vision/classification_data.py
```diff
@@ -123,7 +123,7 @@ def infer_on_batch(self, batch, model, device) -> Union[List[torch.Tensor], torc
 
     def get_classes(self, batch_labels: Union[List[torch.Tensor], torch.Tensor]):
         """Get a labels batch and return classes inside it."""
-        return batch_labels.tolist()
+        return batch_labels.reshape(-1, 1).tolist()
```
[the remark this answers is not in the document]
matanper: Looks like the tests are passing, it is in order to return list per sample

deepchecks/vision/checks/performance/simple_model_comparison.py
```diff
@@ -109,39 +109,13 @@ def __init__(self,
...
-                self._state['dummy_prediction_generator'] = \
-                    lambda: torch.from_numpy(np.random.multinomial(1, class_prior))
-            elif self.strategy == 'uniform':
-                self._state['dummy_prediction_generator'] = \
-                    lambda: torch.from_numpy(np.ones(context.train.num_classes) / context.train.num_classes)
-            else:
-                raise DeepchecksValueError(
-                    f'Unknown strategy type: {self.strategy}, expected one of {_allowed_strategies}.'
-                )
+        self._test_metrics = get_scorers_list(context.train, self.alternative_metrics)
```
ItayGabbay, line 118: Can you just add a simple if to check whether the strategy is valid?

matanper: This `if` exists in the init

ItayGabbay (approved):

PR #1029, opened 2022-03-16
[Issue 985] Confusion matrix report - unseen classes bug fix

### LINKED ISSUE #985: [BUG][CV] Confusion matrix does not support unseen classes
**Describe the bug**
For object detection:
The model might return classes that we haven't seen before from our dataset. For example, our dataset has class ids between 0 to 100. The model can return suddenly 120, and we don't have a row for it in the matrix. Right now, we just put all the "unseen" classes under the same cateogry.

**To Reproduce**
Limit coco dataset to only a few classes and run the check on it

**Expected behavior**
Have each unseen classes in a row of its own.

yromanyshyn:
resolves #985, #1022


deepchecks/vision/checks/performance/confusion_matrix.py
```diff
@@ -72,156 +75,141 @@ def __init__(self,
     def initialize_run(self, context: Context, dataset_kind: DatasetKind = None):
         """Initialize run by creating an empty matrix the size of the data."""
         context.assert_task_type(TaskType.CLASSIFICATION, TaskType.OBJECT_DETECTION)
-
-        if dataset_kind == DatasetKind.TRAIN:
-            dataset = context.train
-        else:
-            dataset = context.test
+        dataset = context.train if dataset_kind == DatasetKind.TRAIN else context.test
```
matanper, line 78:
```suggestion
        dataset = context.get_dataset_by_kind(dataset_kind)
```

deepchecks/vision/checks/performance/confusion_matrix.py
```diff
@@ -72,156 +75,141 @@ def __init__(self,
...
-                     x=x,
-                     y=y,
-                     text_auto=True)
-
-        fig.update_layout(width=600, height=600)
-        fig.update_xaxes(title='Predicted Value', type='category')
-        fig.update_yaxes(title='True value', type='category')
-
+                raise RuntimeError(
+                    'Internall Error! categories list mmust '
```
matanper, line 128: typo in must

deepchecks/vision/checks/performance/confusion_matrix.py
```diff
@@ -72,156 +75,141 @@ def __init__(self,
...
+                    'contain items of type - Union[int, Literal["no-overlapping"]]'
+                )
+
+        # NOTE: 'no-overlapping' is at the end of the list
+        x = classes_to_display[:-1]
```
matanper, line 133: don't you need here some `if 'no-overlapping' in classes_to_display`?

matanper (approved):
Looking good

deepchecks/vision/checks/performance/confusion_matrix.py
yromanyshyn: fixed

Issue #1010, opened 2022-03-13
[BUG] Reduce unit tests time


ItayGabbay:
Currently running the unit tests time takes A LOT. find a way to reduce it

PR #1046, opened 2022-03-18
Adding mocks to yolo and mnist

ItayGabbay:
#### Reference Issues/PRs

#1010 

---


ItayGabbay:
Changed the tests value because the order of the data was changed (shuffle=False)

tests/vision/vision_conftest.py
```diff
@@ -45,10 +50,41 @@
...
 
 
+def _hash_image(image):
+    if isinstance(image, np.ndarray):
+        image = Image.fromarray(image)
+    elif isinstance(image, torch.Tensor):
+        image = Image.fromarray(image.cpu().numpy().squeeze())
+
+    image = image.resize((10, 10))
+    image = image.convert('L')
```
matanper, line 66: what is L?

tests/vision/vision_conftest.py
```diff
@@ -45,10 +50,41 @@
...
+
+    pixel_data = list(image.getdata())
+    avg_pixel = sum(pixel_data) / len(pixel_data)
+
+    bits = ''.join(['1' if (px >= avg_pixel) else '0' for px in pixel_data])
```
matanper, line 71: why do you need this transformation?

tests/vision/vision_conftest.py
ItayGabbay: Converting the image to a single-channel image

tests/vision/vision_conftest.py
ItayGabbay: It's just a way to hash an image - if a pixel is above the avg - put 1, otherwise 0.

matanper: but why a string of '0001011101' is better than '3023408509'?

matanper (approved):
IMO the mock should work per-image and not per-batch. It's just more robust and it's about the same complexity to do.

tests/vision/vision_conftest.py
```diff
@@ -129,6 +165,39 @@ def mnist_dataset_train_imgaug():
 def trained_yolov5_object_detection(device):  # pylint: disable=redefined-outer-name
     return load_yolov5_model(device=device)
 
+@pytest.fixture(scope='session')
+def mock_trained_mnist(device):  # pylint: disable=redefined-outer-name
+
+    class MockMnist:
+        def __call__(self, batch):
```
matanper, line 172: instead of having a mock model and a real model, we should have combination that works like:
```
for img in batch:
  key = _hash_batch(batch)
  if key in mnist_predictions_dict:
     return mnist_predictions_dict[key]
  else:
     real_model.to(device)(img)

```

PR #1047, opened 2022-03-20
Custom vision check

### LINKED ISSUE #1012: [DOCS][CV] Custom check tutorial
Create a tutorial showing how to write and run a custom vision check.

noamzbr:
#### Reference Issues/PRs

Resolves #1012 

Add an example of implementing a simple check to compare the amount of RGB between train and test.


noamzbr:
@ItayGabbay I've left there two TODO comments where I need your help

Another question - do we need to explain the Context and Batch object here.

docs/source/user-guide/vision/custom_checks.py
```diff
@@ -0,0 +1,209 @@
...
+
+
+Write a Basic Check
+========================
+
+Let's implement a simple check, comparing the average of each color channel between the train and the test dataset.
+
+We'll start by writing the simplest possible example, returning only a dict of the color averages.
+
+*Good to know: the return value of a check can be any object, a number, dictionary, string, etc…*
```
ItayGabbay, line 55: I think that the prologue is too long

ItayGabbay, line 55:
```suggestion
**Good to know: the return value of a check can be any object, a number, dictionary, string, etc…**
```

docs/source/user-guide/vision/custom_checks.py
```diff
@@ -0,0 +1,209 @@
...
+result
+
+#%%
+# Our check ran successfully, but we got the print “Nothing found”. This is because we haven’t defined to the check
+# anything to display, so the default behavior is to print “Nothing found”. In order to access the value that we have
+# defined earlier we can use the “value” property on the result.
+
+result.value
+
+#TODO: how does this LINK work?
```
ItayGabbay, line 121: what do you mean?

docs/source/user-guide/vision/custom_checks.py
```diff
@@ -0,0 +1,209 @@
...
+#%%
+# Check Display
+# ========================
+#
+# Most of the times we will want our checks to have a visual display that will quickly summarize the check result. We
+# can pass objects for display to the CheckResult. Objects for display should be of type: html string, dataframe or a
+# function that plots a graph. Let’s define a graph that will be displayed using `Plotly <https://plotly.com/>`_. We
+# will inherit from the original check to shorten the code an update only the compute method.
+#
+# *Good to know: ``display`` can receive a single object to display or a list of objects*
```
ItayGabbay, line 135:
```suggestion
# **Good to know: ``display`` can receive a single object to display or a list of objects**
```

docs/source/user-guide/vision/custom_checks.py
```diff
@@ -0,0 +1,209 @@
...
+# %%
+# Voilà! Now we have a check that prints a graph and has a value. We can add this check
+# to any Suite, and it will run within it.
+
+# %%
+# Defining a Condition
+# ========================
+#
+# Finally, we can add a condition to our check. A condition is a function that receives the result of the check and
+# returns a condition result object. To read more on conditions, check out user guide (LINK). # TODO: how to link to configure a check condition user guide?
```
ItayGabbay, line 175: links in rst are in the following:

```
`user guide <../general/customizations/configure_check_conditions>`_

**Review by ItayGabbay — CHANGES_REQUESTED** [review]
Looks Good!
I think the inheritance from step to step is a bit confusing. Maybe it's better to define outer functions that are responsible for the logic, then implement each time the same class...?

File: docs/source/user-guide/vision/custom_checks.py (hunk 4 above)
Comment (noamzbr) [reply docs/source/user-guide/vision/custom_checks.py:175]: need `ref: ` before?

File: docs/source/user-guide/vision/custom_checks.py (hunk 2 above)
Comment (noamzbr) [reply docs/source/user-guide/vision/custom_checks.py:121]: How do I link to the root of the vision API Reference?

File: docs/source/user-guide/vision/custom_checks.py (hunk 4 above)
Comment (ItayGabbay) [reply docs/source/user-guide/vision/custom_checks.py:175]: Sometimes. Let's try without it and see.

File: docs/source/user-guide/vision/custom_checks.py (hunk 2 above)
Comment (ItayGabbay) [reply docs/source/user-guide/vision/custom_checks.py:121]:
```
`API ref <../../api/deepchecks.vision>`_

ItayGabbay (approved):

PR #1069, opened 2022-03-21
[issue-1016] updated image property drift

### LINKED ISSUE #1016: Add images to check displays
### DoD:
Add images to relevant checks.

### Description:
Change check *logic* when needed - 
1. If possible, keep the relevant image data in the "update" function itself (e.g. in single-property drift, we can save the "extreme" examples as we run)
2. If not possible, in compute, extract the relevant image ids and load the images there.
Note: would be best to have a unified method to extract images.

### Tasks:
ImagePropertyDrift
After the graphs, add examples (from user's images) of different images on different scales (5 images exemplifying brightness, then 5 exemplifying aspect ratio etc.)
Other option - one table of ~10 images and below them all the relevant values per property (takes up less space, but will look worse probably)

ClassPerfromance
Add photos under graph of images the model was successful in and images that it wasn't

ConfusionMatrixReport
add images that are examples of misclassificaitons (from the slots with highest numbers)

ImageSegmentPerformance
Add images that exemplify each segment (at least 3 per segment). Try and make the images different (use similarity algorithm from SimilarImageLeakage)

yromanyshyn:
resolves #1016 

- [x] ImagePropertyDrift
- [x] ClassPerfromance
- [x] ConfusionMatrixReport
- [ ] ImageSegmentPerformance


deepchecks/vision/checks/distribution/image_property_drift.py
```diff
@@ -168,17 +180,87 @@ def compute(self, context: Context) -> CheckResult:
...
+            train_thumbnail_images, *_ = context.train.batch_of_index(*list(train_samples.index))
+            test_thumbnail_images, *_ = context.test.batch_of_index(*list(train_samples.index))
+
+            thumbnails = self._prepare_thumbnails_block(
+                train_properties=train_samples.T,
+                test_properties=test_samples.T,
+                train_images=train_thumbnail_images,
+                test_images=test_thumbnail_images
+            )
+            displays = [headnote] + [figures[col] for col in columns_order] + [thumbnails]
```
ItayGabbay, line 194: are we simply displaying 10 random images from each dataset? Shouldn't we choose the images that represent the drift? (if that is possible, at all)

deepchecks/vision/checks/distribution/image_property_drift.py
```diff
@@ -168,17 +180,87 @@ def compute(self, context: Context) -> CheckResult:
...
+        <h4>Test Images</h4>
+        <hr>
+        {test_thumbnails}
+        """)
+
+        return template.format(
+            train_thumbnails=train_table,
+            test_thumbnails=test_table,
+        )
+
```
ItayGabbay, line 263: maybe it's better to have it in a util file? since many checks will have to display thumbnails of images.

deepchecks/vision/checks/distribution/image_property_drift.py
yromanyshyn: > are we simply displaying 10 random images from each dataset? 

yes

>Shouldn't we choose the images that represent the drift?

yeap, you are right/

> if that is possible, at all

do not know/ 
should it be images from that part of the distribution where the difference is the biggest?

deepchecks/vision/checks/distribution/image_property_drift.py
yromanyshyn: this particular function is too specific but I decoupled grid creation into a separate function called `prepare_grid`

deepchecks/vision/checks/distribution/image_property_drift.py
```diff
@@ -123,28 +139,33 @@ def compute(self, context: Context) -> CheckResult:
             value: dictionary containing drift score for each image property.
             display: distribution graph for each image property.
         """
+        assert self._train_properties is not None
```
noamzbr, line 142: This should return a meaningful Deepchecks error in case it isn't true (such as "No samples belonging to classes...")

deepchecks/vision/checks/distribution/image_property_drift.py
```diff
@@ -123,28 +139,33 @@ def compute(self, context: Context) -> CheckResult:
...
+
         if sorted(self._train_properties.keys()) != sorted(self._test_properties.keys()):
             raise RuntimeError('Internal Error! Vision check was used improperly.')
 
         # if self.classes_to_display is set, check that it has classes that actually exist
         if self.classes_to_display is not None:
-            if not set(self.classes_to_display).issubset(
-                    map(self._class_to_string, context.train.classes_indices.keys())
-            ):
+            class_to_string = context.train.label_id_to_name
```
noamzbr, line 150: Seems like this belongs before the assert

deepchecks/vision/checks/distribution/image_property_drift.py
```diff
@@ -171,17 +195,73 @@ def compute(self, context: Context) -> CheckResult:
...
-            displays = [headnote] + [figures[col] for col in columns_order]
-        else:
-            drifts = None
-            displays = []
+            train_samples = df_train.sample(10)
+            test_samples = df_test.sample(10)
+            train_thumbnail_images, *_ = context.train.sample(*list(train_samples.index))
+            test_thumbnail_images, *_ = context.test.sample(*list(train_samples.index))
+
+            thumbnails = self._prepare_thumbnails_block(
```
noamzbr, line 203: From the issue description:

> After the graphs, add examples (from user's images) of different images on different scales (5 images exemplifying brightness, then 5 exemplifying aspect ratio etc.)

So per property show examples of images with different property values. Can either do it from train and then from test, or mix train and test but then tell us which is which.

deepchecks/vision/checks/distribution/image_property_drift.py
```diff
@@ -171,17 +195,73 @@ def compute(self, context: Context) -> CheckResult:
...
+
+            displays = [headnote] + [figures[col] for col in columns_order] + [thumbnails]
 
         return CheckResult(
             value=drifts,
             display=displays,
             header='Image Property Drift'
         )
 
+    def _prepare_thumbnails_block(
```
noamzbr, line 218: Isn't there already a function like this @matanper ?

deepchecks/vision/checks/performance/confusion_matrix.py
```diff
@@ -199,25 +211,178 @@ def update_object_detection(self, predictions, labels):
...
+        if len(misclassified_images) == 0:
+            return ''
+
+        grid_content = [
+            '<span><b>Ground Truth</b></span>',
+            '<span><b>Detected Truth</b></span>',
+            '<span><b>Image</b></span>',
+        ]
+
+        for label, detected, img in misclassified_images[:self.n_of_images_to_show]:
```
noamzbr, line 319: how do we know this won't be all the same "kind" misclassification? Anything here making sure we're gonna see different kinds of error (between different classes)?

noamzbr (commented):
Hi @yromanyshyn, could you attach examples (pdfs) of how the check display looks like after this change?

deepchecks/vision/checks/distribution/image_property_drift.py
yromanyshyn: this is totally internal stuff, and it could happen only if you forgot to initialize `self._train_properties` within `initialize_run`.

I added this to stop the static type checker from complaining that type of the `self._train_properties` is `Optional[Dict[...]]`

deepchecks/vision/checks/distribution/image_property_drift.py
yromanyshyn: sorry, not sure if I understood, do you mean that this 'if' statement should be moved before the assert statements?

deepchecks/vision/datasets/detection/coco.py
```diff
@@ -349,95 +349,84 @@ def yolo_image_formatter(batch):
...
+    69: 'oven',
+    70: 'toaster',
+    71: 'sink',
+    72: 'refrigerator',
+    73: 'book',
+    74: 'clock',
+    75: 'vase',
+    76: 'scissors',
+    77: 'teddy bear',
+    78: 'hair drier',
```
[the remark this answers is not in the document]
yromanyshyn: from yaml in this zip file - [coco128_with_yaml](https://github.com/ultralytics/yolov5/releases/download/v1.0/coco128_with_yaml.zip)

deepchecks/vision/checks/performance/confusion_matrix.py
```diff
@@ -199,25 +211,178 @@ def update_object_detection(self, predictions, labels):
...
+        misclassifications = (
+            (
+                label_class,
+                detected_class,
+                matrix.at[label_class, detected_class]  # count
+            )
+            for label_class, detected_class in product(list(matrix.index), repeat=2)
+            if label_class != detected_class
+        )
+        misclassifications = sorted(
```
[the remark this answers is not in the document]
yromanyshyn: yes, and `self.n_of_images_to_show` control how many will be shown

deepchecks/vision/checks/performance/confusion_matrix.py
yromanyshyn: sorry, I did not understand what you mean.

>same "kind" misclassification

the same pair of classes? how is it possible?

yromanyshyn:
@noamzbr 

> Hi @yromanyshyn, could you attach examples (pdfs) of how the check display looks like after this change?

sorry, I were not able to transform notebook examples to pdfs/ `pandoc` and `nbformat` returned the errors when I tried, therefore I attached a zip with a pair of example notebooks

[nb-examples.zip](https://github.com/deepchecks/deepchecks/files/8368812/nb-examples.zip)

screenshots
-----------------

image-property-drift

image

confusion-matrix

image

deepchecks/vision/checks/distribution/image_property_drift.py
yromanyshyn: I implemented it like this

>Other option - one table of ~10 images and below them all the relevant values per property (takes up less space, but will look worse probably)

deepchecks/vision/checks/performance/confusion_matrix.py
noamzbr: yeah, it can be that there are 100 images in which class 1 was mixed up with class 2. If they appear first, we'll get only them and miss out on examples of class 4 mixed up with class6 6. Can this happen here?

deepchecks/vision/checks/performance/confusion_matrix.py
```diff
@@ -199,25 +211,178 @@ def update_object_detection(self, predictions, labels):
...
+        if self.n_of_images_to_show == 0:
+            return ''
+
+        template = dedent(f"""
+            <h4>Misclassified Images</h4>
+            <p><b>NOTE:</b> in case of object detection task the next colors
+            are used for bbox identification:</p>
+            <ul>
+                <li>{self._LABEL_COLOR} - ground truth</li>
+                <li>{self._DETECTION_COLOR} - detected truth</li>
```
noamzbr, line 279:
```suggestion
                <li>{self._DETECTION_COLOR} - detected class</li>
```

noamzbr:
> @noamzbr
> 
> > Hi @yromanyshyn, could you attach examples (pdfs) of how the check display looks like after this change?
> 
> sorry, I were not able to transform notebook examples to pdfs/ `pandoc` and `nbformat` returned errors when I tried, therefore I attached a zip with a pair of example notebooks
> 
> [nb-examples.zip](https://github.com/deepchecks/deepchecks/files/8368812/nb-examples.zip)
> 

Thanks @yromanyshyn. 
Regarding the property drift - I now understand better, @shir22 what do you think? If we keep this structure, I'd like to save some space by compressing the table (less space between rows). Is that possible?

Regarding the confusion matrix - We need to make it smaller. Can we display the images horizontally and have the details (Ground Truth and Prediction) beneath, like you did in property drift?

yromanyshyn:
@noamzbr 

updated views

image property drift check (changed grid-gap from `2.0rem` to `0.5rem`)

image

confusion matrix

image

[example notebooks zip](https://github.com/deepchecks/deepchecks/files/8370344/nb-examples.zip)

deepchecks/vision/checks/performance/confusion_matrix.py
yromanyshyn: oh, yeah, you are right.
I added `_IMAGES_PER_MISCLASSIFICATION` to fix this

shir22:
Hi, had a look at the display (but didn't see where are the actual properties and drift in the context of the display).

For confusion matrix:
Didn't understand / couldn't see properly the bounding boxes and their texts.
The general format looks good, but I'd want to see a few more pictures (if their bounding box and text is clearer)

For the property drift:
I don't feel this display helps in understanding the "property image drift"
We may consider having a check for displaying the "image properties" for given image/s where we display an image and its properties, but I don't really see the connection in the context of the drift check.
What is the purpose of displaying the images (and is this check relevant for it)? (if it's just to demonstrate the values of the properties then maybe worth showing a small number of handpicked pictures)

noamzbr:
> Hi, had a look at the display (but didn't see where are the actual properties and drift in the context of the display).
> 
> For confusion matrix: Didn't understand / couldn't see properly the bounding boxes and their texts. The general format looks good, but I'd want to see a few more pictures (if their bounding box and text is clearer)
> 
> For the property drift: I don't feel this display helps in understanding the "property image drift" We may consider having a check for displaying the "image properties" for given image/s where we display an image and its properties, but I don't really see the connection in the context of the drift check. What is the purpose of displaying the images (and is this check relevant for it)? (if it's just to demonstrate the values of the properties then maybe worth showing a small number of handpicked pictures)

It's about demonstrating the values of the properties. Instead of showing random images, we can show perhaps two images below each dist plot, one with a large property value and one with a small one

shir22:
> > Hi, had a look at the display (but didn't see where are the actual properties and drift in the context of the display).
> > For confusion matrix: Didn't understand / couldn't see properly the bounding boxes and their texts. The general format looks good, but I'd want to see a few more pictures (if their bounding box and text is clearer)
> > For the property drift: I don't feel this display helps in understanding the "property image drift" We may consider having a check for displaying the "image properties" for given image/s where we display an image and its properties, but I don't really see the connection in the context of the drift check. What is the purpose of displaying the images (and is this check relevant for it)? (if it's just to demonstrate the values of the properties then maybe worth showing a small number of handpicked pictures)
> 
> It's about demonstrating the values of the properties. Instead of showing random images, we can show perhaps two images below each dist plot, one with a large property value and one with a small one

That may make the user expect/assume an "outliers" check IMO

I think if that's the purpose then it should be clear that these examples are just to "demonstrate the property values", and not that we're showing very interesting specific images. (e.g. make it take up less space as well.  One table?) which is the reasonable thing to assume...

yromanyshyn:
update:

==  1  ==
modified `ClasPerformance` check, screenshot below

--- **detection task** ---
note: each gt&dt pair are displayed on they own image

image
image

--- **classification task** ---

image


==  2  ==

@shir22 
> Didn't understand / couldn't see properly the bounding boxes and their texts.
The general format looks good, but I'd want to see a few more pictures (if their bounding box and text is clearer)

yup, if a bbox is small, it is very hard to understand what is outlined (shown)
with red border - ground truth bbox
with blue border - detected truth bbox
class id is drawn at the top center of a bbox
(note: there was a bug which is present on the screenshot, instead of class id, ious was drawn at the center of bbox)

== 3 ==

@noamzbr @shir22 
> > For the property drift: I don't feel this display helps in understanding the "property image drift" We may consider having a check for displaying the "image properties" for given image/s where we display an image and its properties, but I don't really see the connection in the context of the drift check. What is the purpose of displaying the images (and is this check relevant for it)? (if it's just to demonstrate the values of the properties then maybe worth showing a small number of handpicked pictures)
> 
> It's about demonstrating the values of the properties. Instead of showing random images, we can show perhaps two images below each dist plot, one with a large property value and one with a small one

so, how it should look?

yromanyshyn:
the issue is still actual
but implementation probably not 
closing it at least for now

Issue #1015, opened 2022-03-13
Check display improvements


nirhutnik:
### DoD:
All relevant checks' graphs look pretty

### Tasks:
General
In all drift checks - 
graph "value" x-axis should have the actual property name

HeatmapComparison
Change graph title to "difference" instead of minus
If there is not significant difference - write "no difference" instead of showing black square
Pixel value should be shown together with all images (right now need to scroll sideways for it)
Suggestion - change color scheme for difference and see if that looks better
train and test squares are shaped differently - fix

ClassPerformance
Graphs are too tall
Colors are wrong - should be more deepchecky and less ugly
AR and AP should be full names (not acronym)
classification - class names are bad in example. If classes are ordinal, keep the order (logic change)
Can we handle many class categories? If not, see decide how to cut and what to show

MeanAveragePrecisionReport
graph is ugly - make prettier (less scikit-learny. Maybe different background, stronger line, etc.)
write "IoU threshold" in graph (x-axis)

MeanAverageRecallReport
graph is missing from notebook - check what happened to the graph
+All comments from mAP Report

SimpleModelComparison
"Simple model" is not a dataset (see legend)
value table in notebook - indexes are unclear, output columns should be arranged differently ("value" at end. Also, rename "value" to something indicative)
Performance example in notebook is not good, should try and pick something that is more interesting, or a different method

ConfusionMatrixReport
Try and fidget with the numbers to create more interesting conf matrix

RobustnessReport
Something got ruined in the docs - check what
check value looks weird
blue and red graphs - maybe more deepchecky colors? 
ticks of y axis in performance comparison should be like in top affected classes
example images - brightness on mnist is weird example. Shift scale is not always very clear - 3 looks the same
Usert wants to know the exact parameters of the augmentation and what run.
-100% looks bad and is not clear. 

ImageSegmentPerformance
Put a table that explains the different segments before or after the graphs

JKL98ISR:
RobustnessReport
wdym with the y ticks? looks fine to me
"-100% looks bad and is not clear." ...

nirhutnik:
deepchecks colors - as in drift graphs.
many class categories - Meaning how does this check handle when there are 100+ categories to show? Are the graphs ready for that?

MeanAverageRecallReport has no graph - Right, and, well, that's a problem...

MeanAveragePrecisionReport - there are suggestions there. It looks very matlab-y at the moment.

ImageSegmentPerformance - The data on the segment itself. which properties and what values define each segment.

[...]

PR #1073, opened 2022-03-21
change x axis name to column name instead of value in drift graphs

benisraeldan:
#### Reference Issues/PRs

part of #1015 

Renames the X-axis of distribution graphs from "value" to the `column name`

### DIFF
--- a/deepchecks/utils/distribution/drift.py
+++ b/deepchecks/utils/distribution/drift.py
@@ -10,7 +10,7 @@
 #
 """Common utilities for distribution checks."""
 
-from typing import Tuple, Union, Hashable, Callable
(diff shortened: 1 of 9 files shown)


deepchecks/utils/distribution/drift.py
```diff
@@ -107,10 +108,12 @@ def calc_drift_and_plot(train_column: pd.Series, test_column: pd.Series, plot_ti
         column from train dataset
     test_column : pd.Series
         same column from test dataset
-    plot_title : Hashable
+    value_name : Hashable
```
JKL98ISR, line 111: Switch the names

JKL98ISR (approved):

PR #1078, opened 2022-03-22
Suggestion: Add assert in vision objects to Batch

### LINKED ISSUE #1058: [BUG][CV] Missing data validation in vision checks
**Describe the bug**
Most checks validate the task type, but they do not validate that the data has the properties necessary for running the check, such as a valid batch_to_images method, or batch_to_labels. 

**Expected behavior**
Look at the example validation in the `initialize_run `method of image_property_drift.py
We want to add such validations to all checks.

matanper:
resolve #1058

Problem:
If there are problems with the formatters we want our checks to fail with an informative exception instead of some un-expected one. For example if user did image formatting with transposed shape (channels in start instead of end) then the check will run and will fail with some unknown expected when the image format will not fit what the check is expecting.

Pros of suggested solution:
Seamless assert, no code added in the checks

Cons:
The call to "assert" function is on every call to batch, if we put it in the check's `initialize_run` function it will be called once

### DIFF
--- a/deepchecks/vision/checks/distribution/image_property_drift.py
+++ b/deepchecks/vision/checks/distribution/image_property_drift.py
@@ -70,14 +70,15 @@ class ImagePropertyDrift(TrainTestCheck):
         self.max_num_categories = max_num_categories
         self.classes_to_display = classes_to_display
         self.min_samples = min_samples
-        self._train_properties = defaultdict(list)
-        self._test_properties = defaultdict(list)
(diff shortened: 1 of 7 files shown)


JKL98ISR (approved):

noamzbr:
Not sure about the cons. Seems like it's called only the first time image/label/predictions are called in any check, not afterwards. Seems fine if this is indeed the case.

PR #1080, opened 2022-03-22
json html cond fix + condition refactor

### LINKED ISSUE #928: [BUG] Json import throws exception
1. We have double export of json when exporting suite result as json. First the check results dumps as json, and then the suite result. the dump should happen only once.
2. Our docs mentions to use `json` to load results but when loading with json we get jsonpickle dicts:
`{'py/tuple': ['html', '<b>some text</b>:']}` and when loading jsonpickle some of the results throws exception `TypeError: the first argument must be callable`

Steps to reproduce:
Run full suite on iris dataset (tabular) and then try to export and import:
```
result = suite.run(train_dataset=ds_train, test_dataset=ds_test, model=rf_clf)
str = result.to_json()

jsonloaded = jsonpickle.loads(str)
for s in jsonloaded['results']:
    print(jsonpickle.loads(s))
```

JKL98ISR:
close #928 , #1096

 fix html in condition table and fix pandas error


matanper:
maybe should add test that is doing save json and then loads

JKL98ISR:
> maybe should add test that is doing save json and then loads

added

deepchecks/core/check_result.py
```diff
@@ -36,6 +37,7 @@
 from pandas.io.formats.style import Styler
 
 
+jsonpickle_pd.register_handlers()
```
matanper, line 40: Add comment what it does?

deepchecks/core/condition.py
```diff
@@ -58,6 +58,7 @@ class ConditionCategory(enum.Enum):
 
     FAIL = 'FAIL'
     WARN = 'WARN'
+    PASS = 'PASS'
```
matanper, line 61: when doing inside condition function `return ConditionResult(...)` we will be able to define the category as "passed" + we have the is_pass boolean. Isn't that confusing?

JKL98ISR: Fixed it + fixed confusing category when condition is passed (used to be failed when passed)

JKL98ISR:
further improvement needed as explained in #1096

matanper (approved):

PR #1094, opened 2022-03-23
adding code owners

ItayGabbay:


shir22 (approved):

matanper (approved):
Now that I see it's not so useful, it's basically just limits the docs and there is no difference for all the code

PR #1199, opened 2022-04-05
Fix drift plots for few unique numerical values

### LINKED ISSUE #1129: [BUG] [CV] Image property drift graphs are empty when no drift
**Describe the bug**
 Image property drift graphs are empty when no drift

**To Reproduce**
Run the check on data with same image sizes

**Expected behavior**
To not show the graphs

**Screenshots**
Screen Shot 2022-03-29 at 16 15 17

matanper:
resolve #1129

if less than 20 uniques values in both train + test show it as bar graph.
else if one of train/test have less than 5 unique values, than show it as bars alongside KDE plot of the second one

image

image

image


deepchecks/utils/distribution/drift.py
```diff
@@ -160,9 +160,13 @@ def calc_drift_and_plot(train_column: pd.Series,
         # Should never reach here
         raise DeepchecksValueError(f'Unsupported column type for drift: {column_type}')
 
+    dist_trace_type = 'table' if dist_traces[0].type == 'table' else 'xy'
+    dist_trace_title = 'Single Value in Datasets' if dist_trace_type == 'table' else 'Distribution Plot'
```
nirhutnik, line 164: Title is very unclear. 
It's a statement without context, and as a user I wouldn't know what it means. I think you need to elaborate more.

nirhutnik: However, why even show this? If there's a single value and it's identical - why show the drift graph at all?

matanper: if it's not identical?

matanper: Nir suggestion:
If single values are the same doesn't show any plot.
If single values are different then show them as bars side by side

@shir22 WDYT?

deepchecks/utils/distribution/plot.py
```diff
@@ -204,14 +180,98 @@ def sort_int_and_strings(a, b):
...
-        traces = [go.Scatter(x=xs, y=get_density(train_column, xs), fill='tozeroy', name='Train Dataset',
-                             line_color=colors['Train']),
-                  go.Scatter(x=xs, y=get_density(test_column, xs), fill='tozeroy', name='Test Dataset',
-                             line_color=colors['Test'])]
+        train_density = get_density(train_column, xs)
+        test_density = get_density(test_column, xs)
+        bars_width = (x_range_to_show[1] - x_range_to_show[0]) / 100
+
+        traces = []
+        if train_uniques.size <= 5:
```
noamzbr, line 188: same here, magic number to Const

noamzbr (approved):

PR #1202, opened 2022-04-06
Adding anonymous analytics tracking

ItayGabbay:
Adding anonymous analytics tracking in order to track usage of the package.
Would love to hear your honest opinions about that. 

---

### DIFF
--- a/deepchecks/__init__.py
+++ b/deepchecks/__init__.py
@@ -97,18 +97,12 @@ except:  # pylint: disable=bare-except # noqa
     __version__ = ''
 
 
-# Check for latest version
+# Send an import event
 try:
-    send_anonymous_import_event()
-    disable = os.environ.get('DEEPCHECKS_DISABLE_LATEST', 'false').lower() == 'true'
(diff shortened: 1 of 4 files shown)


deepchecks/__init__.py
```diff
@@ -96,17 +97,12 @@
     __version__ = ''
 
 
-# Check for latest version
+# Send an import event
 try:
-    disable = os.environ.get('DEEPCHECKS_DISABLE_LATEST', 'false').lower() == 'true'
+    disable = not os.environ.get("DEEPCHECKS_ANONYMOUS_TRACKING", True)
```
matanper, line 102:
```suggestion
    disable = os.environ.get("DEEPCHECKS_ANONYMOUS_TRACKING", False)
```

shir22:
When do we generate a .user id and set the env variable to true?

ItayGabbay:
The .user_id is generated when sending the first event to posthog. 

We are sending an event if the env variable is

PR #1205, opened 2022-04-06
Update tabular drift checks to raise correct error without features

### LINKED ISSUE #1133: [BUG] TrainTestFeatureDrift doesn't display anything for dataframe data
**Describe the bug**
When running the check on naïve iris dataframe, we get no display at all, rather than a display showing the exact same distribution which will happen if constructing a Dataset from the dataframe prior.

**To Reproduce**

```
import pandas as pd
from deepchecks.tabular.dataset import Dataset
from deepchecks.tabular.checks import TrainTestFeatureDrift

iris = pd.read_csv('https://raw.githubusercontent.com/mwaskom/seaborn-data/master/iris.csv')
TrainTestFeatureDrift().run(iris, iris)

ds = Dataset(iris)
TrainTestFeatureDrift().run(ds , ds )

```

**Expected behavior**
Exact same behavior for the two cases. 

**Screenshots**
image

matanper:
close #1133 

Right now when passing DataFrame we support only integrity checks. any ML-related check which requires to know which columns are features, label, etc, requires to get a Dataset. 

The error raised is:
DeepchecksNotSupportedError: There are no features defined to use. Did you pass a DataFrame instead of a Dataset?

### DIFF
--- a/deepchecks/tabular/checks/distribution/train_test_feature_drift.py
+++ b/deepchecks/tabular/checks/distribution/train_test_feature_drift.py
@@ -105,6 +105,8 @@ class TrainTestFeatureDrift(TrainTestCheck):
         train_dataset: Dataset = context.train
         test_dataset: Dataset = context.test
         features_importance = context.features_importance
+        train_dataset.assert_features()
+        test_dataset.assert_features()
 
         train_dataset = train_dataset.select(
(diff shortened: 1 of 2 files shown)


noamzbr (approved):
Fine by me, though I think we need #1208 to compensate for that.

PR #1209, opened 2022-04-06
Fix the logic finding the weak segment in categorical features.

### LINKED ISSUE #1183: [BUG] Model Error Analysis tabular example doesn't catch anything
**Describe the bug**
Seems that after updating the log loss function, the example ceased catching the weak segments.

noamzbr:
#### Reference Issues/PRs

Fixes #1183

### DIFF
--- a/deepchecks/utils/performance/error_model.py
+++ b/deepchecks/utils/performance/error_model.py
@@ -179,12 +179,14 @@ def error_model_display(error_fi: pd.Series,
 
             # Partition data into two groups - weak and ok:
 
-            # The weak segment is all the weakest categories accumulated together until they comprise at least
-            # min_segment_size of the total data.
+            # In cum_sum_ratio the index is the categories sorted from "weakest" (the highest model error) to
+            # strongest, and the value is the cumulative fraction of all data.
+            # The weak segment contains all the weakest categories until they reach together a fraction of data of at
+            # least min_segment_size.
             first_weakest_category_to_pass_min_segment_size = np.where(cum_sum_ratio.values >= min_segment_size)[0][0]
-            in_segment_indicis = np.arange(len(cum_sum_ratio)) <= first_weakest_category_to_pass_min_segment_size
-            weak_categories = error_per_segment_ser.index[in_segment_indicis]
-            ok_categories = error_per_segment_ser.index[~in_segment_indicis]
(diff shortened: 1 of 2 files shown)


deepchecks/utils/performance/error_model.py
```diff
@@ -179,7 +179,9 @@ def error_model_display(error_fi: pd.Series,
 
             # Partition data into two groups - weak and ok:
 
-            in_segment_indicis = cum_sum_ratio <= min_segment_size
+            # The weak segment is all the weakest categories accumulated together until they comprise at least
+            # min_segment_size of the total data.
+            in_segment_indicis = np.arange((cum_sum_ratio >= min_segment_size)[0])
```
matanper, line 184: `(cum_sum_ratio >= min_segment_size)` creates boolean array.
`(cum_sum_ratio >= min_segment_size)[0]` returns the first boolean in the array. 
what am I missing?

noamzbr: Instead of creating an array of all categories whose cumulative count is above min_segment_size (which, if the first is above min_segment_size, will cause all the data to be in the second segment), we create an array of all categories **up to** the first category whose cumulative count is above min_segment_size, thus splitting the data into a segment of "all the weakest categories taking at least min_segment_size of the data" and "all the rest".

nirhutnik: I'm with Matan. I still don't understand why you take the first item in the array and feed it to np.arange.

deepchecks/utils/performance/error_model.py
```diff
@@ -179,7 +179,9 @@ def error_model_display(error_fi: pd.Series,
 
             # Partition data into two groups - weak and ok:
 
-            in_segment_indicis = cum_sum_ratio <= min_segment_size
+            # The weak segment is all the weakest categories accumulated together until they comprise at least
+            # min_segment_size of the total data.
```
nirhutnik, line 183: I think the phrasing here can use some touch-ups + be clearer, but for that I still need to understand what's happening there 😅

matanper (approved):

deepchecks/utils/performance/error_model.py
noamzbr: Y'all were right, fixed

ItayGabbay (approved):

PR #1211, opened 2022-04-07
mixed nulls on non string columns

### LINKED ISSUE #1074: [FEAT] Mixed Nulls - catch mixed nulls on non string columns (np.NaN/pd.NaN/Null/None/etc.)
**Docs**

API Reference should include what are the DEFAULT NULLS that are checks (currently this can be found only by going into source)
Example notebook can also print out these nulls for convenience.

**Null Types**
1. NaT nulls not caught @chelseatroy can you elaborate?
3. Seems that list currently includes only strings (and null character). Does this catch also null objects? (e.g. the python None. Numpy and pandas nulls. or any other null that is likely to find it's way due to multiple feature engineering backends)

benisraeldan:
#### Reference Issues/PRs

resolves #1074 

find mixed nulls in non-string columns

### DIFF
--- a/deepchecks/tabular/checks/integrity/mixed_nulls.py
+++ b/deepchecks/tabular/checks/integrity/mixed_nulls.py
@@ -9,7 +9,7 @@
 # ----------------------------------------------------------------------------
 #
 """Module contains Mixed Nulls check."""
-from collections import defaultdict
+from collections import defaultdict, Counter
 from typing import Union, Dict, List, Iterable
 
 import pandas as pd
@@ -84,15 +84,12 @@ class MixedNulls(SingleDatasetCheck):
         # Result value
         display_array = []
         result_dict = defaultdict(dict)
-        display_prefix = ''
 
         for column_name in list(df.columns):
             column_data = df[column_name]
             # Pandas version 1.3.X and lower doesn't support counting separate NaN values in value_counts
             if parse_version(pd.__version__) < parse_version('1.4.0'):
-                column_counts = defaultdict(int)
-                for value in column_data:
-                    column_counts[value] += 1
+                column_counts = Counter(column_data)
             else:
                 # Get counts of all values in series including NaNs
                 column_counts: pd.Series = column_data.value_counts(dropna=False)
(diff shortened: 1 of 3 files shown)


matanper (approved):

deepchecks/utils/ipython.py
```diff
@@ -78,3 +78,8 @@ def is_widgets_enabled() -> bool:
             return not found_disabled and found_enabled
         except Exception:  # pylint: disable=broad-except
             return False
+
+
+def version_tuple(version):
```
matanper, line 83: you don't need this you can use `from pkg_resources import parse_version`

deepchecks/tabular/checks/integrity/mixed_nulls.py
```diff
@@ -79,23 +79,42 @@ def run_logic(self, context: Context, dataset_type: str = 'train') -> CheckResul
...
+                        display_prefix = \
+                            'Pandas version 1.3.X and lower does not support counting different NaN types as separate.'
+
+            else:
+                # Get counts of all values in series including NaNs, in sorted order of count
+                column_counts: pd.Series = column_data.value_counts(dropna=False)
+                # Filter out values not in the nulls list
+                null_counts = {value: count for value, count in column_counts.items()
+                               if (self.check_nan and pd.isnull(value)) or (string_baseform(value) in null_string_list)}
+                num_unique_nans = len(null_counts)
```
matanper, line 115:
```suggestion
            # Pandas version 1.3.X and lower doesn't support counting separate NaN values
            if _versiontuple(pd.__version__) < _versiontuple('1.4.0'):
                column_counts = Counter(column_data)
            else:
                # Get counts of all values in series including NaNs, in sorted order of count
                column_counts = column_data.value_counts(dropna=False)
            # Filter out values not in the nulls list
            null_counts = {value: count for value, count in column_counts.items()
                           if (self.check_nan and pd.isnull(value)) or (string_baseform(value) in null_string_list)}
            num_unique_nans = len(null_counts)
```

matanper (approved):

PR #1214, opened 2022-04-07
1141 bug cv unclear if the prediction was correct in classification data validation

matanper:
1. Remove the usage of plotly from validation. Also removed kaleido from requirements
2. Display for classification only the prediction class and not array
3. Refactored bbox_notation to support also predictions (added score token)
4. For detection shows label & prediction in the same image

image

image


JKL98ISR:
maybe at a comma there? or somthing
i.e label class 6. prediction class 6

ItayGabbay (approved):

PR #1229, opened 2022-04-10
Add tests for docs plots on GPU

### LINKED ISSUE #1200: [BUG][CV] example code is not tested on gpu
**Describe the bug**
Example code is tested only on cpu. If it was tested on gpu, we would have noticed that the line:
`result = suite.run(training_data, val_data, model, device)
`
In plot_detection_tutorial places the device in the wrong place / without a keyname.

**To Reproduce**
Run plot_detection_tutorial  on gpu

matanper:
close #1200


tests/vision/docs/test_plots.py
```diff
@@ -0,0 +1,15 @@
+from runpy import run_path
+import glob
+
+import torch
+
+
+def test_plots_on_gpu():
+    """If there is GPU available running all the docs plot files. Only makes sure the plots don't crash, and not \
+    testing any other display or functionality."""
+    if torch.cuda.is_available() or True:
```
ItayGabbay, line 10: or True?

tests/vision/docs/test_plots.py
```diff
@@ -0,0 +1,15 @@
...
+        plots_files = glob.glob('../docs/source/**/plot_*.py', recursive=True)
```
ItayGabbay, line 11: maybe better to work with Pathlib or os.path to support all OS

tests/vision/docs/test_plots.py
matanper: oops, added for testing

tests/vision/docs/test_plots.py
matanper: how do you know glob doesn't support all OS?

ItayGabbay: glob is fine. but the path you specified as parameter - better use os.path

matanper: I don't really understand why it's better, but updated

ItayGabbay (approved):

Issue #996, opened 2022-03-10
[FEAT][CV] Vision readme


noamzbr:
Adapt readme for CV

noamzbr:
Also add value demonstration section in the docs for CV @shir22

PR #1231, opened 2022-04-10
Update README to accommodate beta release of vision package

shir22:
Updates main package readme
Add first version of vision readme

#### Reference Issues/PRs
Implements #996


README.rst
```diff
@@ -83,7 +91,7 @@ Using conda
 Head over to the `Quickstart Notebook <https://docs.deepchecks.com/en/stable/
 examples/guides/quickstart_in_5_minutes.html?
 utm_source=github.com&utm_medium=referral&utm_campaign=readme&utm_content=try_it_out>`__
-and choose the  |binder badge image|  or the  |colab badge image|  to have it up and running, and to then apply it on your own data and models.
+and see deepchecks output on a built-in dataset, or runi t yourself to apply it on your own data and models.
```
ItayGabbay, line 94:
```suggestion
and see deepchecks output on a built-in dataset, or run it yourself to apply it to your own data and models.
```

ItayGabbay (commented):
It looks really cool!
a few comments:
1. Maybe better to have a short readme as the main one, then one readme for tabular and one for vision?
2. pip install deepchecks[vision] as the main command?

README.rst
shir22: Thanks for the correction, will commit
About number 1 - we discussed previously both alternatives and I implemented it accordingly. Looking at the result I think it's good cause it's "efficient" in demonstrating the value + examples.
Anyway - if we want to reconsider and change let's do it when CV is out of beta.

ItayGabbay (approved):

PR #1251, opened 2022-04-12
Image segment graph improvements

JKL98ISR:
closes #

### DIFF
--- a/deepchecks/vision/checks/performance/image_segment_performance.py
+++ b/deepchecks/vision/checks/performance/image_segment_performance.py
@@ -188,7 +188,7 @@ class ImageSegmentPerformance(SingleDatasetCheck):
             .update_traces(width=bar_width)
          )
 
-        return CheckResult(value=dict(display_df), display=fig)
+        return CheckResult(value=dict(result_value), display=fig)
 
     def _create_bins_and_metrics(self, batch_data: t.List[t.Tuple], dataset):
         """Return dict of bins for each property in format \
--- a/deepchecks/vision/checks/performance/image_segment_performance.py
+++ b/deepchecks/vision/checks/performance/image_segment_performance.py
@@ -48,6 +48,8 @@ class ImageSegmentPerformance(SingleDatasetCheck):
         Maximum number of bins to segment a single property into.
     number_of_samples_to_infer_bins : int, default : 1000
         Minimum number of samples to use to infer the bounds of the segments' bins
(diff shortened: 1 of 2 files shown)


matanper:
Please add screenshot

deepchecks/vision/checks/performance/image_segment_performance.py
```diff
@@ -148,14 +149,23 @@ def compute(self, context: Context, dataset_kind: DatasetKind) -> CheckResult:
...
+                    for metric, val in single_bin['metrics'].items():
+                        display_data.append({'Metric': metric, 'Value': val, **bin_data})
                 # Save for result
                 result_value[property_name].append(single_bin)
 
         display_df = pd.DataFrame(display_data)
-
+        if display_df.empty:
+            return CheckResult(value=dict(result_value))
+        first_metric = list(get_scorers_list(dataset, self.alternative_metrics).keys())[0]
```
matanper, line 163: why not take metric column from the dataframe?

JKL98ISR: not 100% sure if it would be first. probably will, doesn't matter too much, can change it

JKL98ISR:
Screen Shot 2022-04-12 at 14 08 51
AP/AR rename should be another issue imo, it changes too many files and other visualizations we have.
also still all metrics to not have different behaviors on default vs alternative scorers

matanper:
> Screen Shot 2022-04-12 at 14 08 51 AP/AR rename should be another issue imo, it changes too many files and other visualizations we have. also still all metrics to not have different behaviors on default vs alternative scorers

I'm not sure we should have 2 metrics on the same graph. In tabular we had metrics which have different ranges, and it caused problems when displayed together.

JKL98ISR:
> > Screen Shot 2022-04-12 at 14 08 51 AP/AR rename should be another issue imo, it changes too many files and other visualizations we have. also still all metrics to not have different behaviors on default vs alternative scorers
> 
> I'm not sure we should have 2 metrics on the same graph. In tabular we had metrics which have different ranges, and it caused problems when displayed together. For example the AP/AR is 0 to 1, but if a user adds custom metric of "error" it might have value of 2000, and than all the AP plots will be too small to see

so put it back as it was and on default only first metric if no alternative metrics?

JKL98ISR:
[
Screen Shot 2022-04-12 at 18 49 32
](url)

PR #1252, opened 2022-04-12
allow checks to return CheckFailure

### LINKED ISSUE #1250: [FEAT] Model Error Analysis should return CheckFailure instead of raising an error
Raising `DeepchecksProcessError` can break a pipeline

`CheckFailure` should be returned in case of an issue in the check instead of `DeepchecksProcessError`, to inform the user without harming the rest of the run. 

Raising an Error can break a user pipeline, catching the error and printing out warning can be confusing.

benisraeldan:
#### Reference Issues/PRs

Resolves #1250 

`ModelErrorAnalysis` check returns `CheckFailure` instead of raising `DeepchecksProcessError`
also allow other checks to return CheckFailure when needed.

#### Any other comments?

---

### DIFF
--- a/deepchecks/core/check_result.py
+++ b/deepchecks/core/check_result.py
@@ -503,4 +503,13 @@ class CheckFailure:
         """Return string representation."""
         tb_str = traceback.format_exception(etype=type(self.exception), value=self.exception,
                                             tb=self.exception.__traceback__)
-        return self.check.name() +': ' + str(self.exception)
+        return ''.join(tb_str)
+
+    def _ipython_display_(self):
+        """Display the check failure."""
+        check_html = f'<h4>{self.header}</h4>'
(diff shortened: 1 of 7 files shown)


deepchecks/core/check_result.py
```diff
@@ -503,4 +503,4 @@ def __repr__(self):
         """Return string representation."""
         tb_str = traceback.format_exception(etype=type(self.exception), value=self.exception,
```
matanper, line 504: don't you need to remove this var?

matanper (commented):
looking good

deepchecks/core/check_result.py
benisraeldan: Just checked how it looked, left it as is, `print(check_failure)` will still display the whole error trace

matanper (approved):

PR #1270, opened 2022-04-13
Get exception traceback in suite easily

### LINKED ISSUE #1169: [FEAT] Attach exception traces to suite result.
Currently it's often hard to debug exceptions in checks running as parts of suites - all you can see is the error message. 

Attach the exception trace as part of the result object.

benisraeldan:
#### Reference Issues/PRs

Resolves #1169

add `get_errors()` func to suite results for debugging. and print help message in checks that weren't displayed.
Screen Shot 2022-04-13 at 17 11 58

### DIFF
--- a/deepchecks/core/display_suite.py
+++ b/deepchecks/core/display_suite.py
@@ -116,6 +116,7 @@ def _display_suite_widgets(summary: str,
                            checks_wo_conditions_display: List[CheckResult],
                            checks_w_condition_display: List[CheckResult],
                            others_table: List,
+                           contains_check_failures: bool,
                            light_hr: str,
                            html_out,
                            requirejs: bool = True):  # pragma: no cover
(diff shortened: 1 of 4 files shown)


matanper (commented):
I don't understand what "contains_check_failures" is used for
also what is the debug message used for

matanper:
> I don't understand what "contains_check_failures" is used for also what is the debug message used for

NVM, got it but seems too complicated. Why not just add the message as static in the failure tabs?

deepchecks/core/display_suite.py
```diff
@@ -168,7 +169,8 @@ def _display_suite_widgets(summary: str,
         with warnings.catch_warnings():
             warnings.simplefilter(action='ignore', category=FutureWarning)
             others_df = dataframe_to_html(others_table.style.hide_index())
-        h2_widget = widgets.HTML(_CHECKS_WITHOUT_DISPLAY_TITLE)
+        debug_message = '</br>For call traces use get_errors() on result object</br>' if contains_check_failures else ''
```
matanper, line 172: Phrasing is a bit weird "For call traces"

benisraeldan:
> Why not just add the message as static in the failure tabs?
the `Other Checks that wern't Displayed` section also contains check without output and without condictions "nothing found checks" so it would be weird telling the user how to debug it when there is nothing to debug

JKL98ISR (approved):

PR #1285, opened 2022-04-18
Add option to pass FI in tabular to single run

matanper:


JKL98ISR:
can you maybe add a small test for error+passing it?

JKL98ISR (approved):

PR #1288, opened 2022-04-18
New categorical drift method 1138

### LINKED ISSUE #1138: [BUG] [CV] drift plots unclear when there are many classes
**Describe the bug**
the plot makes it look like there is a bug in the classes recognition and the "other" column isn't clear

**To Reproduce**
run the label/prediction drift check on a dataset with many classes

**Expected behavior**
explain better what the other means or remove the other column so the columns will look at a normal height 

**Screenshots**
Screen Shot 2022-03-30 at 16 32 17

image

nirhutnik:
Closes #1138

### DIFF
--- a/deepchecks/tabular/base_checks.py
+++ b/deepchecks/tabular/base_checks.py
@@ -13,7 +13,7 @@ import abc
 from functools import wraps
 from typing import Union, Mapping, List, Any
 
-from deepchecks.tabular import deprecation_warnings  # pylint: disable=unused-import
+from deepchecks.tabular import deprecation_warnings  # pylint: disable=unused-import # noqa: F401
 from deepchecks.tabular.dataset import Dataset
 from deepchecks.tabular.context import Context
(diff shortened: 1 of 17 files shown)


nirhutnik:
What changed:

* Split max_num_categories for drift checks, so it will be different for the drift calculation and the display.
* Changed the display: From now on, doesn't show "other", but just shows the top X categories, by user definition: largest train categories, largest test categories, or largest difference.
* In any way, results are still ordered alphabetically. This may seem like a weird decision, as if we show the "top x largest train categories" we would want them from large to small; However, this looks bad as some categories have a natural order, such as mnist labels.
* Added footnote to graphs when display is cut, that explains that not all data is shown
* Added deprecation warnings mechanism and deprecated max_num_categories

nirhutnik:
This is how the new footnote looks like:
image

matanper (approved):

shir22 (approved):

PR #1298, opened 2022-04-21
Adding examples validation

ItayGabbay:
#### Reference Issues/PRs

N/A

Add validations to the example path and structure when building the docs.

#### Any other comments?

---


.github/scripts/examples_validation.py
```diff
@@ -0,0 +1,44 @@
...
+
+checks_dirs = ["deepchecks/tabular/checks", "deepchecks/vision/checks"]
+
+ignored_files = [
+    "deepchecks/vision/checks/distribution/abstract_property_outliers.py",
+    ]
+
+
+def validate_dir(checks_path, examples_path):
+    for root, subdirs, files in os.walk(checks_path):
```
matanper, line 27: Not sure this is the best way, why not something like
```
from deepchecks.tabular import checks as tabular_checks

tabular_checks.__dict__
```

.github/workflows/docs.yaml
```diff
@@ -45,7 +45,8 @@ jobs:
           sudo apt-get install pandoc dvisvgm texlive texlive-latex-extra
       - name: Build documentation
         run: make docs
-
+      - name: Validate examples
+        run: python3 ./.github/scripts/examples_validation.py
```
matanper, line 49: why is it under .github and not tests?

.github/scripts/examples_validation.py
ItayGabbay: why? I don't care about the module. only about the file path.

.github/workflows/docs.yaml
ItayGabbay: Thought it better to have it there since it's an action logic. but I'll move it

.github/scripts/examples_validation.py
matanper: but you start having ignore_files and stuff instead of just importing the checks you want to validate

matanper (approved):

shir22 (approved):

PR #1315, opened 2022-04-24
Fi pipeline to always use permutation 

### LINKED ISSUE #1268: [BUG][tabular] Feature Importance does not consider preprocessing in pipeline

JKL98ISR:
close #1268

### DIFF
--- a/deepchecks/utils/features.py
+++ b/deepchecks/utils/features.py
@@ -163,7 +163,7 @@ def calculate_feature_importance(
                 permutation_failure = f'{e.message}\n using model\'s built-in feature importance instead'
 
     # If there was no force permutation, or it failed tries to take importance from the model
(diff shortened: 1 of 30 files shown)


deepchecks/utils/features.py
```diff
@@ -163,20 +163,24 @@ def calculate_feature_importance(
                 permutation_failure = f'{e.message}\n using model\'s built-in feature importance instead'
 
     # If there was no force permutation, or it failed tries to take importance from the model
-    if importance is None:
+    # We don't use a built importance in pipelines because the order might have changed
```
matanper, line 166:
```suggestion
    # We don't take built-in importance in pipelines because the pipeline is changing the features (for example one-hot encoding) which leads to the inner model features being different than the original dataset features
```

matanper (approved):
Just change the comment which is unclear

PR #1324, opened 2022-04-26
Various single feature contribution fixes

noamzbr:
Several needed fixes to single/simple feature contribution checks:

1. Expose the random state as a check parameter, as it may have significant influence on the result (PPS trains a model)
2. Make this random state None be default to get more robust results (a single seed may cause the user to get stuck on some noise this check outputted once)
3. Explicitly set the dtype of the label for simple feature contribution. PPS assumes label type based on the dtype of the column, and for the vision package it's reasonable to assume that if there are no actual fractions, the label is always some sort of classification (including detection, segmentation etc). Thus in most cases we need to explicitly cast to 'object' to tell PPS this is a classification label. This changes most test values.

### DIFF
--- a/deepchecks/core/check_utils/single_feature_contribution_utils.py
+++ b/deepchecks/core/check_utils/single_feature_contribution_utils.py
@@ -21,7 +21,7 @@ import plotly.graph_objects as go
 
 def get_single_feature_contribution(train_df: pd.DataFrame, train_label_name: Optional[Hashable], test_df: pd.DataFrame,
                                     test_label_name: Optional[Hashable], ppscore_params: dict, n_show_top: int,
-                                    random_state: int = 42):
+                                    random_state: int = None):
     """
     Calculate the PPS for train, test and difference for single feature contribution checks.
 
@@ -43,8 +43,8 @@ def get_single_feature_contribution(train_df: pd.DataFrame, train_label_name: Op
             dictionary of additional parameters for the ppscore predictor function
         n_show_top: int
             Number of features to show, sorted by the magnitude of difference in PPS
-        random_state: int, default 42
-        Random state for the ppscore.predictors function
+        random_state: int, default None
+            Random state for the ppscore.predictors function
 
     Returns:
         CheckResult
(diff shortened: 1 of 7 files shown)


deepchecks/core/check_utils/single_feature_contribution_utils.py
```diff
@@ -42,18 +43,20 @@ def get_single_feature_contribution(train_df: pd.DataFrame, train_label_name: Op
             dictionary of additional parameters for the ppscore predictor function
         n_show_top: int
             Number of features to show, sorted by the magnitude of difference in PPS
+        random_state: int, default 42
+        Random state for the ppscore.predictors function
```
matanper, line 47: missing tab

deepchecks/vision/checks/methodology/simple_feature_contribution.py
```diff
@@ -145,6 +150,15 @@ def compute(self, context: Context) -> CheckResult:
         df_train = pd.DataFrame(self._train_properties)
         df_test = pd.DataFrame(self._test_properties)
 
+        # PPS task type is inferred from label dtype. For computer vision tasks, it's safe to assume that unless
```
matanper, line 153: don't we have task type explicitly instead of checking column type?

nirhutnik: As said in Slack, I have an issue with this assumption. Age prediction for example is a regression problem with and int label. 
Joining Matan's question

tests/vision/checks/methodology/simple_feature_contribution_test.py
```diff
@@ -57,18 +58,35 @@ def ret_func(batch):
...
 
     # Act
     result = check.run(train, test)
 
     # Assert
     assert_that(result.value, has_entries({
-        'train': has_entries({'Brightness': equal_to(0)}),
-        'test': has_entries({'Brightness': equal_to(0)}),
+        'train': has_entries({'Brightness': close_to(0.08, 0.005)}),
+        'test': has_entries({'Brightness': close_to(0.08, 0.005)}),
```
nirhutnik, line 89: Any idea why this changed?

noamzbr: Previously this was treated as regression. For classification the results are different.

nirhutnik (approved):

matanper (approved):

PR #1333, opened 2022-04-26
Sklearn 0.23.2 support

### LINKED ISSUE #1311: [FEAT] Support sklearn 0.23.2
Currently we only support sklearn versions >= 1.0. We need to try to support the 0.23.2 as it is widely used by many environments and frameworks.

JKL98ISR:
closes #1311 

it supports it but some tests will fail because of things with do in the conftest.
but generally everything will work


matanper (approved):

ItayGabbay:
why not update the requirements? @JKL98ISR

ItayGabbay (approved):

PR #1334, opened 2022-04-26
adds detach for tensor before using .numpy()

### LINKED ISSUE #1332: [BUG] [CV] missing .detach() before using .numpy()
**Describe the bug**
performance check are failing because of missing detach

**To Reproduce**
Run performance check with a model that does something with a gradient on the tensor in some layer 

**Expected behavior**
checks not failing

JKL98ISR:
closes #1332


matanper:
Do we need some global way to handle this or only "on the fly"?

ItayGabbay (approved):

PR #1352, opened 2022-04-28
Simple model global metric

### LINKED ISSUE #1300: [FEAT][CV] Simple model comparison needs to be evaluated on a global metric
Currently the simple model always gets perfect recall, which results in the condition always triggering.

Like in the tabular case, the metric used to compare to the simple model should be a global metric (not per class), such as accuracy.

ItayGabbay:
#### Reference Issues/PRs

Closes #1300 

#### Any other comments?

---


matanper (approved):
It looks fine, but I think we need some better support to avg metrics, like in tabular
`get_scorers_list(context.train, self.alternative_metrics, avg=True)`

shir22 (approved):

PR #1355, opened 2022-04-29
[issue-1248] features inference for passed into check dataframes

### LINKED ISSUE #1248: [FEAT] auto infer features when dataframe is passed to check + warning
Currently when a df is passed to a check all of its columns are treated as non features, this behaviour is different than the behaviour when creating a "plain" dataset from a data frame.

when a df is passed to a check will create a dataset from it in the same default manner as when creating a "plain" dataset. As discussed this behavior should include auto infer of features.
In addition should add a warning when a df is passed, explaining that categorical features will be inferred and explaining / linking to how to define a dataset.

yromanyshyn:
resolves #1248 

@Nadav-Barak 
it is enough to modify line 889 in `Dataset.cast_to_dataset` to solve that issue, but I am not sure how clear the warning message raised by the features inference logic (`Dataset._infer_categorical_features` line 573) will be for the users in this case. What do you think?

### DIFF
--- a/tests/base/dataset_test.py
+++ b/tests/base/dataset_test.py
@@ -16,8 +16,8 @@ import numpy as np
 import pandas as pd
 import pytest
 from hamcrest import (all_of, assert_that, calling, contains_exactly, equal_to,
-                      has_item, has_length, has_property, instance_of, is_,
-                      not_none, raises, greater_than)
+                      greater_than, has_item, has_length, has_property,
+                      instance_of, is_, not_none, raises)
 from sklearn.datasets import load_iris, make_classification
 
 from deepchecks.core.errors import DeepchecksValueError
--- a/tests/base/dataset_test.py
+++ b/tests/base/dataset_test.py
@@ -842,7 +842,7 @@ def test__ensure_not_empty_dataset__with_dataframe(iris: pd.DataFrame):
     ds = Dataset.cast_to_dataset(iris)
     # Assert
     assert_that(ds, instance_of(Dataset))
-    assert_that(ds.features, has_length(0))
+    # assert_that(ds.features, has_length(0))
     assert_that(ds.label_name, equal_to(None))
     assert_that(ds.n_samples, equal_to(len(iris)))
 
--- a/tests/base/dataset_test.py
+++ b/tests/base/dataset_test.py
@@ -17,7 +17,7 @@ import pandas as pd
 import pytest
 from hamcrest import (all_of, assert_that, calling, contains_exactly, equal_to,
                       has_item, has_length, has_property, instance_of, is_,
-                      not_none, raises)
+                      not_none, raises, greater_than)
 from sklearn.datasets import load_iris, make_classification
 
 from deepchecks.core.errors import DeepchecksValueError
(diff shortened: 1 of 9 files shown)


matanper:
I think that in Dataset line 889 we can add the warning "DataFrame was passed, initializing deepchecks.tabular.Dataset using the given dataframe"

Nadav-Barak:
I think the warning massage is good for both cases (if the user passed a DF directly or created a DS without specifying categorical)

tests/base/dataset_test.py
```diff
@@ -842,7 +842,7 @@ def test__ensure_not_empty_dataset__with_dataframe(iris: pd.DataFrame):
     ds = Dataset.cast_to_dataset(iris)
     # Assert
     assert_that(ds, instance_of(Dataset))
-    assert_that(ds.features, has_length(0))
+    # assert_that(ds.features, has_length(0))
```
Nadav-Barak, line 845: maybe change to assert that num features is greater then 0

tests/checks/overview/columns_info_test.py
```diff
@@ -48,9 +48,17 @@ def test_columns_info():
...
+        'date': 'numerical feature', 
+        'a': 'categorical feature', 
+        'b': 'numerical feature', 
+        'c': 'numerical feature', 
+        'label': 'categorical feature'
+    }
     # in df all columns are other
-    expected_res_df = {'index': 'other', 'date': 'other', 'a': 'other',
-                       'b': 'other', 'c': 'other', 'label': 'other'}
+    # expected_res_df = {'index': 'other', 'date': 'other', 'a': 'other',
```
Nadav-Barak, line 60: Think this commented section can be deleted

Nadav-Barak (approved):

tests/checks/overview/columns_info_test.py
ItayGabbay: Agree

tests/checks/integrity/string_mismatch_comparison_test.py
```diff
@@ -82,16 +82,16 @@ def test_no_mismatch_on_numeric_string_column():
...
 
-    # Act
-    result = StringMismatchComparison().run(pd.DataFrame(data=data), pd.DataFrame(data=compared_data)).value
+#     # Act
+#     result = StringMismatchComparison().run(pd.DataFrame(data=data), pd.DataFrame(data=compared_data)).value
 
-    # Assert
-    assert_that(result, has_length(0))
+#     # Assert
+#     assert_that(result, has_length(0))
```
ItayGabbay, line 95: Let's delete commented code

ItayGabbay (approved):

tests/checks/integrity/string_mismatch_comparison_test.py
yromanyshyn: removed

tests/checks/overview/columns_info_test.py
yromanyshyn: removed

tests/base/dataset_test.py
yromanyshyn: fixed

yromanyshyn:
@matanper 

> I think that in Dataset line 889 we can add the warning "DataFrame was passed, initializing deepchecks.tabular.Dataset using the given dataframe"

Think it will be redundant, this function is part of a public API and it could be strange for user to see warning as this when he/she will try to use this function directly

matanper:
> @matanper
> 
> > I think that in Dataset line 889 we can add the warning "DataFrame was passed, initializing deepchecks.tabular.Dataset using the given dataframe"
> 
> Think it will be redundant, this function is part of a public API and it could be strange for user to see warning as this when he/she will try to use this function directly

I don't see a reason for it to be a public API, it doesn't help the user more than calling the ctor

yromanyshyn:
@matanper 

> I don't see a reason for it to be a public API, it doesn't help the user more than calling the ctor

ok, I have not changed its visibility marker, but I have added a warning

PR #1356, opened 2022-04-29
[issue-892] modified VisionData ctor

### LINKED ISSUE #892: [FEAT][CV] Enable to create VisionData object from dataset, not only dataloader
It's possible to create a VisionData object from a dataset, and build the dataloader internally (assuming it's default)

yromanyshyn:
resolves #892


matanper (approved):

tests/vision/base/test_vision_data.py
```diff
@@ -416,3 +416,9 @@ def infer_on_batch(self, batch, model, device):
                 raises(ValidationError,
                        'Check requires detection predictions to be a list of 2D tensors, when '
                        'each row has 6 columns: \[x, y, width, height, class_probability, class_id\]'))
+
+
+def test_vision_data_initialization_from_dataset_instance(coco_train_dataloader: DataLoader):
```
noamzbr, line 421: I think this dataset won't actually work if you try to get a sample from it, because the default collate_fn is to create a torch Tensor and you can't create a tensor from a list of tensors of un-equal sizes.
This stresses the fact that there must be an option IMO to pass collate_fn.
I see three alternatives:

1. Keep as is, and if a user sees an error he'll need to understand he needs to create his own dataloader
2. add collate_fn arg to VisionData
3. Do this with a seperate from_dataset() class method that accepts all the args dataloader accepted (perhaps in kwargs)

@yromanyshyn @matanper What do you think?

matanper: 2 is too confusing IMO with a dataloader. 3 sounds a better option IMO

yromanyshyn: added `from_dataset()` constructor

PR #1357, opened 2022-05-01
Torch transformers support

### LINKED ISSUE #1303: [FEAT] Support torch transforms in the RobustnessReport
All the relevant code is under `deepchecks.vision.utils.transformations`

- [ ] We have an implementation of 2 classes that handles the transformations for albumentations and imgaug. Need to add a new class to handle torch transforms
- [ ] Add to `get_transforms_handler` the new class 
- [ ] Add test for robustness report with Dataset that works with torch transformations

JKL98ISR:
closes #1303


deepchecks/vision/datasets/classification/mnist.py
```diff
@@ -44,7 +45,8 @@ def load_dataset(
     batch_size: t.Optional[int] = None,
     shuffle: bool = True,
     pin_memory: bool = True,
-    object_type: Literal['VisionData', 'DataLoader'] = 'DataLoader'
+    object_type: Literal['VisionData', 'DataLoader'] = 'DataLoader',
+    is_torch_transform: bool = False,
```
matanper, line 49: don't do this like this. Look how we did it in COCO

deepchecks/vision/utils/transformations.py
```diff
@@ -118,24 +144,50 @@ def get_robustness_augmentations(cls, data_dim: Optional[int] = 3):
...
+class TorchTransformations(AlbumentationsTransformations):
+    """Class containing supporting functions for torch transforms."""
+
+    @classmethod
+    def add_augmentation_in_start(cls, aug, transforms: T.Compose):
+        """Add given transformations to the start of given transforms object."""
+        if isinstance(aug, (albumentations.Compose, albumentations.BasicTransform)):
+            alb_aug = aug
+
+            class TorchWrapper:
```
matanper, line 156: how this handle transformations on the label?

noamzbr: torch.transforms does not transform labels to the best of my knowledge, so makes sense we won't have this option when using an albumentation aug inside a torch pipeline...

deepchecks/vision/utils/transformations.py
```diff
@@ -118,24 +144,50 @@ def get_robustness_augmentations(cls, data_dim: Optional[int] = 3):
         return augmentations
 
 
-def get_transforms_handler(transforms):
+class TorchTransformations(AlbumentationsTransformations):
```
noamzbr, line 147: TorchTransformations should have a different get_robustness_augmentations than AlbumentationsTransformations. 
Ideally, if it's classification we want the same things we had in AlbumentationsTransformations, and if it's any other task we want them excluding ShiftScaleRotate. Can we do that?

noamzbr: Also, why not use native torch augmentations for TorchTransformations?

tests/vision/checks/performance/robustness_report_test.py
```diff
@@ -48,6 +49,49 @@ def test_mnist(mnist_dataset_train, mock_trained_mnist, device):
     }))
 
 
+def test_mnist_torch(mnist_dataset_train_torch, mock_trained_mnist, device):
```
noamzbr, line 52: We should also have a test for some RGB dataset.
For all I care, can create a fixture that has an version of MNIST that is RGB instead of greyscale.

matanper: We can just test on COCO

noamzbr (approved):

deepchecks/vision/utils/transformations.py
```diff
@@ -118,24 +147,59 @@ def get_robustness_augmentations(cls, data_dim: Optional[int] = 3):
...
+                        return T.ToTensor()(alb_aug(image=image)['image'])
+                    return alb_aug(image=image)['image']
+            aug = TorchWrapper()
+        elif not isinstance(aug, torch.nn.Module):
+            raise DeepchecksValueError(f'Transforms is of type torch, can\'t add to it type '
+                                       f'{type(aug).__name__}')
+        return T.Compose([aug, *transforms.transforms])
+
+
+class TorchTransformations(TorchTransformationsClassification):
```
matanper, line 174: 1. Instead of having a different class maybe add task_type to `get_robustness_augmentations`
2. If you really want 2 classes, makes more sense that `TorchTransformationsClassification` will inherit from `TorchTransformations` and not the opposite

deepchecks/vision/utils/transformations.py
```diff
@@ -9,21 +9,50 @@
...
-__all__ = ['get_transforms_handler', 'add_augmentation_in_start', 'un_normalize_batch',
+__all__ = ['get_transforms_handler', 'un_normalize_batch', 'AbstractTransformations',
            'ImgaugTransformations', 'AlbumentationsTransformations']
 
 
-class ImgaugTransformations:
+class AbstractTransformations(abc.ABC):
+    """Abstract class for supporting functions for various transforms."""
+
+    is_image_shifting = True
```
matanper, line 34: @noamzbr is "shifting" the only transformation that affects labels? I wonder if this should be called "is_transforming_labels" or something

noamzbr: yeah agree is_transforming_labels is better

deepchecks/vision/utils/transformations.py
JKL98ISR: I want it to inherit from classification because then I can filter the transforms

matanper: so inherit both from Albumentations. this is too confusing IMO it's like `VisionData` will inherit from `DetectionData`

JKL98ISR: but both using same add_aug_at_start function

JKL98ISR: and I don't want to add task type as parameter, it make the usage more wonky and I need it in the class

deepchecks/vision/utils/transformations.py
```diff
@@ -118,24 +147,71 @@ def get_robustness_augmentations(cls, data_dim: Optional[int] = 3):
...
+                def __call__(self, image):
+                    if isinstance(image, Image.Image):
+                        return T.ToPILImage()(alb_aug(image=np.array(image))['image'])
+                    elif isinstance(image, torch.Tensor):
+                        image = image.cpu().detach().numpy()
+                        return T.ToTensor()(alb_aug(image=image)['image'])
+                    return alb_aug(image=image)['image']
+            aug = TorchWrapper()
+        elif not isinstance(aug, torch.nn.Module):
+            raise DeepchecksValueError(f'Transforms is of type torch, can\'t add to it type '
```
matanper, line 169: this is a bit weird here, no? can't add torch transform inside TorchTransformations

matanper (approved):

deepchecks/vision/utils/transformations.py
JKL98ISR: it's if not

PR #1368, opened 2022-05-02
Added margin_quantile_filter to EMD drift score

nirhutnik:
Added margin_quantile_filter to EMD and the checks that use it.
Added tests for EMD.
Right now changed the default setting of margin_quantile_filter to 0 (values do not change).


deepchecks/utils/distribution/drift.py
```diff
@@ -69,27 +70,39 @@ def earth_movers_distance(dist1: Union[np.ndarray, pd.Series], dist2: Union[np.n
...
     """
-    unique1 = np.unique(dist1)
-    unique2 = np.unique(dist2)
+    if not isinstance(margin_quantile_filter, Number) or margin_quantile_filter < 0 or margin_quantile_filter >= 0.5:
+        raise DeepchecksValueError(
+            f'margin_quantile_filter expected a value in range [0, 0.5), instead got {margin_quantile_filter}')
 
-    sample_space = list(set(unique1).union(set(unique2)))
+    if margin_quantile_filter != 0:
+        dist1_qt_min = np.quantile(dist1, margin_quantile_filter)
```
noamzbr, line 92: Think you can make it two lines by `np.quantile(dist1, [margin_quantile_filter, 1-margin_quantile_filter])`. Would probably be more efficient.

noamzbr (commented):
1. Think the default value should be something like 0.02
2. Need to remove the FI docs committed by mistake (:

deepchecks/utils/distribution/drift.py
nirhutnik: Thanks!

noamzbr (approved):

matanper (approved):

PR #1369, opened 2022-05-02
Nb/vision/new label

Nadav-Barak:
Check for labels in test set that do not appear in the train set

### DIFF
--- a/deepchecks/vision/checks/distribution/new_labels.py
+++ b/deepchecks/vision/checks/distribution/new_labels.py
@@ -15,13 +15,13 @@ from typing import Dict
 
 import torch
 
-from deepchecks.core import ConditionResult
-from deepchecks.core import DatasetKind, CheckResult
+from deepchecks.core import CheckResult, ConditionResult, DatasetKind
 from deepchecks.core.condition import ConditionCategory
 from deepchecks.core.errors import DeepchecksValueError
-from deepchecks.utils.strings import format_percent, format_number
-from deepchecks.vision import Context, TrainTestCheck, Batch, VisionData
-from deepchecks.vision.utils.image_functions import prepare_thumbnail, draw_bboxes
+from deepchecks.utils.strings import format_number, format_percent
+from deepchecks.vision import Batch, Context, TrainTestCheck, VisionData
+from deepchecks.vision.utils.image_functions import (draw_bboxes,
+                                                     prepare_thumbnail)
 
 __all__ = ['NewLabels']
 
--- a/deepchecks/vision/checks/distribution/new_labels.py
+++ b/deepchecks/vision/checks/distribution/new_labels.py
@@ -1,0 +1,204 @@
+# ----------------------------------------------------------------------------
+# Copyright (C) 2021-2022 Deepchecks (https://www.deepchecks.com)
+#
+# This file is part of Deepchecks.
+# Deepchecks is distributed under the terms of the GNU Affero General
+# Public License (version 3 or later).
+# You should have received a copy of the GNU Affero General Public License
+# along with Deepchecks.  If not, see <http://www.gnu.org/licenses/>.
+# ----------------------------------------------------------------------------
+#
+"""Module contains Train Test label Drift check."""
+import string
+from collections import defaultdict
+from secrets import choice
+from typing import Dict
+
+import torch
+
+from deepchecks.core import ConditionResult
+from deepchecks.core import DatasetKind, CheckResult
+from deepchecks.core.condition import ConditionCategory
+from deepchecks.core.errors import DeepchecksValueError
+from deepchecks.utils.strings import format_percent
+from deepchecks.vision import Context, TrainTestCheck, Batch, VisionData
+from deepchecks.vision.utils.image_functions import prepare_thumbnail, draw_bboxes
+
+__all__ = ['NewLabels']
+
+from deepchecks.vision.vision_data import TaskType
+
+THUMBNAIL_SIZE = (200, 200)
+
+
+def draw_image(data: VisionData, sample_index: int, class_id: int) -> str:
+    """Return an image to show as output of the display.
+
+    Parameters
+    ----------
+    data : VisionData
+        The vision data object used in the check.
+    sample_index : int
+        The batch index of the sample to draw the image for.
+    class_id : int
+        The class_id of the image or relevant bounding box inside image.
+    """
+
+    sample = data.data_loader.dataset[sample_index]
+    batch = data.to_batch(sample)
+
+    image = data.batch_to_images(batch)[0]
+    if data.task_type == TaskType.OBJECT_DETECTION:
+        formatted_labels_of_image = data.batch_to_labels(batch)[0].numpy()
+        bboxes_of_class_id = torch.tensor([x for x in formatted_labels_of_image if x[0] == class_id])
+        image = draw_bboxes(image, bboxes_of_class_id, copy_image=False, border_width=5)
+
+    image_thumbnail = prepare_thumbnail(
+        image=image,
+        size=THUMBNAIL_SIZE,
+        copy_image=False
+    )
+    return image_thumbnail
+
+
+class NewLabels(TrainTestCheck):
+    """Finds labels in the test set that do not exist in the train set.
+
+    Parameters
+    ----------
+    max_images_to_display_per_label : int , default: 3
+        maximum number of images to show from each newly found label in the test set.
+    max_new_labels_to_display : int , default: 3
+        Maximum number of new labels to display in output.
+    """
+
+    def __init__(
+            self,
+            max_images_to_display_per_label: int = 3,
+            max_new_labels_to_display: int = 3,
+            **kwargs
+    ):
+        super().__init__(**kwargs)
+        # validate input parameters:
+        if not isinstance(max_images_to_display_per_label, int):
+            raise DeepchecksValueError('max_num_images_to_display_per_label must be an integer')
+        if not isinstance(max_new_labels_to_display, int):
+            raise DeepchecksValueError('max_num_new_labels_to_display must be an integer')
+
+        self.max_images_to_display_per_label = max_images_to_display_per_label
+        self.max_new_labels_to_display = max_new_labels_to_display
+        self._class_id_counter = defaultdict(list)
+
+    def update(self, context: Context, batch: Batch, dataset_kind):
+        """Counts number of appearances for each class_id in train and test."""
+        data = context.get_data_by_kind(dataset_kind)
+        classes_in_batch = data.get_classes(batch.labels)
+        for image_idx, labels_per_image in enumerate(classes_in_batch):
+            for label in labels_per_image:
+                if label not in self._class_id_counter:
+                    self._class_id_counter[label] = [0, 0]
+                self._class_id_counter[label][dataset_kind == DatasetKind.TEST] += 1
+
+    def compute(self, context: Context) -> CheckResult:
+        """Calculate which class_id are only available in the test data set and display them.
+
+        Returns
+        -------
+        CheckResult
+            value: A dictionary showing new class_ids introduced in the test set and number of times they were spotted.
+            display: Images containing said class_ids from the test set.
+        """
+        data = context.get_data_by_kind(DatasetKind.TEST)
+        class_id_only_in_test = defaultdict(list)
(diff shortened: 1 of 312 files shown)


tests/vision/checks/distribution/new_labels_test.py
```diff
@@ -0,0 +1,151 @@
...
+        if idx % 7 == 0:
+            data, label = orig_dataset[idx]
+            return data, new_labels[idx % 5]
+        else:
+            return orig_dataset[idx]
+
+    return add_new_test_labels
+
+
+def get_modification_func_classification_ver2():
```
matanper, line 36: give more informative name than ver2, what's the difference between them?

docs/source/checks/vision/distribution/plot_new_labels.py
```diff
@@ -0,0 +1,33 @@
+# -*- coding: utf-8 -*-
+"""
+New Labels
```
matanper, line 3: Notebook needs a bit more work

docs/source/checks/vision/distribution/plot_new_labels.py
```diff
@@ -0,0 +1,33 @@
...
+* `Run the check <#run-the-check>`__
+
+How the check works
+========================
+In this check we count the frequency of each class id in the test set then check which of them
+do not apper in the training set.
+"""
+
+# %%
+# Imports
```
matanper, line 20: No reason to separate imports from the run IMO

deepchecks/vision/checks/distribution/new_labels.py
```diff
@@ -0,0 +1,204 @@
...
+
+            html = HTML_TEMPLATE.format(
+                label_name=data.label_id_to_name(class_id),
+                images=images_combine,
+                count=format_number(num_occurrences),
+                id=sid
+            )
+            display.append(html)
+        class_id_only_in_test['all_labels'] = sum(data.n_of_samples_per_class.values())
+        return CheckResult(class_id_only_in_test, display=''.join(display[:self.max_new_labels_to_display]))
```
matanper, line 134: instead of `display[:self.max_new_labels_to_display]`
add inside the `for` a counter with break 
```
for index, (class_id, num_occurrences) in class_id_only_in_test.items():
...
  if index == self.max_new_labels_to_display:
     break
```

deepchecks/vision/checks/distribution/new_labels.py
```diff
@@ -0,0 +1,204 @@
...
+from deepchecks.vision.utils.image_functions import prepare_thumbnail, draw_bboxes
+
+__all__ = ['NewLabels']
+
+from deepchecks.vision.vision_data import TaskType
+
+THUMBNAIL_SIZE = (200, 200)
+
+
+def draw_image(data: VisionData, sample_index: int, class_id: int) -> str:
```
noamzbr, line 34: That looks pretty generic to me. @matanper @yromanyshyn any place or way we can make stuff like this a common util?

deepchecks/vision/checks/distribution/new_labels.py
```diff
@@ -0,0 +1,204 @@
...
+        CheckResult
+            value: A dictionary showing new class_ids introduced in the test set and number of times they were spotted.
+            display: Images containing said class_ids from the test set.
+        """
+        data = context.get_data_by_kind(DatasetKind.TEST)
+        class_id_only_in_test = defaultdict(list)
+        for class_id, value in self._class_id_counter.items():
+            if value[0] == 0:
+                class_id_only_in_test[class_id] = value[1]
+        class_id_only_in_test = dict(sorted(class_id_only_in_test.items(), key=lambda item: -item[1]))
```
noamzbr, line 115: Add a comment here explaining that your sorting by number of occurrences in test in descending order

deepchecks/vision/checks/distribution/new_labels.py
```diff
@@ -0,0 +1,204 @@
...
+    def compute(self, context: Context) -> CheckResult:
+        """Calculate which class_id are only available in the test data set and display them.
+
+        Returns
+        -------
+        CheckResult
+            value: A dictionary showing new class_ids introduced in the test set and number of times they were spotted.
+            display: Images containing said class_ids from the test set.
+        """
+        data = context.get_data_by_kind(DatasetKind.TEST)
```
matanper, line 110: call it test_data

deepchecks/vision/checks/distribution/new_labels.py
noamzbr, line 134: display can also be a list of display objects, can do:
```suggestion
        return CheckResult(class_id_only_in_test, display=display[:self.max_new_labels_to_display])
```

deepchecks/vision/checks/distribution/new_labels.py
```diff
@@ -0,0 +1,204 @@
...
+
+    def update(self, context: Context, batch: Batch, dataset_kind):
+        """Count number of appearances for each class_id in train and test."""
+        data = context.get_data_by_kind(dataset_kind)
+        classes_in_batch = data.get_classes(batch.labels)
+        for labels_per_image in classes_in_batch:
+            for label in labels_per_image:
+                if label not in self._class_id_counter:
+                    self._class_id_counter[label] = [0, 0]
+                self._class_id_counter[label][dataset_kind == DatasetKind.TEST] += 1
```
matanper, line 99: you need to initialize this var in `initalize run`

also we said you don't need this if you use `classes_indices` right?

deepchecks/vision/checks/distribution/new_labels.py
```diff
@@ -0,0 +1,204 @@
...
+        data = context.get_data_by_kind(DatasetKind.TEST)
+        class_id_only_in_test = defaultdict(list)
+        for class_id, value in self._class_id_counter.items():
+            if value[0] == 0:
+                class_id_only_in_test[class_id] = value[1]
+        class_id_only_in_test = dict(sorted(class_id_only_in_test.items(), key=lambda item: -item[1]))
+
+        # Create display
+        display = []
+        for class_id, num_occurrences in class_id_only_in_test.items():
```
noamzbr, line 119: We're running this loop for all `class_id_only_in_test` but are displaying only `self.max_new_labels_to_display` of them, why not stop the loop after `self.max_new_labels_to_display` rounds?

deepchecks/vision/checks/distribution/new_labels.py
```diff
@@ -0,0 +1,204 @@
...
+                label_name=data.label_id_to_name(class_id),
+                images=images_combine,
+                count=format_number(num_occurrences),
+                id=sid
+            )
+            display.append(html)
+        class_id_only_in_test['all_labels'] = sum(data.n_of_samples_per_class.values())
+        return CheckResult(class_id_only_in_test, display=''.join(display[:self.max_new_labels_to_display]))
+
+    def add_condition_new_label_percentage_not_greater_than(self, max_allowed_new_labels: float = 0.005):
```
noamzbr, line 136:
```suggestion
    def add_condition_new_label_ratui_not_greater_than(self, max_allowed_new_labels_ratio: float = 0.005):
```

deepchecks/vision/checks/distribution/new_labels.py
```diff
@@ -0,0 +1,204 @@
...
+        # Default value set to 0.005 because of sampling mechanism
+        """
+        Add condition - Percentage of labels that apper only in the test set required to be below specified threshold.
```
noamzbr, line 139:
```suggestion
        Add condition - Ratio of labels that appear only in the test set required to be below specified threshold.
```

deepchecks/vision/checks/distribution/new_labels.py
```diff
@@ -0,0 +1,204 @@
...
+            the max threshold for percentage of labels that only apper in the test set.
+        """
+
+        def condition(result: Dict) -> ConditionResult:
+            total_labels_in_test_set = result['all_labels']
+            new_labels_in_test_set = sum(result.values()) - total_labels_in_test_set
+            present_new_labels = new_labels_in_test_set / total_labels_in_test_set
+
+            if present_new_labels > max_allowed_new_labels:
+                massage = f'{format_percent(present_new_labels)} of labels found in test set were not in train set.'
```
noamzbr, line 153: Would be better if the fail message would list the class names of the top new labels.

deepchecks/vision/checks/distribution/new_labels.py
```diff
@@ -0,0 +1,204 @@
...
+
+    def add_condition_new_label_percentage_not_greater_than(self, max_allowed_new_labels: float = 0.005):
+        # Default value set to 0.005 because of sampling mechanism
+        """
+        Add condition - Percentage of labels that apper only in the test set required to be below specified threshold.
+
+        Parameters
+        ----------
+        max_allowed_new_labels: float , default: 0.005
+            the max threshold for percentage of labels that only apper in the test set.
```
noamzbr, line 144: Add note explaining why the default value should be that and not 0 (due to sampling)

docs/source/checks/vision/distribution/plot_new_labels.py
noamzbr, line 20: Don't think that imports deserves it's own section. Should be done in start of Run the Check section.

docs/source/checks/vision/distribution/plot_new_labels.py
```diff
@@ -0,0 +1,33 @@
...
+from deepchecks.vision.checks import NewLabels
+
+
+# %%
+# Run the Check
+# =============
+
+coco_train = coco.load_dataset(train=True, object_type='VisionData', shuffle=False)
+coco_test = coco.load_dataset(train=False, object_type='VisionData', shuffle=False)
+check = NewLabels().add_condition_new_label_percentage_not_greater_than(0.05)
```
noamzbr, line 32: Explain also here why you are using this threshold and not 0 (and why a threshold of 0 won't work well sometimes)

noamzbr (commented):
Nice work! Suggested inline that we may be able to use the draw_image function in a more general setting.

deepchecks/vision/checks/distribution/new_labels.py
Nadav-Barak: There was an additional variable I removed, 
This can also be calculated based on classes_indices now that you mention it

deepchecks/vision/checks/distribution/new_labels.py
Nadav-Barak: I think fail massage should be concise but I will add some info on top classes

deepchecks/vision/checks/distribution/new_labels.py
matanper: it's not generic enough, for example:
here filtering specific class_id for the bboxes.
in another place not showing bboxes at all,
in third place (outliers) filerting only single bbox by it's index in the list of bboxes

tests/vision/checks/distribution/new_labels_test.py
```diff
@@ -0,0 +1,147 @@
...
+# Copyright (C) 2021 Deepchecks (https://www.deepchecks.com)
+#
+# This file is part of Deepchecks.
+# Deepchecks is distributed under the terms of the GNU Affero General
+# Public License (version 3 or later).
+# You should have received a copy of the GNU Affero General Public License
+# along with Deepchecks.  If not, see <http://www.gnu.org/licenses/>.
+# ----------------------------------------------------------------------------
+#
+"""Image Property Drift check tests"""
```
nirhutnik, line 11: This is not what this is :)

tests/vision/checks/distribution/new_labels_test.py
```diff
@@ -0,0 +1,147 @@
...
+from deepchecks.vision.checks.distribution import NewLabels
+from deepchecks.vision.utils.test_utils import get_modified_dataloader
+from tests.checks.utils import equal_condition_result
+
+
+def get_modification_func_classification(new_labels):
+    def add_new_test_labels(orig_dataset, idx):
+        if idx % 7 == 0:
+            data, label = orig_dataset[idx]
+            return data, new_labels[idx % 5]
```
nirhutnik, line 29: why the mod here?

Nadav-Barak: Just a method to change a small fraction of the labels

ItayGabbay (approved):

PR #1372, opened 2022-05-02
[issue-811] TrainTestFeatureDrift - fi order bug

### LINKED ISSUE #811: [BUG] TrainTestFeatureDrift not showing features sorted by feature importance
**Describe the bug**
When running TrainTestFeatureDrift on breast cancer dataset from the module the results are different than expected.
the test writes that the features will be shown according to their FI in a sorted way but, the 5 features shown are not the top 5 FI.
one of them is the 10th, and they are not sorted

**To Reproduce**
1. load breast cancer train and test
2. transform to Dataset
3. load fitted model
4. run TrainTestFeatureDrift
**Expected behavior**
the features shown should be sorted by FI and should be the first 5 FI features

**Screenshots**
image
image

google colab
latest deepcheck version as of 3.2.22

yromanyshyn:
resolves #811

### DIFF
--- a/deepchecks/tabular/checks/distribution/train_test_feature_drift.py
+++ b/deepchecks/tabular/checks/distribution/train_test_feature_drift.py
@@ -139,6 +139,18 @@ class TrainTestFeatureDrift(TrainTestCheck):
 
         values_dict = OrderedDict()
         displays_dict = OrderedDict()
+
+        features_order = (
+            tuple(
+                features_importance
+                .sort_values(ascending=False)
+                .rank(method='first', ascending=False)
+                .index
+            )
+            if features_importance is not None
+            else None
+        )
+
         for column in train_dataset.features:
             if column in train_dataset.numerical_features:
                 column_type = 'numerical'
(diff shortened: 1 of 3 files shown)


noamzbr (commented):
Looks good, can we make sure we don't run into the same problem in run_whole_dataset_drift and other checks? Most of the others (integrity checks) use column_importance_sorter_df to sort by FI.

yromanyshyn:
@noamzbr 

> Looks good, can we make sure we don't run into the same problem in run_whole_dataset_drift and other checks? Most of the others (integrity checks) use column_importance_sorter_df to sort by FI.

hm ... do not know/ 

I do not think that it requires some additional solution.
It was just a mistake made by whoever wrote/modified this check. 
Instead of creating a separate variable with the list of sorted features, 
he sorted thems two times by two different methods.

What can save from this kind of mistake/

`run_whole_dataset_drift` is correct, because fi is sorted only once and then used whenever it is needed (`whole_dataset_drift_utils.run_whole_dataset_drift` line 79 and below)

matanper (approved):

PR #1375, opened 2022-05-03
Change PSI calculation + EMD default

nirhutnik:
PSI - (effectively preprocess_2_cat_cols...)
* Changed order of sorting (by difference instead of by train), affecting which categories will be used as is for PSI and which will be binned into "other"
* Added tests

EMD - 
* Changed default to 0.025 instead of 0 and fixed tests accordingly


deepchecks/tabular/checks/distribution/train_test_feature_drift.py
```diff
@@ -145,17 +145,6 @@ def run_logic(self, context: Context) -> CheckResult:
 
         values_dict = OrderedDict()
         displays_dict = OrderedDict()
-
```
JKL98ISR, line 148: fix overridden code

noamzbr (approved):
Approved, but as said make sure not to override the changes to FI sorting made by [#1372](https://github.com/deepchecks/deepchecks/pull/1372)

nirhutnik:
Recreated in #1390

PR #1402, opened 2022-05-09
From json

### LINKED ISSUE #1401: [FEAT] adding from_json for any class that has to_json
related to #1307,

adding easy conversion to check result or suite result with nearly full functionality and display

JKL98ISR:
closes #1401


deepchecks/core/check_json.py
```diff
@@ -0,0 +1,114 @@
...
+        Json data
+    """
+
+    def __init__(self, json_dict: Union[str, Dict]):
+        if isinstance(json_dict, str):
+            json_dict = jsonpickle.loads(json_dict)
+
+        self.value = json_dict.get('value')
+        self.header = json_dict.get('header')
+        self.check = FakeCheck(json_dict.get('check'))
```
yromanyshyn, line 61: why we need this `FakeCheck` type?
could not we recreate an original check?
we store a check type name and its parameters, I think that it is enough info to recreate an original check type instance/

deepchecks/core/check_result.py
```diff
@@ -58,7 +56,49 @@
 TDisplayItem = Union[str, pd.DataFrame, Styler, BaseFigure, TDisplayCallable]
 
 
-class CheckResult:
+class CheckOutput:
```
yromanyshyn, line 59: maybe let's rename it to the `BaseCheckResult`?

deepchecks/core/check_json.py
JKL98ISR: we can't create a check class from that and we can't use those functions

deepchecks/core/check_json.py
```diff
@@ -0,0 +1,114 @@
...
+        self._metadata = metadata
+
+    def metadata(self, *args, **kwargs):  # pylint: disable=unused-argument
+        return self._metadata
+
+    def name(self):
+        return self._metadata['name']
+
+
+class CheckResultJson(CheckResult):
```
matanper, line 41: Why do we need `CheckResultJson` if we have the `FakeCheck`?
can we just initialize `CheckResult` with `FakeCheck` instance?

JKL98ISR: we still have some different functions and I prefer having an init instead of seperate util functions

deepchecks/core/serialization/check_result/json.py
```diff
@@ -129,9 +127,9 @@ def handle_dataframe(
     def handle_callable(cls, item: t.Callable, index: int, **kwargs) -> t.Dict[str, t.Any]:
         """Handle callable."""
         return {
-            'type': 'images',
+            'type': 'html',
             'payload': [
-                base64.b64encode(buffer.read()).decode('ascii')
+                imagetag(base64.b64encode(buffer.read()))
```
yromanyshyn, line 132: why? what if the user wants to process those images in a different environment, with help of other tools and not only to display thems in the browser? he would need to parse that image tag then/

deepchecks/core/check_result.py
JKL98ISR: sure

deepchecks/core/check_json.py
matanper: why? that we discussed about, having general `deepchecks.from_json`

matanper: and I don't see any different functions except the pass in process_conditions

JKL98ISR: @matanper it is still a different function, and there is a global from json

yromanyshyn (approved):

ItayGabbay (approved):

PR #1412, opened 2022-05-10
Airflow integration

### LINKED ISSUE #1411: [FEAT] Airflow Integration

ItayGabbay:
#### Reference Issues/PRs

Closes #1411 

Adding an integration with apache airflow.

image
image
image

---


matanper:
I think the DAG would be more realistic if it had simple model training as a step before the checks. but if it's too complex this is also fine

docs/source/user-guide/integrations/airflow.rst
```diff
@@ -0,0 +1,172 @@
...
+
+    from airflow import DAG
+    from airflow.operators.python import PythonOperator
+    import joblib
+    import pandas as pd
+
+    from deepchecks.tabular.datasets.classification import adult
+
+    dir_path = "suite_results"
+    data_path = os.path.join(os.getcwd(), "data")  # For demo only. Replace that with a S3/GCS other than local filesystem
```
shir22, line 33: split the comment here to two lines so it will be viewable without scrolling

docs/source/user-guide/integrations/airflow.rst
```diff
@@ -0,0 +1,172 @@
...
+    def load_adult_model(**context):
+        from deepchecks.tabular.datasets.classification.adult import load_fitted_model
+
+        model = load_fitted_model()
+        with open(os.path.join(data_path, "adult_model.joblib"), "wb") as f:
+            joblib.dump(model, f)
+
+        context["ti"].xcom_push(key="adult_model", value=os.path.join(data_path, "adult_model.joblib"))
+
+.. warning::
```
shir22, line 64:
```suggestion
.. note::
```

shir22: (as the code is not intended to be used as-is, warning feels a bit "too serious" given the fact that it's just a mock anyways)

docs/source/user-guide/integrations/airflow.rst
```diff
@@ -0,0 +1,172 @@
...
+    The dataset and the model are saved in the local filesystem for simplicity. For most use-cases,
+    it is recommended to save the data and the model in a S3/GCS/other intermediate storage.
+
+Defining the Integrity Report Task
+----------------------------------
+
+The :func:`~deepchecks.tabular.suites.single_dataset_integrity` suite will be used to evaluate the training and test
```
shir22, line 71:
```suggestion
The :func:`~deepchecks.tabular.suites.single_dataset_integrity` suite will be used to evaluate the train and test
```

docs/source/user-guide/integrations/airflow.rst
```diff
@@ -0,0 +1,172 @@
...
+datasets. It will check for integrity issues and will save the reports in the ``suite_results`` directory.
```
shir22, line 72:
```suggestion
datasets. It will check for integrity issues and will save the output html reports to the ``suite_results`` directory.
```

docs/source/user-guide/integrations/airflow.rst
```diff
@@ -0,0 +1,172 @@
...
+
+        run_time = datetime.now().strftime("%Y-%m-%d_%H-%M-%S")
+        train_results.save_as_html(os.path.join(dir_path, f'train_integrity_{run_time}.html'))
+        test_results.save_as_html(os.path.join(dir_path, f'test_integrity_{run_time}.html'))
+
+Defining the Model Evaluation Task
+----------------------------------
+
+The :func:`~deepchecks.tabular.suites.model_evaluation` suite will be used to evaluate the model itself.
+It will check for model performance and overfit issues and will save the reports in the ``suite_results`` directory.
```
shir22, line 103:
```suggestion
It will check for model performance and overfit issues and will save the report to the ``suite_results`` directory.
```

docs/source/user-guide/integrations/airflow.rst
```diff
@@ -0,0 +1,172 @@
...
+        )
+
+    load_adult_dataset >> integrity_report
+    load_adult_dataset >> load_adult_model >> evaluation_report
+
+
+And that's it! In order to run the dag, make sure you place the file in your DAGs folder referenced in your
+``airflow.cfg``. The default location for your DAGs is ``~/airflow/dags``.
+
+After that, you can run the dag using the following command:
```
shir22, line 168: This is a bit confusing, as you wrote that we've scheduled the dag...

Do you mean - the DAG can be also manually triggered (for a single run) by using the following command?
After that you need to initialize (the scheduled) DAG by running the following command...
(whichever - fix phrasing accordingly)

examples/integrations/airflow/deepchecks_airflow_tutorial.py
```diff
@@ -0,0 +1,115 @@
+from datetime import datetime, timedelta
```
shir22, line 1: 1. Is it possible for the rst file to take "code blocks" as lines from here?
(so that in the document it will be a reference to line numbers instead of copy-paste of code...?)
Worth checking (and if so implementing) for finding a good and maintainable solution for the integration tutorial status (that is currently double...)
2. What's the purpose of this file?
(Is it quite easy to run it as-is? can we test it?
and do we think it's simpler than copy-pasting from the docs these code blocks?)

examples/integrations/airflow/README.rst
```diff
@@ -0,0 +1,6 @@
+Deepchecks Airflow Integration
+==============================
+
+This directory contains the implementation of the Airflow integration. It is not
+intended to be used directly, but rather to be used as a reference for
```
shir22, line 5:
```suggestion
intended to be used as-is, but rather to serve as a reference for
```

docs/source/user-guide/integrations/airflow.rst
noamzbr, line 33:
```suggestion
    data_path = os.path.join(os.getcwd(), "DATA_PATH")  # For demo only. Replace that with a S3/GCS other than local filesystem
```

docs/source/user-guide/integrations/airflow.rst
noamzbr, line 71: Best practice would be to use this suite on the dataset before splitting, so either demonstrate that, say that or say it's training & production data.

examples/integrations/airflow/deepchecks_airflow_tutorial.py
ItayGabbay: For 1 - unfortunately, I can't see a good maintainable solution here. If we want explanations within the code it can be an ipynb, and then refer the doc page to a Colab link. If it's a simple Python script - then probably have it inside the RST. I know it's not maintainable, will have to decide what's the best way to do that.

2 - in order to get it tested we need airflow as a dependency, which I don't think will be a good idea. The script itself is quite easy to run assuming you have airflow installed.

docs/source/user-guide/integrations/airflow.rst
ItayGabbay: I did not understand

shir22 (approved):

PR #1417, opened 2022-05-10
Run the phishing integrity suite on a Dataset instead of a DataFrame.

noamzbr:


docs/source/tutorials/tabular/plot_phishing_urls.py
```diff
@@ -151,9 +151,16 @@
 integ_suite = single_dataset_integrity()
 
 #%%
-# We will now run that suite on our dataframe:
+# We will now run that suite on our data. While running on the native DataFrame is possible in some cases, it is
+# recommended to wrap it with the ``deepchecks.tabular.Dataset`` object instead, to give
+# the package a bit more context, namely what is the label column, and whether
+# we have a datetime column (we have, as an index, so we'll set
+# ``set_datetime_from_dataframe_index=True``), or any categorical features (we have
+# none after one-hot encoding them, so we'll set ``cate_features=[]`` explicitly).
```
shir22, line 159:
```suggestion
# none after one-hot encoding them, so we'll set ``cat_features=[]`` explicitly).
```

docs/source/tutorials/tabular/plot_phishing_urls.py
```diff
@@ -234,14 +241,13 @@
...
 
 #%%
-# First, we have to wrap the dataframes in ``deepchecks.Dataset`` objects to give
-# the package a bit more context, namely what is the label column, and whether
-# we have a datetime column (we have, as an index, so we'll set
-# ``set_datetime_from_dataframe_index=True``), or any categorical features (we have
-# none after one-hot encoding them, so we'll set ``cate_features=[]`` explicitly).
+# Now that we have separate train and test DataFrames, we create two ``deepchecks.tabular.Dataset`` objects to enable
+# this suite and the next one to run addressing the train and test dataframes according to their role. That here
+# we pass the label as a column instead of a column name, because we've seperated the feature DataFrame from the target.
```
shir22, line 246:
```suggestion
# Now that we have separate train and test DataFrames, we will create two ``deepchecks.tabular.Dataset`` objects to enable
# this suite and the next one to run addressing the train and test dataframes according to their role. Notice that here
# we pass the label as a column instead of a column name, because we've seperated the feature DataFrame from the target.
```

shir22 (approved):

ItayGabbay (approved):

PR #1419, opened 2022-05-11
Simple feature fixes

### LINKED ISSUE #1004: Add "per class" to SimpleFeatureContribution
### DoD - 
Update check, tests and notebook with new feature of showing check SimpleFeatureContribution per class as well

### Inputs
Add input "per_class" boolean flag to check

### Outputs
Check will output its regular (?) graphs per class instead of overall

### Logic - 
Run the ppscore package on data multiple times (per class) by transforming the label to "one-vs-all".

### Comments:
Make sure that ppscore package doesn't already have a flag that supports this input and may run this code faster

JKL98ISR:
closes #1004

docs have now example per class and per all. added min pps to init + test
also fixed not showing negative difference in the condition (if train < test)


docs/source/checks/vision/methodology/plot_simple_feature_contribution.py
```diff
@@ -180,7 +174,7 @@ def mnist_batch_to_images_with_bias(batch):
 
 #%%
 
-check = SimpleFeatureContribution()
+check = SimpleFeatureContribution(per_class=False)
```
nirhutnik, line 177: But... you just reverted the doc to use the per_class=False. But we do want to show the per_class=True feature. We actually want to show both.

deepchecks/core/check_utils/single_feature_contribution_utils.py
```diff
@@ -52,8 +52,11 @@ def pps_df_to_trace(s_pps: pd.Series, name: str):
                   )
 
 
-def get_single_feature_contribution(train_df: pd.DataFrame, train_label_name: Optional[Hashable], test_df: pd.DataFrame,
-                                    test_label_name: Optional[Hashable], ppscore_params: dict, n_show_top: int,
+def get_single_feature_contribution(train_df: pd.DataFrame, train_label_name: Optional[Hashable],
+                                    test_df: pd.DataFrame,
+                                    test_label_name: Optional[Hashable], ppscore_params: dict,
+                                    n_show_top: int,
+                                    min_pps_to_show: float = 0.05,
```
nirhutnik, line 59: You didn't add min_pps_to_show to the check's parameters...
To this one or to single feature

docs/source/checks/vision/methodology/plot_simple_feature_contribution.py
JKL98ISR: it has both

deepchecks/core/check_utils/single_feature_contribution_utils.py
JKL98ISR: i added to simple feature, should i add to single too?

docs/source/checks/vision/methodology/plot_simple_feature_contribution.py
JKL98ISR: mnist is per class

deepchecks/core/check_utils/single_feature_contribution_utils.py
nirhutnik: Yes, I think so. Also missed that it was added to the check, sorry

docs/source/checks/vision/methodology/plot_simple_feature_contribution.py
nirhutnik: :+1:

deepchecks/core/check_utils/single_feature_contribution_utils.py
JKL98ISR: added

nirhutnik:
1. MNIST after bias - there's rarely any difference... Did you insert bias both into train and test? It should be inserted differently, if at all (can only insert to train, but good to insert a bit into test)

2. COCO - because there is already a bit of bias, I would insert only into train. Again, the differences here as well are not significant. We want to have an example with a real difference.

ItayGabbay (approved):

PR #1420, opened 2022-05-11
[BUGFIX] Serialization logic - additional fixes.

### LINKED ISSUE #1392: [BUG] Suite display not rendered correct in IFrame
**Describe the bug**
When displaying suite result in IFrame (for example in colab) the style is not rendered correctly (very narrow and tables are ugly)

image

**To Reproduce**
Run in colab:
```
from deepchecks.tabular.datasets.classification import iris
from deepchecks.tabular.suites import single_dataset_integrity

train, test = iris.load_data()
single_dataset_integrity().run(train)
```

yromanyshyn:
resolves #1392 but contains much more modifications

List of changes:

== 1 == 

added a new type of serialization protocol, called `IPythonSerializer`.
Serializers that implement this kind of protocol must return a list of objects that implements `IPython.core.formatters` traits.
For more info please refer to the IPython [documentation](https://ipython.readthedocs.io/en/stable/api/generated/IPython.core.formatters.html#IPython.core.formatters.MimeBundleFormatter)

We need this type of protocol in order to make sure that the check/suite result display works properly in different environments like JupyterLab/Classical Notebooks, especially when widgets are not enabled or not supported.

== 2 ==

added `display_check`, `show`, `save_as_html` methods to the `CheckFailure` type.

== 3 ==

now, 'Widgets UI' is also used in the colab env.


deepchecks/core/serialization/check_result/wandb.py
```diff
@@ -32,12 +32,11 @@
...
-try:
-    if t.TYPE_CHECKING:
-        from wandb.sdk.data_types.base_types.wb_value import \
-            WBValue  # pylint: disable=unused-import
-except ImportError:
-    pass
+
+if t.TYPE_CHECKING:
+    from wandb.sdk.data_types.base_types.wb_value import \
+        WBValue  # pylint: disable=unused-import
```
JKL98ISR, line 38: you should keep the try catch because it was only added to the newest wandb version and it is only for type checking and we shouldn't make users update it

deepchecks/core/check_result.py
```diff
@@ -146,40 +159,71 @@ def __init__(
...
+        return {'name': check_name, 'params': parameters, 'header': header,
+                'summary': get_docs_summary(self.check, with_doc_link=with_doc_link)}
 
-        Parameters
-        ----------
-        unique_id : str
-            The unique id given by the suite that displays the check.
-        show_additional_outputs : bool
-            Boolean that controls if to show additional outputs.
+    def get_header(self) -> str:
```
JKL98ISR, line 169: remove methods that are already in BaseCheckResult

deepchecks/core/serialization/suite_result/wandb.py
```diff
@@ -21,11 +21,13 @@
...
 
 try:
-    if t.TYPE_CHECKING:
-        from wandb.sdk.data_types.base_types.wb_value import \
-            WBValue  # pylint: disable=unused-import
-except ImportError:
-    pass
+    from wandb.sdk.data_types.base_types.wb_value import \
+        WBValue  # pylint: disable=unused-import
+except ImportError as error:
```
JKL98ISR, line 26: see other wandb comment

JKL98ISR (changes requested):

deepchecks/core/serialization/common.py
```diff
@@ -336,3 +338,29 @@ def concatv_images(images, gap=10):
...
 
         return dst
+
+
+T = t.TypeVar('T')
+DeepIterable = t.Iterable[t.Union[T, 'DeepIterable[T]']]
+
+
+def flatten(l: DeepIterable[T]) -> t.Iterable[T]:
+    """Flatten nested terables."""
```
JKL98ISR, line 348:
```suggestion
    """Flatten nested iterables."""
```

deepchecks/core/check_result.py
yromanyshyn: removed

deepchecks/core/serialization/check_result/wandb.py
yromanyshyn: from python docs

>typing.TYPE_CHECKING - A special constant that is assumed to be True by 3rd party static type checkers. It is False at runtime.

I am not sure if that try-except is needed here/ in general it will never raise an error, and static type checker probably will mark it as unresolved import/

deepchecks/core/serialization/common.py
yromanyshyn: fixed

deepchecks/core/serialization/suite_result/wandb.py
yromanyshyn: replaced with `if t.TYPE_CHECKING` statement

deepchecks/core/serialization/check_result/wandb.py
JKL98ISR: oh ok

JKL98ISR (approved):

deepchecks/core/suite.py
```diff
@@ -143,24 +173,27 @@ def save_as_html(
...
+        """
         if file is None:
             file = 'output.html'
         if isinstance(file, str):
             file = create_new_file_name(file)
 
         if as_widget is True:
             widget_to_html(
-                widget=self.to_widget(unique_id=output_id),
+                widget=self.to_widget(unique_id=unique_id or get_random_string(n=25)),
```
JKL98ISR, line 189: why the change?

JKL98ISR (changes requested):

deepchecks/core/suite.py
yromanyshyn: previously `output_id` was declared above `if file is None:` now I just replaced it with one statement/

it means that if a user did not pass a `unique_id` we need to generate it by ourselves in order to make sure that navigation links will be rendered

JKL98ISR (approved):

JKL98ISR (approved):
check.show doesn't work in jupyter lab

JKL98ISR (changes requested):
check.show doesn't work in jupyter lab

JKL98ISR (approved):
did a mistake while checking :)

PR #1427, opened 2022-05-12
[issue-1365] plotly figures width/height

### LINKED ISSUE #1365: [BUG] remove width/height in plotly graphs
**Describe the bug**
we set the pixel width/height instead of it being automatic

yromanyshyn:
resolves #1365 

I have experimented with height/width options, and noticed that:
- yup, it is Indeed responsive to the window/screen size
- it looks like by default (without explicitly provided width/height) it uses something like - `width: 100%; height: max-content;`. Added screenshots below to illustrate this.

IMO, in some cases more compact figure (with explicit width/height) could look better/
It all depends on the situation 

also, should I attach screenshots of the affected checks?

-- with `width=800, height=500` -- 
image

-- without explicit width/height --
image


deepchecks/utils/distribution/plot.py
```diff
@@ -204,17 +204,15 @@ def feature_distribution_traces(train_column,
 
         train_density = get_density(train_column, xs)
         test_density = get_density(test_column, xs)
-        bars_width = (x_range_to_show[1] - x_range_to_show[0]) / 100
+        # bars_width = (x_range_to_show[1] - x_range_to_show[0]) / 100
```
matanper, line 207: I believe this won't look well here without width

matanper:
Our problem is that when specifying size, then on small screens it doesn't scale down and gets cropped. If it's possible it would be best to have something like "max-width"/"max-height"

yromanyshyn:
@matanper 
> Our problem is that when specifying size, then on small screens it doesn't scale down and gets cropped. If it's possible it would be best to have something like "max-width"/"max-height"

I took a look at the Figure constructor docstring, unfortunately, it only has width/height parameters and it supports only 'px' units

```
height - Sets the plot's height (in px).
width - Sets the plot's width (in px).
```

deepchecks/utils/distribution/plot.py
yromanyshyn: not big difference as for me, but you are right it is not related to our problem, will revert


-- with width --

image

-- without width --

image

yromanyshyn:
attaching screenshots to illustrate how current modifications will affect plots. In general, the only difference probably is that now plots are always stretched horizontally up to the parent container width 

-- SingleFeatureContribution --

image

-- SingleFeatureContributionTrainTest --

image

-- WholeDatasetDrift --

image
image

-- IdentifierLeakage --

image

-- CalibrationScore --

image

-- RobustnessReport -- 

image

matanper:
It looks not so good, I think we should drop this issue for now

ItayGabbay:
Can we control the parent container size? @matanper @yromanyshyn

matanper:
> @ItayGabbay
> 
> > Can we control the parent container size? @matanper @yromanyshyn
> 
> we probably can add some CSS rule for this if they have an id attr or some common CSS class, but:
> 
> * it could have an effect on the entire notebook appearance and style
> * it could have an effect on all ipywidgets
> * different environments (jupyterlab, classical notebook, ...) probably will require different approaches
> * **it is a weird hack/**
> 
> note:
> 
> under parent container, I mean cell output div tag or div tag of a rendered ipywidget

I agree it's a bit hacky but it's not so bad. Let's say we create class PlotlyWrapper.
For HTML serializer we just do a `<div style="max-width: 400px">{fig}</div>`
For widgets serializer we will wrap it in `HBox([fig], Layout(width=400))`

yromanyshyn:
@matanper 

> I agree it's a bit hacky but it's not so bad. Let's say we create class PlotlyWrapper which accepts in ctor a plotly figure.
For HTML serializer we just do a <div style="max-width: 400px">{fig}</div>
For widgets serializer we will wrap it in HBox([fig], Layout(width=400))

yes it will work, but from my current point of view I am not sure if it is worth adding it/
How small screen should be that a user will feel discomfort with a plot of h:500&w:700?/
Or are you expecting that they will use smartphones to view results?

Plus, there is some technical complexity. In short, yes we can add that kind of wrappers but they will not help if the widgets extension is not enabled/supported, in more detail below.

By default, we are using `WidgetSerializer` for check results and indeed it will not be hard to add an additional wrapper for the figure in this case, but we do not use `HtmlSerializer` anymore as a fallback serializer (in cases when widgets extension is not enabled or not supported), instead, we have `IPythonSerializer` that can ensure that results will be displayed in any type of environment whether it is Jupyterlab or Classical Notebooks, and that is something that is hard to achieve with `HtmlSerializer` (`HtmlSerializer` is used only in `save_as_html` methods). To give you an insight about why it is so, here is how plotly figures work in that two environments:

-- Jupyter lab --

`figure._repr_mimebundle_()` is called and its output (a dictionary with items like `text/html`, `application/vnd.plotly.v1+json`) is sent to the client (jupyterlab frontend). If  the `jupyterlab-plotly` extension on the client  is installed then `application/vnd.plotly.v1+json` item will be chosen and the plot will be displayed without any problem, if the extension is not enabled/installed then `text/html` item will be chosen but plot will not be displlayed because it requires `requirejs` library that is not present in case of jupyterlab

-- Classical Notebook --

`figure._repr_mimebundle_()` is called and its output (a dictionary with items like `text/html`, `application/vnd.plotly.v1+json`) is sent to the frontend, there `text/html` item will be chosen and it will work without any problem because it is formed exactly for the classical notebooks

**Note:** 
the is no way for us (at least an easy way) to figure out whether a user uses jupyterlab UI or classical notebook UI, moreover, if `jupyter-server` was started up with 'jupyterlab' and 'nbclassic' extensions enabled then it becomes possible to open the same kernel(notebook) in two UIs silmontenuasly

**Note:**
ipywidgets library does not serialize plotly figures to the HTML before sending them to the UI, it serializes figures to some kind of special mime-type `application/vnd.jupyter.widget-view+json` which is then consumed by the widgets extension

yromanyshyn:
@matanper 

> I agree it's a bit hacky but it's not so bad. Let's say we create class PlotlyWrapper which accepts in ctor a plotly figure.
For HTML serializer we just do a <div style="max-width: 400px">{fig}</div>
For widgets serializer we will wrap it in HBox([fig], Layout(width=400))

added to the `html.CheckResultSerializer` and `widget.CheckResultSerializer`

matanper:
> but they will not help if the widgets extension is not enabled/supported

not sure I understand why. If the widgets are disabled, we are still using them in the IPython serializer to generate the HTML, no? then we can also there set the size

yromanyshyn:
@matanper 

> 
> > but they will not help if the widgets extension is not enabled/supported
> 
> not sure I understand why. If the widgets are disabled, we are still using them in the IPython serializer to generate the HTML, no? then we can also there set the size

no, within html serializer we do not use ipywidgets
but yes you are partially correct, but that is true only for google collab or when we save results as an HTML
take a look at the `CheckResult.display_check` and `CheckResult.save_as_html` for more info

yromanyshyn:
@matanper 

sorry, only now noticed, that I misread your previous comment
no, we do not use IPython serializer to generate HTML
the task of the IPython serializer is to produce a list of [IPython formatters objects](https://ipython.readthedocs.io/en/stable/api/generated/IPython.core.formatters.html)

yromanyshyn:
I create a new [PR](https://github.com/deepchecks/deepchecks/pull/1483) for this issue

PR #1432, opened 2022-05-12
Fix string mismatch performance

matanper:
1. The sum(count) was the function which took forever
2. The string_baseform is slow. Managed to cut it by half but it's still like 10 seconds for 1M strings


noamzbr (commented):
Increase the n_samples for both checks.

noamzbr (approved):

Issue #1409, opened 2022-05-10
[FEAT] Single dataset performance check


ItayGabbay:
This check should calculate a single performance metric on a single dataset.

PR #1442, opened 2022-05-15
1409 feat single dataset performance check

TheSolY:
Issue #1409 
New check: SingleDatasetScalarPerformance to summarize a given metric over a given dataset to return a performance score as a scalar.

### DIFF
--- a/deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
+++ b/deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
@@ -35,6 +35,7 @@ class SingleDatasetScalarPerformance(SingleDatasetCheck):
         The function to reduce the scores tensor into a single scalar. For metrics that return a scalar use None
         (default).
     """
+    
     def __init__(self,
                  metric: Metric = None,
                  reduce: t.Callable = None,
@@ -42,10 +43,10 @@ class SingleDatasetScalarPerformance(SingleDatasetCheck):
         super().__init__(**kwargs)
         self.metric = metric
         self.reduce = reduce
-        if 'metric_name' in kwargs.keys():
+        if 'metric_name' in kwargs:
             self.metric_name = kwargs['metric_name']
 
-        if 'reduce_name' in kwargs.keys():
+        if 'reduce_name' in kwargs:
             self.reduce_name = kwargs['reduce_name']
 
     def initialize_run(self, context: Context, dataset_kind: DatasetKind.TRAIN):
@@ -72,12 +73,12 @@ class SingleDatasetScalarPerformance(SingleDatasetCheck):
         """Compute the metric result using the ignite metrics compute method and reduce to a scalar."""
         metric_result = self.metric.compute()
         if self.reduce is not None:
-            if type(metric_result) is float:
+            if isinstance(metric_result, float):
                 warnings.warn(SyntaxWarning('Metric result is already scalar, skipping reduce operation'))
                 result_value = metric_result
             else:
                 result_value = float(self.reduce(metric_result))
-        elif type(metric_result) is float:
+        elif isinstance(metric_result, float):
             result_value = metric_result
         else:
             raise DeepchecksValueError(f'The metric {self.metric.__class__} return a non-scalar value, '
@@ -97,7 +98,7 @@ class SingleDatasetScalarPerformance(SingleDatasetCheck):
         return CheckResult(result_dict)
 
     def add_condition_greater_than(self, threshold: float) -> ConditionResult:
-        """Add condition - the result is greater than the threshold"""
+        """Add condition - the result is greater than the threshold."""
         def condition(check_result):
             if check_result['score'] > threshold:
                 return ConditionResult(ConditionCategory.PASS)
@@ -107,7 +108,7 @@ class SingleDatasetScalarPerformance(SingleDatasetCheck):
         return self.add_condition(f'Score is greater than {threshold}', condition)
 
     def add_condition_greater_equal_to(self, threshold: float) -> ConditionResult:
-        """Add condition - the result is greater than the threshold"""
+        """Add condition - the result is greater than the threshold."""
 
         def condition(check_result):
             if check_result['score'] >= threshold:
@@ -119,7 +120,7 @@ class SingleDatasetScalarPerformance(SingleDatasetCheck):
         return self.add_condition(f'Score is greater than or equal to {threshold}', condition)
 
     def add_condition_less_than(self, threshold: float) -> ConditionResult:
-        """Add condition - the result is greater than the threshold"""
+        """Add condition - the result is greater than the threshold."""
         def condition(check_result):
             if check_result['score'] < threshold:
                 return ConditionResult(ConditionCategory.PASS)
@@ -129,7 +130,7 @@ class SingleDatasetScalarPerformance(SingleDatasetCheck):
         return self.add_condition(f'Score is less than {threshold}', condition)
 
     def add_condition_less_equal_to(self, threshold: float) -> ConditionResult:
-        """Add condition - the result is greater than the threshold"""
+        """Add condition - the result is greater than the threshold."""
 
         def condition(check_result):
             if check_result['score'] <= threshold:
--- a/deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
+++ b/deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
@@ -19,6 +19,7 @@ from deepchecks.core.errors import DeepchecksValueError
 from deepchecks.vision.vision_data import TaskType
 from deepchecks.vision.metrics_utils.object_detection_precision_recall import \
     ObjectDetectionAveragePrecision
+import warnings
 
 __all__ = ['SingleDatasetScalarPerformance']
 
@@ -48,6 +49,7 @@ class SingleDatasetScalarPerformance(SingleDatasetCheck):
                 self.metric = Accuracy()
             elif context.train.task_type == TaskType.OBJECT_DETECTION:
                 self.metric = ObjectDetectionAveragePrecision()
+                self.reduce = torch.nanmean
             else:
                 raise DeepchecksValueError('For task types other then classification or object detection, '
                                            'pass a metric explicitly')
@@ -64,10 +66,10 @@ class SingleDatasetScalarPerformance(SingleDatasetCheck):
         metric_result = self.metric.compute()
         if self.reduce is not None:
             if type(metric_result) is float:
-                Warning('Metric result is already scalar, skipping reduction')
+                warnings.warn(SyntaxWarning('Metric result is already scalar, skipping reduction'))
                 return CheckResult(metric_result)
             else:
-                return CheckResult(self.reduce(metric_result))
+                return CheckResult(float(self.reduce(metric_result)))
         elif type(metric_result) is float:
             return CheckResult(metric_result)
         else:
@@ -84,6 +86,36 @@ class SingleDatasetScalarPerformance(SingleDatasetCheck):
                 return ConditionResult(ConditionCategory.FAIL, details)
         return self.add_condition(f'Score is grater than {threshold}', condition)
 
+    def add_condition_greater_equal_to(self, threshold: float) -> ConditionResult:
+        """Add condition - the result is greater than the threshold"""
 
+        def condition(check_result):
+            if check_result >= threshold:
+                return ConditionResult(ConditionCategory.PASS)
+            else:
+                details = f'The result is not greater than or equal to {threshold}'
+                return ConditionResult(ConditionCategory.FAIL, details)
 
+        return self.add_condition(f'Score is grater than or equal to {threshold}', condition)
 
+    def add_condition_less_than(self, threshold: float) -> ConditionResult:
+        """Add condition - the result is greater than the threshold"""
+        def condition(check_result):
+            if check_result < threshold:
+                return ConditionResult(ConditionCategory.PASS)
+            else:
+                details = f'The result is not less than {threshold}'
+                return ConditionResult(ConditionCategory.FAIL, details)
+        return self.add_condition(f'Score is less than {threshold}', condition)
+
+    def add_condition_less_equal_to(self, threshold: float) -> ConditionResult:
+        """Add condition - the result is greater than the threshold"""
+
+        def condition(check_result):
+            if check_result <= threshold:
+                return ConditionResult(ConditionCategory.PASS)
+            else:
+                details = f'The result is not less than or equal to {threshold}'
+                return ConditionResult(ConditionCategory.FAIL, details)
+
+        return self.add_condition(f'Score is less than or equal to {threshold}', condition)
--- a/deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
+++ b/deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
@@ -69,30 +69,34 @@ class SingleDatasetScalarPerformance(SingleDatasetCheck):
         if self.reduce is not None:
             if type(metric_result) is float:
                 warnings.warn(SyntaxWarning('Metric result is already scalar, skipping reduce operation'))
-                return CheckResult(metric_result)
+                result_value = metric_result
             else:
-                return CheckResult(float(self.reduce(metric_result)))
+                result_value = float(self.reduce(metric_result))
         elif type(metric_result) is float:
-            return CheckResult(metric_result)
+            result_value = metric_result
         else:
-            raise DeepchecksValueError(f'The metric {repr(self.metric)} return a non-scalar value, '
+            raise DeepchecksValueError(f'The metric {self.metric.__class__} return a non-scalar value, '
                                        f'please specify a reduce function or choose a different metric')
+        result_dict = {'score': result_value,
+                       'metric': self.metric.__class__.__name__,
+                       'reduce': None if self.reduce is None else self.reduce.__name__}
+        return CheckResult(result_dict)
 
     def add_condition_greater_than(self, threshold: float) -> ConditionResult:
         """Add condition - the result is greater than the threshold"""
         def condition(check_result):
-            if check_result > threshold:
+            if check_result['score'] > threshold:
                 return ConditionResult(ConditionCategory.PASS)
             else:
                 details = f'The score is not greater than {threshold}'
                 return ConditionResult(ConditionCategory.FAIL, details)
-        return self.add_condition(f'Score is grater than {threshold}', condition)
+        return self.add_condition(f'Score is greater than {threshold}', condition)
 
     def add_condition_greater_equal_to(self, threshold: float) -> ConditionResult:
         """Add condition - the result is greater than the threshold"""
 
         def condition(check_result):
-            if check_result >= threshold:
+            if check_result['score'] >= threshold:
                 return ConditionResult(ConditionCategory.PASS)
             else:
                 details = f'The score is not greater than or equal to {threshold}'
@@ -103,7 +107,7 @@ class SingleDatasetScalarPerformance(SingleDatasetCheck):
     def add_condition_less_than(self, threshold: float) -> ConditionResult:
         """Add condition - the result is greater than the threshold"""
         def condition(check_result):
-            if check_result < threshold:
+            if check_result['score'] < threshold:
                 return ConditionResult(ConditionCategory.PASS)
             else:
                 details = f'The score is not less than {threshold}'
@@ -114,7 +118,7 @@ class SingleDatasetScalarPerformance(SingleDatasetCheck):
         """Add condition - the result is greater than the threshold"""
 
         def condition(check_result):
-            if check_result <= threshold:
+            if check_result['score'] <= threshold:
                 return ConditionResult(ConditionCategory.PASS)
             else:
                 details = f'The score is not less than or equal to {threshold}'
--- a/deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
+++ b/deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
@@ -10,16 +10,19 @@
 #
 """Module containing a check for computing a scalar performance metric for a single dataset."""
 import typing as t
+import warnings
+
 import torch
-from ignite.metrics import Metric, Accuracy
+from ignite.metrics import Accuracy, Metric
+
 from deepchecks.core import CheckResult, ConditionResult, DatasetKind
 from deepchecks.core.condition import ConditionCategory
-from deepchecks.vision import Batch, Context, SingleDatasetCheck
 from deepchecks.core.errors import DeepchecksValueError
-from deepchecks.vision.vision_data import TaskType
+from deepchecks.vision import Batch, Context, SingleDatasetCheck
 from deepchecks.vision.metrics_utils.object_detection_precision_recall import \
     ObjectDetectionAveragePrecision
-import warnings
+from deepchecks.vision.vision_data import TaskType
+
 
 __all__ = ['SingleDatasetScalarPerformance']
 
--- a/deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
+++ b/deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
@@ -29,7 +29,7 @@ class SingleDatasetScalarPerformance(SingleDatasetCheck):
 
     Parameters
     ----------
-        metric: Metric,  default: None
+        metric: Metric, default: None
         An ignite.Metric object whose score should be used. If None is given, use the default metric.
         reduce: torch function, default: None
         The function to reduce the scores tensor into a single scalar. For metrics that return a scalar use None
--- a/deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
+++ b/deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
@@ -9,6 +9,7 @@
 # ----------------------------------------------------------------------------
 #
 """Module containing a check for computing a scalar performance metric for a single dataset."""
+import numbers
 import typing as t
 import warnings
 
@@ -36,18 +37,25 @@ class SingleDatasetScalarPerformance(SingleDatasetCheck):
         reduce: torch function, default: None
         The function to reduce the scores tensor into a single scalar. For metrics that return a scalar use None
         (default).
+        metric_name: str, default: None
+        A name for the metric to show in the check results.
+        reduce_name: str, default: None
+        A name for the reduce function to show in the check results.
+
     """
 
     def __init__(self,
                  metric: Metric = None,
                  reduce: t.Callable = None,
+                 metric_name: str = None,
+                 reduce_name: str = None,
                  **kwargs):
         super().__init__(**kwargs)
         self.metric = metric
         self.reduce = reduce
 
-        self.metric_name = kwargs.get('metric_name') or (metric.__class__.__name__ if metric else None)
-        self.reduce_name = kwargs.get('reduce_name') or (reduce.__name__ if reduce else None)
+        self.metric_name = metric_name or (metric.__class__.__name__ if metric else None)
+        self.reduce_name = reduce_name or (reduce.__name__ if reduce else None)
 
     def initialize_run(self, context: Context, dataset_kind: DatasetKind.TRAIN):
         """Initialize the metric for the check, and validate task type is relevant."""
@@ -78,10 +86,10 @@ class SingleDatasetScalarPerformance(SingleDatasetCheck):
         """Compute the metric result using the ignite metrics compute method and reduce to a scalar."""
         metric_result = self.metric.compute()
         if self.reduce is not None:
-            if isinstance(metric_result, float):
+            if isinstance(metric_result, numbers.Real):
                 warnings.warn(SyntaxWarning('Metric result is already scalar, skipping reduce operation.'
                                             'Pass reduce=None to prevent this'))
-                result_value = metric_result
+                result_value = float(metric_result)
             else:
                 result_value = float(self.reduce(metric_result))
         elif isinstance(metric_result, float):
--- a/deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
+++ b/deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
@@ -42,6 +42,11 @@ class SingleDatasetScalarPerformance(SingleDatasetCheck):
         super().__init__(**kwargs)
         self.metric = metric
         self.reduce = reduce
+        if 'metric_name' in kwargs.keys():
+            self.metric_name = kwargs['metric_name']
+
+        if 'reduce_name' in kwargs.keys():
+            self.reduce_name = kwargs['reduce_name']
 
     def initialize_run(self, context: Context, dataset_kind: DatasetKind.TRAIN):
         """Initialize the metric for the check, and validate task type is relevant."""
@@ -77,9 +82,18 @@ class SingleDatasetScalarPerformance(SingleDatasetCheck):
         else:
             raise DeepchecksValueError(f'The metric {self.metric.__class__} return a non-scalar value, '
                                        f'please specify a reduce function or choose a different metric')
+
+        metric_name = self.metric_name if hasattr(self, 'metric_name') else self.metric.__class__.__name__
+        if hasattr(self, 'reduce_name'):
+            reduce_name = self.reduce_name
+        elif self.reduce is None:
+            reduce_name = None
+        else:
+            reduce_name = self.reduce.__name__
+
         result_dict = {'score': result_value,
-                       'metric': self.metric.__class__.__name__,
-                       'reduce': None if self.reduce is None else self.reduce.__name__}
+                       'metric': metric_name,
+                       'reduce': reduce_name}
         return CheckResult(result_dict)
 
     def add_condition_greater_than(self, threshold: float) -> ConditionResult:
--- a/deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
+++ b/deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
@@ -12,9 +12,11 @@
 import typing as t
 import torch
 from ignite.metrics import Metric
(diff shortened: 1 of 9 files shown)


deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
```diff
@@ -0,0 +1,121 @@
...
+# Copyright (C) 2021-2022 Deepchecks (https://www.deepchecks.com)
+#
+# This file is part of Deepchecks.
+# Deepchecks is distributed under the terms of the GNU Affero General
+# Public License (version 3 or later).
+# You should have received a copy of the GNU Affero General Public License
+# along with Deepchecks.  If not, see <http://www.gnu.org/licenses/>.
+# ----------------------------------------------------------------------------
+#
+"""Module containing a single dataset with a signle scalar output performance check."""
```
noamzbr, line 11:
```suggestion
"""Module containing a check for computing a scalar performance metric for a single dataset."""
```

deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
```diff
@@ -0,0 +1,121 @@
...
+from deepchecks.vision.vision_data import TaskType
+from deepchecks.vision.metrics_utils.object_detection_precision_recall import \
+    ObjectDetectionAveragePrecision
+import warnings
+
+__all__ = ['SingleDatasetScalarPerformance']
+
+
+class SingleDatasetScalarPerformance(SingleDatasetCheck):
+    """Summarize a given metric over a given dataset to return a performance score as a scalar.
```
noamzbr, line 28:
```suggestion
    """Calculate a scalar performance metrics for a given model and a given dataset.
```
Just a suggestion for improvement, you can think perhaps of something else

deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
```diff
@@ -0,0 +1,121 @@
...
+
```
noamzbr, line 29: Add an additional short paragraph explaining the role of the reduce and when should you use it

deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
```diff
@@ -0,0 +1,121 @@
...
+    Parameters
+    ----------
+        metric : default: None
```
noamzbr, line 32:
```suggestion
        metric: Metric, default: None
```

deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
```diff
@@ -0,0 +1,121 @@
...
+        An ignite.Metric object whose score should be used. If None are given, use the default metrics.
+        reduce: torch function, default: None (for metrics that already return a scalar)
```
noamzbr, line 34: I think it's better to explain the "(for metrics that already return a scalar)" in the line below, and keep the definition line clean.

deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
```diff
@@ -0,0 +1,121 @@
...
+    def initialize_run(self, context: Context, dataset_kind: DatasetKind.TRAIN):
+        """Initialize the metric for the check, and validate task type is relevant."""
+        if self.metric is None:
+            if context.train.task_type == TaskType.CLASSIFICATION:
+                self.metric = Accuracy()
+            elif context.train.task_type == TaskType.OBJECT_DETECTION:
+                self.metric = ObjectDetectionAveragePrecision()
+                self.reduce = torch.nanmean
+            else:
+                raise DeepchecksValueError('For task types other then classification or object detection, '
```
noamzbr, line 54: Would consider a rephrase, maybe "No metric was passed. A metric must be passed for task types other than classification or object detection."

deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
```diff
@@ -0,0 +1,121 @@
...
+        label = batch.labels
+        prediction = batch.predictions
+        self.metric.update((prediction, label))
+
+    def compute(self, context: Context, dataset_kind: DatasetKind.TRAIN) -> CheckResult:
+        """Compute the metric result using the ignite metrics compute method and reduce to a scalar."""
+        metric_result = self.metric.compute()
+        if self.reduce is not None:
+            if type(metric_result) is float:
+                warnings.warn(SyntaxWarning('Metric result is already scalar, skipping reduction'))
```
noamzbr, line 69:
```suggestion
                warnings.warn(SyntaxWarning('Metric result is already scalar, skipping reduce operation'))
```

deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
```diff
@@ -0,0 +1,121 @@
...
+
+    def update(self, context: Context, batch: Batch, dataset_kind: DatasetKind.TRAIN):
+        """Update the metrics by passing the batch to ignite metric update method."""
+        label = batch.labels
+        prediction = batch.predictions
+        self.metric.update((prediction, label))
+
+    def compute(self, context: Context, dataset_kind: DatasetKind.TRAIN) -> CheckResult:
+        """Compute the metric result using the ignite metrics compute method and reduce to a scalar."""
+        metric_result = self.metric.compute()
```
noamzbr, line 66: How does metric_result look like? I would like it to contain the value, the metric name and the reduction name, so that whoever is receiving this result will know for what metric and what reduce was used.

deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
```diff
@@ -0,0 +1,121 @@
...
+            raise DeepchecksValueError(f'The metric {repr(self.metric)} return a non-scalar value, '
+                                       f'please specify a reduce function or choose a different metric')
+
+    def add_condition_greater_than(self, threshold: float) -> ConditionResult:
+        """Add condition - the result is greater than the threshold"""
+        def condition(check_result):
+            if check_result > threshold:
+                return ConditionResult(ConditionCategory.PASS)
+            else:
+                details = f'The result is not greater than {threshold}'
```
noamzbr, line 85:
```suggestion
                details = f'The score is not greater than {threshold}'
```

deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
```diff
@@ -0,0 +1,121 @@
...
+                return ConditionResult(ConditionCategory.FAIL, details)
+        return self.add_condition(f'Score is grater than {threshold}', condition)
```
noamzbr, line 87:
```suggestion
        return self.add_condition(f'Score is greater than {threshold}', condition)
```

deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
```diff
@@ -0,0 +1,121 @@
...
+
+    def add_condition_greater_equal_to(self, threshold: float) -> ConditionResult:
+        """Add condition - the result is greater than the threshold"""
+
+        def condition(check_result):
+            if check_result >= threshold:
+                return ConditionResult(ConditionCategory.PASS)
+            else:
+                details = f'The result is not greater than or equal to {threshold}'
```
noamzbr, line 96:
```suggestion
                details = f'The score is not greater than or equal to {threshold}'
```

deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
```diff
@@ -0,0 +1,121 @@
...
+from deepchecks.core.errors import DeepchecksValueError
+from deepchecks.vision.vision_data import TaskType
+from deepchecks.vision.metrics_utils.object_detection_precision_recall import \
+    ObjectDetectionAveragePrecision
+import warnings
+
+__all__ = ['SingleDatasetScalarPerformance']
+
+
+class SingleDatasetScalarPerformance(SingleDatasetCheck):
```
noamzbr, line 27: Opening the discussion of the name here. Perhaps SingleDatasetPerformance is simpler, though less explicit.

deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
```diff
@@ -0,0 +1,127 @@
...
+    """Calculate a performance metric as a scalar for a given model and a given dataset.
+
+    Parameters
+    ----------
+        metric: Metric,  default: None
```
noamzbr, line 32:
```suggestion
        metric: Metric, default: None
```

deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
```diff
@@ -0,0 +1,127 @@
...
+                result_value = metric_result
+            else:
+                result_value = float(self.reduce(metric_result))
+        elif type(metric_result) is float:
+            result_value = metric_result
+        else:
+            raise DeepchecksValueError(f'The metric {self.metric.__class__} return a non-scalar value, '
+                                       f'please specify a reduce function or choose a different metric')
+        result_dict = {'score': result_value,
+                       'metric': self.metric.__class__.__name__,
```
noamzbr, line 81: What is self.metric.__class__.__name__ for example for ObjectDetectionAveragePrecision? Because if it's 'ObjectDetectionAveragePrecision' we cant show it to the user like this. We probably should receive a dict {metric_name: metric_value} for the metric argument if that is the case, so the user / our default values will be the ones deciding the display name.

docs/source/checks/vision/performance/plot_single_dataset_scalar_performance.py
```diff
@@ -0,0 +1,70 @@
...
+
+**Structure:**
+
+* `What Is the Purpose of the Check? <#what-is-the-purpose-of-the-check>`__
+* `Generate data an model <#generate-data-and-model>`__
+* `Run the check <#run-the-check>`__
+
+What Is the Purpose of the Check?
+=================================
+This check returns a metric result as a single scalar, which is especially useful for monitoring.
```
noamzbr, line 16:
```suggestion
This check returns a metric result as a single scalar, which is especially useful for monitoring a model in production.
```

noamzbr, line 16: Explain here about the relationship between metric return type and the reduce function.

docs/source/checks/vision/performance/plot_single_dataset_scalar_performance.py
```diff
@@ -0,0 +1,70 @@
...
+result = check.run(train_ds, mnist_model)
+
+#%%
+# The result is a dictionary with the following fields: score - the actual result,
+# metric - the name of metric used, reduce - the name of the reduce function used.
+result
+
+#%%
+# Now we will run a check with parameters, to use a metric and a reduce function different from the defaults.
+# We will also pass names for them, so that the result dictionary will look neat.
```
noamzbr, line 51:
```suggestion
# We will also pass names for them, so that the return value will look neat.
```

docs/source/checks/vision/performance/plot_single_dataset_scalar_performance.py
```diff
@@ -0,0 +1,70 @@
...
+Single Dataset Scalar Performance
+******************************
+
+This notebooks provides an overview for using and understanding single dataset scalar performance check.
+
+**Structure:**
+
+* `What Is the Purpose of the Check? <#what-is-the-purpose-of-the-check>`__
+* `Generate data an model <#generate-data-and-model>`__
+* `Run the check <#run-the-check>`__
```
noamzbr, line 12: Add the define a condition section here.

deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
```diff
@@ -0,0 +1,148 @@
...
+                 reduce: t.Callable = None,
+                 **kwargs):
+        super().__init__(**kwargs)
+        self.metric = metric
+        self.reduce = reduce
+        if 'metric_name' in kwargs:
+            self.metric_name = kwargs['metric_name']
+
+        if 'reduce_name' in kwargs:
+            self.reduce_name = kwargs['reduce_name']
```
matanper, line 52: it would be better to just have the var as None instead of checking `not hasattr(self, 'reduce_name')`
```suggestion
self.reduce_name = kwargs.get('reduce_name') or (reduce.__name__ if reduce else None)
```

deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
```diff
@@ -0,0 +1,148 @@
...
+
+    def initialize_run(self, context: Context, dataset_kind: DatasetKind.TRAIN):
+        """Initialize the metric for the check, and validate task type is relevant."""
+        if self.metric is None:
+            if context.train.task_type == TaskType.CLASSIFICATION:
+                self.metric = Accuracy()
```
matanper, line 58: I think it's better to not define here manually the metrics, but maybe in a more global way in the metric utils. @noamzbr WDYT?

deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
matanper: same for metric_name

deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
```diff
@@ -0,0 +1,148 @@
...
+        label = batch.labels
+        prediction = batch.predictions
+        self.metric.update((prediction, label))
+
+    def compute(self, context: Context, dataset_kind: DatasetKind.TRAIN) -> CheckResult:
+        """Compute the metric result using the ignite metrics compute method and reduce to a scalar."""
+        metric_result = self.metric.compute()
+        if self.reduce is not None:
+            if isinstance(metric_result, float):
+                warnings.warn(SyntaxWarning('Metric result is already scalar, skipping reduce operation'))
```
matanper, line 83: is there something the user can do to prevent this warning? something like 
"Metric result is already scalar, skipping reduce operation. Pass reduce=None to prevent this"

deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
```diff
@@ -0,0 +1,148 @@
...
+        result_dict = {'score': result_value,
+                       'metric': metric_name,
+                       'reduce': reduce_name}
+        return CheckResult(result_dict)
+
+    def add_condition_greater_than(self, threshold: float) -> ConditionResult:
+        """Add condition - the result is greater than the threshold."""
+        def condition(check_result):
+            if check_result['score'] > threshold:
+                return ConditionResult(ConditionCategory.PASS)
```
matanper, line 110: We want to start passing details also on passing conditions. so maybe just do:
```
details = "Score found: {score}"
```
and pass it both pass & fail

matanper: same for the rest of conditions

matanper (approved):
Overall looking good 😃

deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
JKL98ISR: better to just add it reduce_name and metric_name to the function parameters with None as default (same for metric_name)

deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
JKL98ISR: maybe also check if it's a numpy float? (or maybe even a number/numpy number)

deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
JKL98ISR: + 1 for that, also no need for if/else if you're returning a value

tests/vision/checks/performance/single_dataset_scalar_performance_test.py
```diff
@@ -0,0 +1,92 @@
...
+    check = SingleDatasetScalarPerformance()
+
+    # Act
+    result = check.run(mnist_dataset_train, mock_trained_mnist, device=device)
+
+    # Assert
+    # scalar
+    assert_that(type(result.value['score']), equal_to(float))
+    # metric
+    assert_that(result.value['score'], greater_than_or_equal_to(0))
```
JKL98ISR, line 56: better to use close_to with precision of 0.001, it helps to see if anything is random in our code

tests/vision/checks/performance/single_dataset_scalar_performance_test.py
```diff
@@ -0,0 +1,92 @@
...
+    # Arrange
+    check = SingleDatasetScalarPerformance()
+    check.add_condition_greater_than(0.5)
+    check.add_condition_less_equal_to(0.2)
+
+    # Act
+    result = check.run(mnist_dataset_train, mock_trained_mnist, device=device)
+
+    # Assert
+    assert_that(result.conditions_results[0].is_pass)
```
JKL98ISR, line 69: also we try to use less test function for vision so you can add the condition to the same function that asserts value and assert the condition results there

tests/vision/checks/performance/single_dataset_scalar_performance_test.py
```diff
@@ -0,0 +1,92 @@
...
+    check = SingleDatasetScalarPerformance()
+
+    # Act
+    result = check.run(coco_train_visiondata, mock_trained_yolov5_object_detection, device=device)
+
+    # Assert
+    # scalar
+    assert_that(type(result.value['score']), equal_to(float))
+    # metric
+    assert_that(result.value['score'], greater_than_or_equal_to(0))
```
JKL98ISR, line 35: better to use close_to with precision of 0.001, it helps to see if anything is random in our code

tests/vision/checks/performance/single_dataset_scalar_performance_test.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+def test_detection_w_params(coco_train_visiondata, mock_trained_yolov5_object_detection, device):
+    # params that should run normally
+    check = SingleDatasetScalarPerformance(reduce=torch.max, reduce_name='torch_max')
+    result = check.run(coco_train_visiondata, mock_trained_yolov5_object_detection, device=device)
+    assert_that(type(result.value['score']), equal_to(float))
```
JKL98ISR, line 42: assert the value too and use close_to with precision of 0.001, it helps to see if anything is random in our code
(sry for duplicated comments)

tests/vision/checks/performance/single_dataset_scalar_performance_test.py
JKL98ISR: also no need to assert type is float unless type float is really necessary (not sure about that), because if you compare to a number it is a number

docs/source/checks/vision/performance/plot_single_dataset_scalar_performance.py
```diff
@@ -0,0 +1,80 @@
...
+# Define a Condition
+# ==================
+# We can define a condition to validate that our model performance score is above or below a certain threshold.
+# The condition is defined as a function that takes the results of the check as input and
+# returns a ConditionResult object.
+
+check = SingleDatasetScalarPerformance()
+check.add_condition_greater_than(0.5)
+result = check.run(train_ds, mnist_model)
+result
```
JKL98ISR, line 78: for same checks that already ran with just a new condition we usually use result.show(with_additional_outputs=False)

deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
```diff
@@ -0,0 +1,148 @@
...
+        """Add condition - the result is greater than the threshold."""
+        def condition(check_result):
+            if check_result['score'] < threshold:
+                return ConditionResult(ConditionCategory.PASS)
+            else:
+                details = f'The score is not less than {threshold}'
+                return ConditionResult(ConditionCategory.FAIL, details)
+        return self.add_condition(f'Score is less than {threshold}', condition)
+
+    def add_condition_less_equal_to(self, threshold: float) -> ConditionResult:
```
JKL98ISR, line 138: is there a real reason for less equal? isn't less enough? just that equal for floats have no real meaning (same comment for greater equal)

JKL98ISR (changes requested):
noticing I didn't press request changes so putting it

deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
TheSolY: Why specifically numpy?

deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
TheSolY: @JKL98ISR I think the users will feel more comfortable with this, about the syntax I thought we decided to change the negative conditions to positive ones

deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
TheSolY: Setting global defaults requires some group discussion

deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
JKL98ISR: seems possible for a meric to return numpy no?

tests/vision/checks/performance/single_dataset_scalar_performance_test.py
TheSolY: It's necessary to assert the type because it's supposed to be used inside a monitoring system that is likely to expect a specific type.

deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
matanper: so not urgent to this PR, we can open issue for that

deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
matanper: `isinstance(metric_result, numbers.Number)` will catch both

TheSolY: number.Number doesn't have a default conversion to float, we're expecting an ignite Metric so i don't see a case that will return numpy.

deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
JKL98ISR: why not in function params?

deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
JKL98ISR: ignite metrics can technically return anything

deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
```diff
@@ -0,0 +1,138 @@
...
+    def initialize_run(self, context: Context, dataset_kind: DatasetKind.TRAIN):
+        """Initialize the metric for the check, and validate task type is relevant."""
+        if self.metric is None:
+            if context.train.task_type == TaskType.CLASSIFICATION:
+                self.metric = Accuracy()
+                if self.metric_name is None:
+                    self.metric_name = 'accuracy'
+            elif context.train.task_type == TaskType.OBJECT_DETECTION:
+                self.metric = ObjectDetectionAveragePrecision()
+                if self.metric_name is None:
```
noamzbr, line 61: I think user shouldn't be able to pass a name if he didn't pass a metric.

deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
```diff
@@ -0,0 +1,138 @@
...
+class SingleDatasetScalarPerformance(SingleDatasetCheck):
+    """Calculate a performance metric as a scalar for a given model and a given dataset.
+
+    Parameters
+    ----------
+        metric: Metric, default: None
+        An ignite.Metric object whose score should be used. If None is given, use the default metric.
+        reduce: torch function, default: None
+        The function to reduce the scores tensor into a single scalar. For metrics that return a scalar use None
+        (default).
```
noamzbr, line 38: I think metric_name and reduce_name should be official parameters that appear in the docstrings.

docs/source/checks/vision/performance/plot_single_dataset_scalar_performance.py
```diff
@@ -0,0 +1,80 @@
...
+
+mnist_model = mnist.load_model()
+train_ds = mnist.load_dataset(train=True, object_type='VisionData')
+
+#%%
+# Run the check
+# -------------
+# We will run the check with the model defined above.
+#
+# The check will use the default classification metric - ignite's Accuracy.
```
noamzbr, line 45: Link to https://pytorch.org/ignite/generated/ignite.metrics.Accuracy.html

deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
JKL98ISR: +1 also, why even pass a metric_name? can't we just use the class name and maybe just split by camel case if we really want

deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
TheSolY: Since they only affect the display and not the actual functionality, they seem more like kwargs to me, but I'm open to change

deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
TheSolY: So parameters instead of kwargs?

deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
TheSolY: I think the user should be able to pass a name because they might want to use the default metric with a different name like "default" or something else according to their conventions.

deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
JKL98ISR: how much does it matter if we have the or not? won't name be enough?

deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
```diff
@@ -0,0 +1,148 @@
...
+
+    def compute(self, context: Context, dataset_kind: DatasetKind.TRAIN) -> CheckResult:
+        """Compute the metric result using the ignite metrics compute method and reduce to a scalar."""
+        metric_result = self.metric.compute()
+        if self.reduce is not None:
+            if isinstance(metric_result, float):
+                warnings.warn(SyntaxWarning('Metric result is already scalar, skipping reduce operation'))
+                result_value = metric_result
+            else:
+                result_value = float(self.reduce(metric_result))
```
JKL98ISR, line 86: about same comment on the floats, do we care? maybe it's a numpy float, why not let it be a numpy float

deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
TheSolY: @noamzbr your call

deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
TheSolY: I'm less concerned about numpy floats and more about non float numbers such as integers

deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
noamzbr: I'm for Check params - we have other checks with params that control displays, and I think it should be obvious to the user how to control that (display is important)

deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
noamzbr: if it's gonna be serialized, having it being a pure python float is good IMO

deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
noamzbr: yep

deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
JKL98ISR: well the json/wandb serialization is doing un_numpy to the value anyways, isn't critical but the user might prefer numpy

deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
TheSolY: changed to number.Real

deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
```diff
@@ -0,0 +1,146 @@
...
+            if check_result['score'] >= threshold:
+                return ConditionResult(ConditionCategory.PASS, details)
+            else:
+                return ConditionResult(ConditionCategory.FAIL, details)
+        return self.add_condition(f'Score is greater than or equal to {threshold}', condition)
+
+    def add_condition_less_than(self, threshold: float) -> ConditionResult:
+        """Add condition - the result is greater than the threshold."""
+        def condition(check_result):
+            details = f'The score {self.metric_name} is {check_result["score"]}'
```
matanper, line 129: are you sure inside the condition it have access to `self`?
maybe you need to do
```
def condition(check_result, metric_name=self.metric_name):
```

deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
TheSolY: Changed to params

deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
TheSolY: It works...

noamzbr (approved):

docs/source/checks/vision/performance/plot_single_dataset_scalar_performance.py
```diff
@@ -0,0 +1,82 @@
+# -*- coding: utf-8 -*-
+"""
+Single Dataset Scalar Performance
+******************************
```
ItayGabbay, line 4:
```suggestion
Single Dataset Scalar Performance
*********************************
```

docs/source/checks/vision/performance/plot_single_dataset_scalar_performance.py
```diff
@@ -0,0 +1,82 @@
...
+* `Define a condition <#define-a-condition>`__
+
+What Is the Purpose of the Check?
+=================================
+This check returns a metric result as a single scalar, which is especially useful for monitoring a model in production.
+
+Some metrics return a single score, but others return a tensor of scores.
+
+For example, Precision returns a tensor in the size of the number of classes. In that case, we will use a
+reduce function- a function that aggregates the scores into a scalar.
```
ItayGabbay, line 22:
```suggestion
reduce function - a function that aggregates the scores into a scalar.
```

docs/source/checks/vision/performance/plot_single_dataset_scalar_performance.py
ItayGabbay: We need this to be in the same length otherwise the docs build raises warnings

docs/source/checks/vision/performance/plot_single_dataset_scalar_performance.py
```diff
@@ -0,0 +1,82 @@
...
+#%%
+# Now we will run a check with parameters, to use a metric and a reduce function different from the defaults.
+# We will also pass names for them, so that the return value will look neat.
+from ignite.metrics import Precision
+from torch import nanmean
+
+check = SingleDatasetScalarPerformance(Precision(), nanmean, metric_name='precision', reduce_name='mean')
+result = check.run(train_ds, mnist_model)
+result.value
+#%%
```
ItayGabbay, line 70:
```suggestion

#%%
```

deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
```diff
@@ -0,0 +1,146 @@
...
+    ----------
+        metric: Metric, default: None
+        An ignite.Metric object whose score should be used. If None is given, use the default metric.
+        reduce: torch function, default: None
+        The function to reduce the scores tensor into a single scalar. For metrics that return a scalar use None
+        (default).
+        metric_name: str, default: None
+        A name for the metric to show in the check results.
+        reduce_name: str, default: None
+        A name for the reduce function to show in the check results.
```
ItayGabbay, line 43: No need to indent (it will look bad on the docs when indenting). Take a look at other checks.

deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
```diff
@@ -0,0 +1,146 @@
...
+        """Update the metrics by passing the batch to ignite metric update method."""
+        label = batch.labels
+        prediction = batch.predictions
+        self.metric.update((prediction, label))
+
+    def compute(self, context: Context, dataset_kind: DatasetKind.TRAIN) -> CheckResult:
+        """Compute the metric result using the ignite metrics compute method and reduce to a scalar."""
+        metric_result = self.metric.compute()
+        if self.reduce is not None:
+            if isinstance(metric_result, numbers.Real):
```
ItayGabbay, line 89: What will happen if the result is a Tensor of a real number (Tensor(0.687))?

ItayGabbay (approved):

deepchecks/vision/checks/performance/single_dataset_scalar_performance.py
JKL98ISR: Also kinda continuing on gabbay comment, is it possible the reduce will return a tensor? Cuz if so I think it will also fail

JKL98ISR (approved):

PR #1450, opened 2022-05-16
Doc mainpage redirect

ItayGabbay:
#### Reference Issues/PRs

#### Any other comments?

---

### DIFF
--- a/docs/source/user-guide/index.rst
+++ b/docs/source/user-guide/index.rst
@@ -53,3 +53,4 @@ Integrations
     integrations/pytest
     integrations/h2o
     integrations/airflow
+    general/exporting_results/examples/plot_exports_output_to_wandb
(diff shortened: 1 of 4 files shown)


docs/source/user-guide/index.rst
```diff
@@ -52,4 +52,5 @@ Integrations
     integrations/spark_databricks
     integrations/pytest
     integrations/h2o
-    integrations/airflow
\ No newline at end of file
+    integrations/airflow
+    general/exporting_results/examples/plot_exports_output_to_wandb
```
shir22, line 56: I see we can combine sphinx galler and rst files in same hierarchy? (good news)

Shouldn't it also be deleted from the other toc...?

shir22 (approved):

docs/source/user-guide/index.rst
ItayGabbay: image

shir22: You like it better like this? (feels a bit weird to have it double-pointed, would put it only in integrations...)

PR #1460, opened 2022-05-17
Replace logger.warnings with warnings.warn

### LINKED ISSUE #1328: [FEAT] Use warnings in a single way
Right now we have 2 types of warnings, using `logger` and using `warnings`. We should decide which one is better to use and then update in all places to use that one

matanper:
close #1328

### DIFF
--- a/deepchecks/tabular/dataset.py
+++ b/deepchecks/tabular/dataset.py
@@ -9,7 +9,6 @@
 # ----------------------------------------------------------------------------
 #
 """The dataset module containing the tabular Dataset class and its functions."""
-import logging
 # pylint: disable=inconsistent-quotes,protected-access
 import typing as t
 import warnings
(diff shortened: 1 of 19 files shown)


tests/base/dataset_test.py
```diff
@@ -1041,17 +1042,16 @@ def random_classification_dataframe(n_samples=100, n_features=5) -> pd.DataFrame
 
 def test_cat_features_warning(iris, caplog):
```
ItayGabbay, line 1043: you don't need the caplog anymore

deepchecks/utils/distribution/drift.py
```diff
@@ -294,7 +294,7 @@ def calc_drift_and_plot(train_column: pd.Series,
             'largest_difference': 'largest difference between categories'
         }
         train_data_percents = dist_traces[0].y.sum()
```
ItayGabbay, line 296: ?

ItayGabbay (approved):

deepchecks/tabular/dataset.py
```diff
@@ -12,6 +12,7 @@
 import logging
```
Nadav-Barak, line 12: what is logging used for?

JKL98ISR:
remove all `logger = logging.getLogger('deepchecks')` and such (no import logging should be there)

PR #1468, opened 2022-05-18
Hide "nothing found" in suite result display

matanper:
Reassured suite display sections:
1. All checks with conditions. If no display then hide the "additional output"
2. Other checks without condition but with display
3. Rest of the checks: failures and checks without condition & without display

image


JKL98ISR:
is this pr about updating section or just hiding nothing found? because they were different issues if I recall correctly.
also can you add before/after screenshots?

matanper:
> is this pr about updating section or just hiding nothing found? because they were different issues if I recall correctly. also can you add before/after screenshots?

It's about fixing the bug which is linked to the PR

PR #1471, opened 2022-05-18
[Issue 1410] Progress Bar fix

### LINKED ISSUE #1410: [BUG] progress bars in suite run without widgets are duplicating
**Describe the bug**
progress bars in suite run without widgets are duplicating
(in checks it's ok so probably we can copy from that)
**To Reproduce**
Run the suite without widgets enabled

**Expected behavior**
bars continuing the progress

**Screenshots**
Screen Shot 2022-05-10 at 13 43 06

yromanyshyn:
resolves #1410

This bug (that is described in issue 1410) happens because tqdm does not tolerate when something else is writing at the same time to the stdout/stderr (Attaching screenshot below to illustrate this)

image
image

The way to solve this is to give tqdm a separate files stream and to use IPyhton ability to redraw the same output cell. More in the code.

Screenshots to illustrate the result:

image


matanper:
Where do you remove the progress bars from the screen at the end?

ItayGabbay:
@JKL98ISR

deepchecks/utils/ipython.py
```diff
@@ -116,39 +129,189 @@ def is_widgets_use_possible() -> bool:
...
+    for _ in create_progress_bar(
+        list(range(1)),
+        name=name,
+        unit=unit
+    ):
+        yield
+
+
+# TODO:
+# NOTE:
```
JKL98ISR, line 201: why is it commented?

JKL98ISR:
> Where do you remove the progress bars from the screen at the end?

which screen?

yromanyshyn:
@matanper 

> Where do you remove the progress bars from the screen at the end?

I do not remove them`s, they are left
if they should be removed at the end I will add that

deepchecks/utils/ipython.py
yromanyshyn: I will create another PR to finish that

in general, I need to find a way to determine which jupyter-server extensions are enabled in order to make it work correctly, currently, I cannot find how to do that

matanper (approved):

ItayGabbay (approved):

PR #1473, opened 2022-05-18
Fix grouping nan types in MixedNulls for pandas < 1.4

matanper:


Nadav-Barak:
Does this also address the latency problem he spoke of?

matanper:
> Does this also address the latency problem he spoke of?

No

noamzbr (commented):
Maybe easier to just use the id version for all, or is it slower?

matanper:
> Maybe easier to just use the id version for all, or is it slower?

It's definitely slower, don't know by how much. but using the "id version" for both doesn't remove a lot of complexity (it's 3 rows difference) so I think it's better to leave this like that

PR #1483, opened 2022-05-19
[Issue-1365][V2] Plotly figures width&height

### LINKED ISSUE #1365: [BUG] remove width/height in plotly graphs

yromanyshyn:
resolves #1365 
previous PR #1427 

attaching a zip file with a set of screenshots to demonstrate a difference between plots with width&height and without
(_there are too many of them to include directly in the PR description therefore attaching a zip file_)

[screenshots-issue-1365.zip](https://github.com/deepchecks/deepchecks/files/8727460/issue-1365.zip)

Description and notes regarding the screenshots:
- screenshots with name `{CheckName}-before.png` demonstrate plots with explicitly set width&height parameters
- screenshots with name `{CheckName}-after.png` demonstrate plots without width&height parameters
- screenshots with name `{CheckName}-wrapper.png` demonstrate plots width&height of which is limited by wrapper container (method suggested by the Matan; I used max-width: 900px; max-height: 400px;), 
-  a name of a screenshot like `{CheckName}.png` mean that width&height parameters for plots are not used for that check, and that screenshot demonstrates check output as it looks now.
- I left `height=200px` parameter for the drift plots even with the wrapper method enabled, otherwise it stretches the plot vertically too much (for example screenshots: **ImageDatasetDrift-after.png**, **ImageDatasetDrift-wrapper.png**)
- if you set width&height (with wrapper method enabled) that is bigger than max width&height of the wrapper container then plot will be partially hidden (take a look at the screenshot **ConfusionMatrixReport-overflow-hiden.png**)

== Summary ==
- some plots without explicitly set width&height look ugly
- a wrapper container for the plotly figures will not make any harm and looks good enough, but:    
     + we can add it only to the `WidgetSerializer`, and it will work only in envs where widgets are supported/enabled

== IMO ==

it is probably worth just to remove the width parameter but leave the height, and do not bother with any kind of wrappers, nevertheless, it is not a big deal when a plot is stretched horizontally, the user always can zoom in and zoom out to adjust a plot as he likes (ploty adds zoom-in/zoom-out buttons to each plot)


matanper:
About the wrapper, we can control it from the check, so if we want larger size than the wrapper will support it (it will at least add a scroll bar)

yromanyshyn:
@matanper 
>About the wrapper, we can control it from the check, so if we want larger size than the wrapper will support it (it will at least add a scroll bar)

yup, we can add something like `seirlization_kwargs: DIct[str, Any]` to the `CheckResult/SuiteResult` which then will be passed to the serializer

matanper:
> @matanper
> 
> > About the wrapper, we can control it from the check, so if we want larger size than the wrapper will support it (it will at least add a scroll bar)
> 
> yup, we can add something like `serialization_kwargs: Dict[str, Any]` to the `CheckResult/SuiteResult` which then will be passed to the serializer

I actually imagine it already in a check to do:
```
display = [Wrapper(fig, width=600, height=400))]
```
but let's wait to hear @shir22 and @ItayGabbay

ItayGabbay:
@yromanyshyn looks good. I reviewed the images, and I think that specifically for barcharts we do need the wrapper mechanism since it doesn't look so good. 
You mentioned that the wrapper object will work only on widgets. It won't work on regular HTML?

yromanyshyn:
@matanper 

> I actually imagine it already in a check to do:
> 
> display = [Wrapper(fig, width=600, height=400))]
> but let's wait to hear @shir22 and @ItayGabbay

that even better

@ItayGabbay 

> You mentioned that the wrapper object will work only on widgets. It won't work on regular HTML?

yes, we also will add that to the `HTMLSerializer`, but we use `HTMLSerializer` only in `save_as_html` and when `as_widget` flag is set to False. At all other places when widgets are not enabled, we use `IPythonSerializer` to make sure that results will be correctly displayed in all possible environments (jupyterlab, classical notebooks, etc).

`IPythonSerializer`s (comparing them other serializers) are a little bit specific. Their task is to serialize check/suite results to a list of objects that implement `IPython.formatter` traits. It is hard to describe IPython formatter/mimetypes/etc protocols in two-three sentences, but I think my comment from the previous PR describes it in the shortest way


> we have IPythonSerializer that can ensure that results will be displayed in any type of environment whether it is Jupyterlab or Classical Notebooks, and that is something that is hard to achieve with HtmlSerializer (HtmlSerializer is used only in save_as_html methods). To give you an insight about why it is so, here is how plotly figures work in that two environments:
> 
> -- Jupyter lab --
> 
> `figure._repr_mimebundle_()` is called (by the `IPython` display logic when you pass figure to the `display` function) and its output (a dictionary with items like `text/html`, `application/vnd.plotly.v1+json`) is sent to the client (jupyterlab frontend). If the `jupyterlab-plotly` extension on the client is installed then `application/vnd.plotly.v1+json` item will be chosen and the plot will be displayed without any problem, if the extension is not enabled/installed then `text/html` item will be chosen but plot will not be displlayed because it requires `requirejs` library that is not present in case of jupyterlab
> 
> **note:** 
> jupyterlab requires `jupyterlab-plotly` extension to make plotly figures work
> 
> 
> -- Classical Notebook --
> 
> `figure._repr_mimebundle_()` is called (by the `IPython` display logic when you pass figure to the `display` function) and its output (a dictionary with items like `text/html`, `application/vnd.plotly.v1+json`) is sent to the frontend, there `text/html` item will be chosen (because there are no any mime-type extension that can process `application/vnd.plotly.v1+json` item) and it will work without any problem because it is formed exactly for the classical notebooks
> 

actually, even shorter, this process could be described in a next way:

IPython delegates display logic to the objects that want to be displayed, it collects all possible representations (mime-types) of that objects and sends them in form of the dictionary (`Dict[mimetype, Any]`) to the frontend/UI, and which one representation/mime-type will be chosen depends upon frontend (_note: this is totally correct in case of plotly figures but I am not sure if it always works this way_)

deepchecks/utils/distribution/plot.py
```diff
@@ -163,23 +173,22 @@ def feature_distribution_traces(train_column,
     if is_categorical:
         traces, y_layout = _create_distribution_bar_graphs(train_column, test_column, max_num_categories,
                                                            show_categories_by)
-        xaxis_layout = dict(type='category')
+        xaxis_layout = dict(type='category', range=(-3, max_num_categories + 2))
```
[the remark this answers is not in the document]
yromanyshyn: > Also why use max_num_categories and not the actual number of categories?

that number determines how many bars will on the plot

matanper (approved):

deepchecks/utils/plot.py
```diff
@@ -207,7 +207,7 @@ def create_confusion_matrix_figure(confusion_matrix: np.ndarray, x: np.ndarray,
                 texttemplate=texttemplate))
     fig.data[0].colorbar.title = colorbar_title
     fig.update_layout(title=plot_title)
-    fig.update_layout(width=600, height=600)
+    # fig.update_layout(width=600, height=600)
```
matanper, line 210:
```suggestion
   fig.update_layout(height=600)
```

yromanyshyn: fixed

noamzbr (approved):

PR #1484, opened 2022-05-19
Nb/feat/correlation methods

Nadav-Barak:
Implementation of method as discussed in attached issue.

### DIFF
--- a/deepchecks/utils/correlation_methods.py
+++ b/deepchecks/utils/correlation_methods.py
@@ -75,7 +75,8 @@ def theil_u_correlation(x: Union[List, np.ndarray, pd.Series], y: Union[List, np
 
 
 def correlation_ratio(categorical_variable: Union[List, np.ndarray, pd.Series],
-                      numerical_variable: Union[List, np.ndarray, pd.Series]) -> float:
+                      numerical_variable: Union[List, np.ndarray, pd.Series],
+                      indexes_to_ignore: Union[List[bool], np.ndarray] = None) -> float:
     """
     Calculate the correlation ratio of numerical_variable to categorical_variable.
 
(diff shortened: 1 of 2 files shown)


deepchecks/utils/correlation_methods.py
```diff
@@ -0,0 +1,111 @@
...
+        Returns:
+        --------
+        float
+            Representing the correlation ratio between the features.
+    """
+    fcat, _ = pd.factorize(categorical_feature)
+    cat_num = np.max(fcat) + 1
+    y_avg_array = np.zeros(cat_num)
+    n_array = np.zeros(cat_num)
+    for i in range(0, cat_num):
```
nirhutnik, line 100: no need for 0 in range

deepchecks/utils/correlation_methods.py
```diff
@@ -0,0 +1,111 @@
...
+        -----------
+        categorical_feature: Union[List, np.ndarray, pd.Series]
+            A sequence of a categorical feature values without nulls
+        numerical_feature: Union[List, np.ndarray, pd.Series]
+            A sequence of a numerical feature values without nulls
+        Returns:
+        --------
+        float
+            Representing the correlation ratio between the features.
+    """
```
nirhutnik, line 95: If you require values without nulls, why not do a null validation?

deepchecks/utils/correlation_methods.py
```diff
@@ -0,0 +1,111 @@
...
+# Copyright (C) 2021-2022 Deepchecks (https://www.deepchecks.com)
+#
+# This file is part of Deepchecks.
+# Deepchecks is distributed under the terms of the GNU Affero General
+# Public License (version 3 or later).
+# You should have received a copy of the GNU Affero General Public License
+# along with Deepchecks.  If not, see <http://www.gnu.org/licenses/>.
+# ----------------------------------------------------------------------------
+#
+"""Module containing methods for calculating correlation between features."""
```
nirhutnik, line 11: change usages of "feature" to "variable" - feature is necessarily part of X, but these can be used for correlations with y as well.

Nadav-Barak: Good sugestion

deepchecks/utils/correlation_methods.py
Nadav-Barak: It is computationally better to it once for the all dataset in the check

deepchecks/utils/correlation_methods.py
```diff
@@ -0,0 +1,111 @@
...
+    Returns:
+    --------
+    float
+        Representing the conditional entropy
+    """
+    y_counter = Counter(y)
+    xy_counter = Counter(list(zip(x, y)))
+    total_occurrences = sum(y_counter.values())
+    s_xy = 0.0
+    for xy in xy_counter:
```
noamzbr, line 42: Can this be vectorized?

noamzbr (approved):

Issue #1338, opened 2022-04-27
[FEAT] Add "more info" data to conditions even if they didn't fail


shir22:
For example if a condition is on above 10% duplicates, and there were 8% duplicates, this info can be included and is useful for the user (both for a quick view and also in case they want to adapt threshold).
Especially since conditions are many time used in default suites and before customization...

This issue is relevant for all checks and conditions that have relevant "more info" to display also when condition passes.

- [X] Tabular
- [ ] Vision

PR #1485, opened 2022-05-19
Update conditions details for tabular integrity checks

matanper:
Partial PR of #1338 

image

### DIFF
--- a/deepchecks/tabular/checks/data_integrity/is_single_value.py
+++ b/deepchecks/tabular/checks/data_integrity/is_single_value.py
@@ -11,8 +11,6 @@
 """Module contains is_single_value check."""
 from typing import List, Union
 
-import pandas as pd
-
 from deepchecks.core import CheckResult, ConditionCategory, ConditionResult
 from deepchecks.tabular import Context, SingleDatasetCheck
 from deepchecks.utils.dataframes import select_from_dataframe
--- a/deepchecks/tabular/checks/data_integrity/is_single_value.py
+++ b/deepchecks/tabular/checks/data_integrity/is_single_value.py
@@ -11,6 +11,8 @@
 """Module contains is_single_value check."""
 from typing import List, Union
 
+import pandas as pd
+
 from deepchecks.core import CheckResult, ConditionCategory, ConditionResult
 from deepchecks.tabular import Context, SingleDatasetCheck
 from deepchecks.utils.dataframes import select_from_dataframe
@@ -48,7 +50,7 @@ class IsSingleValue(SingleDatasetCheck):
         Returns
         -------
         CheckResult
-            value is a boolean if there was at least one column with only one unique,
+            value of result is a dict of all columns with number of uniques values as: {column: number_of_uniques, ...}
             display is a series with columns that have only one unique
         """
         # Validate parameters
@@ -59,29 +61,32 @@ class IsSingleValue(SingleDatasetCheck):
 
         df = select_from_dataframe(df, self.columns, self.ignore_columns)
 
-        is_single_unique_value = (df.nunique(dropna=False) == 1)
+        num_unique_per_col = df.nunique(dropna=False)
+        is_single_unique_value = (num_unique_per_col == 1)
 
         if is_single_unique_value.any():
             # get names of columns with one unique value
             # pylint: disable=unsubscriptable-object
             cols_with_single = is_single_unique_value[is_single_unique_value].index.to_list()
-            value = list(cols_with_single)
             uniques = df.loc[:, cols_with_single].head(1)
             uniques.index = ['Single unique value']
             display = ['The following columns have only one unique value', uniques]
         else:
-            value = None
             display = None
 
-        return CheckResult(value, header='Single Value in Column', display=display)
+        return CheckResult(num_unique_per_col.to_dict(), header='Single Value in Column', display=display)
 
     def add_condition_not_single_value(self):
         """Add condition - not single value."""
         name = 'Does not contain only a single value'
 
         def condition(result):
-            if result:
-                return ConditionResult(ConditionCategory.FAIL, f'Found columns with a single value: {result}')
-            return ConditionResult(ConditionCategory.PASS)
+            single_value_cols = [k for k, v in result.items() if v == 1]
+            if single_value_cols:
+                details = f'Found {len(single_value_cols)} columns with a single value out of {len(result)} ' \
+                          f'columns: {single_value_cols}'
+                return ConditionResult(ConditionCategory.FAIL, details)
+            else:
+                return ConditionResult(ConditionCategory.PASS, f'Passed for {len(result)} columns')
 
         return self.add_condition(name, condition)
--- a/deepchecks/tabular/checks/data_integrity/is_single_value.py
+++ b/deepchecks/tabular/checks/data_integrity/is_single_value.py
@@ -48,7 +48,7 @@ class IsSingleValue(SingleDatasetCheck):
         Returns
         -------
         CheckResult
-            value of result is a dict of all columns with number of uniques values as: {column: number_of_uniques, ...}
+            value of result is a dict of all columns with number of unique values in format {column: number_of_uniques}
             display is a series with columns that have only one unique
         """
         # Validate parameters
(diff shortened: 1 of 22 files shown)


deepchecks/tabular/checks/data_integrity/is_single_value.py
```diff
@@ -48,7 +48,7 @@ def run_logic(self, context: Context, dataset_type: str = 'train') -> CheckResul
         Returns
         -------
         CheckResult
-            value is a boolean if there was at least one column with only one unique,
+            value of result is a dict of all columns with number of uniques values as: {column: number_of_uniques, ...}
```
nirhutnik, line 51:
```suggestion
            value of result is a dict of all columns with number of unique values in format {column: number_of_uniques}
```

deepchecks/tabular/checks/data_integrity/is_single_value.py
```diff
@@ -59,29 +59,32 @@ def run_logic(self, context: Context, dataset_type: str = 'train') -> CheckResul
 
         df = select_from_dataframe(df, self.columns, self.ignore_columns)
 
-        is_single_unique_value = (df.nunique(dropna=False) == 1)
+        num_unique_per_col = df.nunique(dropna=False)
```
nirhutnik, line 62: Make sure it doesn't count each none separately

deepchecks/tabular/checks/data_integrity/is_single_value.py
```diff
@@ -59,29 +59,32 @@ def run_logic(self, context: Context, dataset_type: str = 'train') -> CheckResul
...
         name = 'Does not contain only a single value'
 
         def condition(result):
-            if result:
-                return ConditionResult(ConditionCategory.FAIL, f'Found columns with a single value: {result}')
-            return ConditionResult(ConditionCategory.PASS)
+            single_value_cols = [k for k, v in result.items() if v == 1]
+            if single_value_cols:
+                details = f'Found {len(single_value_cols)} columns with a single value out of {len(result)} ' \
+                          f'columns: {single_value_cols}'
```
nirhutnik, line 85: No need for total number of columns here in my opinion. Can be just "found X columns with single value: [...]"

deepchecks/tabular/checks/data_integrity/mixed_data_types.py
```diff
@@ -58,7 +58,8 @@ def run_logic(self, context: Context, dataset_type: str = 'train') -> CheckResul
         Returns
         -------
         CheckResult
-            DataFrame with rows ('strings', 'numbers') for any column with mixed types.
+            value is a dict with column name as key and dict of 'strings' and 'numbers' ratio for any column with
```
nirhutnik, line 61:
```suggestion
            value is a dict where the key is the column name as key and the value is the ratio 'strings' and 'numbers' for any column with
```

deepchecks/tabular/checks/data_integrity/mixed_nulls.py
```diff
@@ -69,7 +68,9 @@ def run_logic(self, context: Context, dataset_type: str = 'train') -> CheckResul
         Returns
         -------
         CheckResult
-            DataFrame with columns ('Column Name', 'Value', 'Count', 'Percentage') for any column which
+            Value is dict with columns as key, and dict of null values as value:
+            {column: {null_value: {count: x, percent: y}, ...}, ...}
+            display is DataFrame with columns ('Column Name', 'Value', 'Count', 'Percentage') for any column which
             have more than 1 null values.
```
nirhutnik, line 74:
```suggestion
            has more than 1 null values.
```

deepchecks/tabular/checks/data_integrity/mixed_nulls.py
```diff
@@ -69,7 +68,9 @@ def run_logic(self, context: Context, dataset_type: str = 'train') -> CheckResul
         Returns
         -------
         CheckResult
-            DataFrame with columns ('Column Name', 'Value', 'Count', 'Percentage') for any column which
+            Value is dict with columns as key, and dict of null values as value:
+            {column: {null_value: {count: x, percent: y}, ...}, ...}
+            display is DataFrame with columns ('Column Name', 'Value', 'Count', 'Percentage') for any column which
```
nirhutnik, line 73:
```suggestion
            display is DataFrame with columns ('Column Name', 'Value', 'Count', 'Percentage') for any column that
```

deepchecks/tabular/checks/data_integrity/mixed_nulls.py
```diff
@@ -164,18 +163,14 @@ def add_condition_different_nulls_not_more_than(self, max_allowed_null_types: in
...
-                num_nulls = len(nulls)
-                if num_nulls > max_allowed_null_types:
-                    not_passing_columns[column] = num_nulls
+            not_passing_columns = [k for k, v in result.items() if len(v) > max_allowed_null_types]
             if not_passing_columns:
-                return ConditionResult(ConditionCategory.FAIL,
-                                       'Found columns with amount of null types above threshold: '
-                                       f'{not_passing_columns}')
+                details = f'Found {len(not_passing_columns)} columns with amount of null types above threshold out ' \
+                          f'of {len(result)} columns: {not_passing_columns}'
```
nirhutnik, line 169: Again, total number of columns is not interesting in my opinion

deepchecks/tabular/checks/data_integrity/single_feature_contribution.py
```diff
@@ -125,10 +125,12 @@ def condition(value: t.Dict[Hashable, float]) -> ConditionResult:
             }
 
             if failed_features:
-                message = f'Features with PPS above threshold: {failed_features}'
+                message = f'Found {len(failed_features)} features with PPS above threshold out of {len(value)} ' \
+                          f' features: {failed_features}'
```
nirhutnik, line 129: Same comment about columns. Will stop commenting here

deepchecks/tabular/checks/data_integrity/string_length_out_of_bounds.py
```diff
@@ -249,19 +247,17 @@ def compare_outlier_ratio(result: Dict):
...
             if not_passing_columns:
-                return ConditionResult(ConditionCategory.WARN,
-                                       f'Found columns with outliers ratio above threshold: '
-                                       f'{not_passing_columns}')
+                details = f'Found {len(not_passing_columns)} columns with outliers ratio above threshold out of ' \
+                          f'{len(result)} columns: {not_passing_columns}'
+                return ConditionResult(ConditionCategory.WARN, details)
             else:
-                return ConditionResult(ConditionCategory.PASS)
+                details = f'Passed for {len(result)} columns'
```
nirhutnik, line 259: Why not give any data?

tests/tabular/checks/integrity/string_mismatch_test.py
```diff
@@ -154,6 +159,7 @@ def test_condition_percent_variants_no_more_than_pass():
     # Assert
     assert_that(result, has_items(
         equal_condition_result(is_pass=True,
+                               details='Passed for 1 string columns',
```
nirhutnik, line 162: Here and in other places, we say "columns" even when the number is one.
Maybe add some sort of if so it won't be plural? I know it's annoying, and users will get why it's like that, but it's more professional like this.
Can also create a common method for that  - 

Suggestion:
def add_plural(num: int, postfix: str='s'):
    return postfix if num == 1 else ''

and then use as:
f'Found {num_columns} column{add_plural(num_columns)} ... '
and you can change the postfix into "es" where relevant.
Not great, but can change the experience a bit

nirhutnik (approved):

deepchecks/tabular/checks/data_integrity/string_length_out_of_bounds.py
matanper: if there are no outliers, there isn't any data to give

deepchecks/tabular/checks/data_integrity/is_single_value.py
ItayGabbay: I agree with Matan

ItayGabbay (approved):

PR #1491, opened 2022-05-22
Static model

JKL98ISR:


deepchecks/tabular/context.py
```diff
@@ -102,7 +196,23 @@ def __init__(self,
         if features_importance is not None:
             if not isinstance(features_importance, pd.Series):
                 raise DeepchecksValueError('features_importance must be a pandas Series')
-
+        # TODO: a little trickery
```
JKL98ISR, line 199: ignore that block it is for my testing

deepchecks/tabular/context.py
```diff
@@ -30,6 +32,106 @@
 ]
 
 
+class _FakeModel:
+    """Contains all the data + properties the user has passed to a check/suite, and validates it seamlessly.
+
+    Parameters
+    ----------
+    train :Dataset
```
ItayGabbay, line 40:
```suggestion
    train: Dataset
```

deepchecks/tabular/context.py
```diff
@@ -30,6 +32,106 @@
...
+        if dataset_kind == DatasetKind.TRAIN:
+            return self.train
+        if dataset_kind == DatasetKind.TEST:
+            return self.test
+        raise DeepchecksValueError(f'Unexpected dataset kind {dataset_kind}')
+
+    def _validate_data(self, data: pd.DataFrame, dataset_kind: DatasetKind):
+        sub_index = list(data.index)[:10]
+        return (data.loc[sub_index].fillna('nan') == \
+            self._get_dataset(dataset_kind).loc[sub_index].fillna('nan')).all().all()
```
ItayGabbay, line 110: I didn't like it

deepchecks/tabular/context.py
```diff
@@ -102,7 +208,23 @@ def __init__(self,
...
+                y_proba_test = model.predict_proba(test.features_columns)
+        if model is not None and hasattr(model, 'predict'):
+            if train is not None:
+                y_pred_train = model.predict(train.features_columns)
+            if test is not None:
+                y_pred_test = model.predict(test.features_columns)
+        model = None
+        # we do it after validation because the model we are creating isn't valid (we are always valid just not here)
+        if model is None and \
+            not pd.Series([y_pred_train, y_pred_test, y_proba_train, y_proba_test]).isna().all():
```
ItayGabbay, line 225: do we want to raise an exception here? if one of the arrays contains na?

deepchecks/tabular/context.py
```diff
@@ -30,16 +33,118 @@
...
+        self.ind_map = defaultdict(dict)
+
+        if train is not None and test is not None:
+            # check if datasets have same indexes
+            if set(train.data.index) & set(test.data.index):
+                train.data.index = map(lambda x: f'train-{x}', list(train.data.index))
+                test.data.index = map(lambda x: f'test-{x}', list(test.data.index))
+                warnings.warn('train and test datasets have common index - adding "train"/"test"'
+                              ' prefixes. To avoid that provide datasets with no common indexes '
+                              'or don\'t use the static predictions feature.')
```
noamzbr, line 77: As a user, I won't know what a "static prediction" is

deepchecks/tabular/context.py
```diff
@@ -30,16 +33,118 @@
...
+        if dataset_kind == DatasetKind.TEST:
+            return self.test
+        raise DeepchecksValueError(f'Unexpected dataset kind {dataset_kind}')
+
+    def _validate_data(self, data: pd.DataFrame, dataset_kind: DatasetKind):
+        sub_index = list(data.index)[:10]
+        return (data.loc[sub_index].fillna('nan') ==
+                self._get_dataset(dataset_kind).loc[sub_index].fillna('nan')).all().all()
+
+    def _get_dataset_kind(self, data: pd.DataFrame):
```
noamzbr, line 113: I have a slight concern that this constant computation of set intersection can become costly when the data is big, and we expect to do this a lot (every time we do predict on any dataset). 
Can we have something that will enable us not to do all this in case we get the whole dataset to predict on? I expect that a bit % of operations will be "just give me the predictions for the whole train/test".

deepchecks/tabular/context.py
```diff
@@ -30,16 +33,118 @@
...
+        Array of the model prediction probabilities over the test dataset.
+    """
+
+    train: pd.DataFrame
+    test: pd.DataFrame
+    ind_map: t.Dict[str, t.Dict[t.Any, int]]
+    y_pred: t.Dict[str, np.ndarray]
+    y_proba: t.Dict[str, np.ndarray]
+
+    def __init__(self,
```
noamzbr, line 61: Think that there should be input validation. Are the inputs ndarrays? are they of the right shape? are the proba values in [0, 1] and sum to 1 on each row? Does the n_cols in proba match the class idx we see in pred? Look in the vision classification data class validations for inspiration.

deepchecks/tabular/context.py
JKL98ISR: That is why it is a tiny sample

deepchecks/tabular/context.py
JKL98ISR: We validate in vision before the user does stuff for our format. Here he just sends ys what ge has

deepchecks/tabular/context.py
JKL98ISR: Can you offer another name?

deepchecks/tabular/context.py
noamzbr: But he still can send the wrong thing by mistake, and we should try to warn him if that the case (better than failing with  a cryptic message somewhere inside).
Can perhaps check how sklearn validates this stuff.

deepchecks/tabular/context.py
noamzbr: "Or pass the model object instead of the predictions"?

deepchecks/tabular/context.py
noamzbr: What is a tiny sample? Can you elaborate?

deepchecks/tabular/context.py
```diff
@@ -30,16 +33,118 @@
 ]
 
 
+class _DummyModel:
+    """Contains all the data + properties the user has passed to a check/suite, and validates it seamlessly.
```
matanper, line 37: Seems unrelated

deepchecks/tabular/context.py
```diff
@@ -30,16 +33,118 @@
...
+            self.predict_proba = self._predict_proba
+
+    def _get_dataset(self, dataset_kind: DatasetKind):
+        if dataset_kind == DatasetKind.TRAIN:
+            return self.train
+        if dataset_kind == DatasetKind.TEST:
+            return self.test
+        raise DeepchecksValueError(f'Unexpected dataset kind {dataset_kind}')
+
+    def _validate_data(self, data: pd.DataFrame, dataset_kind: DatasetKind):
```
matanper, line 108: what it does and why? (explain in the code)

deepchecks/tabular/context.py
```diff
@@ -30,16 +33,118 @@
...
+        sub_index = list(data.index)[:10]
+        return (data.loc[sub_index].fillna('nan') ==
+                self._get_dataset(dataset_kind).loc[sub_index].fillna('nan')).all().all()
+
+    def _get_dataset_kind(self, data: pd.DataFrame):
+        assert isinstance(data, pd.DataFrame)
+        if set(data.index).issubset(list(self.ind_map[DatasetKind.TRAIN].keys())):
```
matanper, line 115: why do it on all the indices? if they are exclusive you can just check the first index

deepchecks/tabular/context.py
```diff
@@ -30,16 +33,118 @@
...
+
+        if train is not None:
+            self.train = train.features_columns
+            for i, ind in enumerate(self.train.index):
+                self.ind_map[DatasetKind.TRAIN][ind] = i
+
+        if test is not None:
+            self.test = test.features_columns
+            for i, ind in enumerate(self.test.index):
+                self.ind_map[DatasetKind.TEST][ind] = i
```
matanper, line 87: why do you need to separate the map to train & test if the indices are mutually exclusive?

deepchecks/tabular/context.py
```diff
@@ -102,7 +219,11 @@ def __init__(self,
         if features_importance is not None:
             if not isinstance(features_importance, pd.Series):
                 raise DeepchecksValueError('features_importance must be a pandas Series')
-
+        # we do it after validation because the model we are creating isn't valid (we are always valid just not here)
```
matanper, line 222: >  we are creating isn't valid 

what's not valid about him?
>  (we are always valid just not here)

not sure what you mean

JKL98ISR: dummy model isn't of type base estimator so it will fail with our validation
"(we are always valid just not here)" this one I tried to be funny

deepchecks/tabular/context.py
JKL98ISR: what if they are not? what if the common index is at the middle?

deepchecks/tabular/context.py
JKL98ISR: because I need to know which indices belong to where to pick from the right prediction

deepchecks/tabular/context.py
JKL98ISR: ok it isn't needed

deepchecks/tabular/context.py
matanper: but there is no common index you validate it at the ctor

deepchecks/tabular/context.py
matanper: you are anyway mapping the indices why not just have a map of index to prediction?

matanper:
```
for index, prediction in zip(test.index, test_predictions):
   self.pred_map[index] = prediction
```

deepchecks/tabular/context.py
JKL98ISR: ctor?

deepchecks/tabular/context.py
JKL98ISR: explained

deepchecks/tabular/context.py
JKL98ISR: but then I can't validate the data

JKL98ISR: also it will make the prediction to be of type dataframe instead of numpy array (or like need to be aggregated)

matanper: what does it mean validate the data

matanper: and dataframe and numpy are very easy to convert between each other

JKL98ISR: it is but it makes the data preparation + returning harder
and i can't validate if the data is from the right dataset if I don't know which dataset it is

deepchecks/tabular/context.py
ItayGabbay: constructor

deepchecks/tabular/context.py
matanper: why do we need to validate that? can we just say that if the id is the same than the row is the same?

deepchecks/tabular/context.py
JKL98ISR: but it might be new data, I want to make sure

deepchecks/tabular/context.py
JKL98ISR: "row is the same" is the validation

JKL98ISR: row as in full row not index

deepchecks/tabular/context.py
```diff
@@ -27,16 +31,108 @@
...
+                predictions.append(pd.Series(dict(zip(train.data.index, y_pred_train))))
+                if y_proba_train is not None:
+                    ensure_predictions_proba(y_proba_train, y_pred_train)
+                    probas.append(pd.DataFrame(data=y_proba_train, index=train.data.index))
+
+        if test is not None:
+            if y_pred_test is not None:
+                ensure_predictions_shape(y_pred_test, test.data)
+                predictions.append(pd.Series(dict(zip(test.data.index, y_pred_test))))
+                if y_proba_train is not None:
```
matanper, line 93: typo: y_proba_test

deepchecks/tabular/context.py
noamzbr: I tend to agree with @matanper, matching 10 samples may not be fool-proof enough.

PR #1496, opened 2022-05-23
[Refactoring] resolved some `TODOs` items from the source code

yromanyshyn:
resolved some `TODOs` from the source code

list of changes:
- added context manager for the wandb run creation
- added comments for the collected check results indexes in the `SuiteResult` constructor
- added deprecation warning for the `with_display` parameter (`CheckFailure.to_json`, `SuiteResult.to_json`)


ItayGabbay:
@JKL98ISR can you review?

deepchecks/core/check_result.py
```diff
@@ -632,13 +624,19 @@ def to_json(self, with_display: bool = True):
...
 
         Returns
         -------
         str
         """
-        # TODO: not sure if the `with_display` parameter is needed
-        # add deprecation warning if it is not needed
+        if with_display is not None:
+            warnings.warn(
+                '"with_display" parameter is deprecated and does not have any effect '
```
JKL98ISR, line 636: it will not be removed, it is just that check failure doesnt have a display.
the function needs to be the same or else the suite.to_json would fail

deepchecks/core/suite.py
```diff
@@ -224,37 +231,42 @@ def to_widget(
...
+            (the parameter is deprecated and does not have any effect since version 0.6.4,
+            it will be removed in future versions)
 
         Returns
         -------
         str
         """
-        # TODO: not sure if the `with_display` parameter is needed
-        # add deprecation warning if it is not needed
+        if with_display is not None:
```
JKL98ISR, line 248: it should have an effect, it used to have.
probably haven't noticed when I reviewed your refactor

yromanyshyn: fixed

yromanyshyn:
@JKL98ISR 

**note regarding `to_wandb` method**

I marked `dedicated_run` parameter as deprecated because there is no sense in having it.
It looks like `wandb.run` is a singleton and you cannot have two or more instances of it simultaneously.

example
```python
import wandb

r1 = wandb.init(project='test1')
r2 = wandb.init(project='test2')

print(wandb.run is r2)  # True
r2.log({'accuraccy': 0.8})
r2.finish()

print(wandb.run is None)  # True
r1.log({'accuraccy': 0.8})  # << will raise "Exception: The wandb backend process has shutdown"    
```

so, now `to_wandb` method works in the next way:
if `wandb.run` is not `None`, use it otherwise initialize new instance and close it at the end

deepchecks/core/check_result.py
JKL98ISR: didn't you just re-added it?

JKL98ISR: oh not for check fail

deepchecks/utils/wandb_utils.py
```diff
@@ -16,6 +16,9 @@
 __all__ = ['wandb_run']
 
 
+WANDB_INSTALLATION_CMD = 'pip install "wandb>=0.12.15" "protobuf>=3.12.0,<4.0dev"'
```
JKL98ISR, line 19: why we need a specific version? latest is enough

yromanyshyn: `"wandb>=0.12.15"` - this I copied from the dev-requirements
 "protobuf>=3.12.0,<4.0dev" - this is needed because otherwise wandb will raise a `ImportError`

JKL98ISR: no need for the user to have a specific wandb, would also look weird.
is the thing with protobuf a known issue? seems weird it is needed

yromanyshyn: > no need for the user to have a specific wandb, would also look weird.

it is not a specific wandb 
it is any wandb that is greater than 0.12

> is the thing with protobuf a known issue? seems weird it is needed

this problem appeared today in the morning CI/CD checks were failing
because somewhy `protobuf` greater than 4.0 were installed

yromanyshyn: hm, indeed if deepchecks is installed from the python package index this problem with protobuf is not present, probably something is in our `dev-requirements` that installs protobuf>4.0

fixed

JKL98ISR (approved):

noamzbr (approved):

PR #1520, opened 2022-05-26
Update dataset exceptions

matanper:


deepchecks/tabular/checks/train_test_validation/identifier_leakage.py
```diff
@@ -18,6 +18,7 @@
 from deepchecks.core import CheckResult, ConditionCategory, ConditionResult
 from deepchecks.core.errors import DatasetValidationError
 from deepchecks.tabular import Context, SingleDatasetCheck
+from deepchecks.tabular.dataset import _get_dataset_docs_tag
```
ItayGabbay, line 21: is it ok to import privates?

ItayGabbay (approved):

PR #1522, opened 2022-05-27
[issue-1448] `IsSingleValue` check - ignore nan

### LINKED ISSUE #1448: [FEAT] Enable ignoring Null values in the Is Single Value check
Currently, if the data has both a single value and NaNs in the same column, this column will be considered to have 2 unique values.

Add a Boolean parameter telling the check to disregard NaN's

**Additional context**
Issue raised by @galsr in the [Deepchecks Community](https://join.slack.com/t/deepcheckscommunity/shared_invite/zt-1973c7twx-Ytb7_yJefTT4PmnMZj2p7A) support channel.

yromanyshyn:
resolves #1448


deepchecks/tabular/checks/data_integrity/is_single_value.py
```diff
@@ -31,17 +31,21 @@ class IsSingleValue(SingleDatasetCheck):
     ignore_columns : Union[Hashable, List[Hashable]] , default: None
         Columns to ignore, if none given checks based
         on columns variable.
+    ignore_nan : bool, default True
+        whether to ignore NaN values in a column or not
```
noamzbr, line 35:
```suggestion
        Whether to ignore NaN values in a column when counting the number of unique values.
```

noamzbr (commented):
Please also add an example of this option to plot_is_single_value.py

matanper (changes requested):
notice that line `uniques = df.loc[:, cols_with_single].head(1)`
it won't work currently, since it just takes the first value (which might be null)

deepchecks/tabular/checks/data_integrity/is_single_value.py
```diff
@@ -31,17 +31,21 @@ class IsSingleValue(SingleDatasetCheck):
...
     """
 
     def __init__(
         self,
         columns: Union[Hashable, List[Hashable], None] = None,
         ignore_columns: Union[Hashable, List[Hashable], None] = None,
+        ignore_nan: bool = True,
```
matanper, line 42: why the default is True? until now it was False

yromanyshyn: > By default: ignore Nan

Itay asked for this

yromanyshyn:
@matanper 

> notice that line `uniques = df.loc[:, cols_with_single].head(1)`
it won't work currently, since it just takes the first value (which might be null)

nice catch, thanks!
fixed

yromanyshyn:
@noamzbr 

> Please also add an example of this option to plot_is_single_value.py

added

deepchecks/tabular/checks/data_integrity/is_single_value.py
```diff
@@ -60,14 +66,17 @@ def run_logic(self, context: Context, dataset_type: str = 'train') -> CheckResul
...
+        num_unique_per_col = df.nunique(dropna=self.ignore_nan)
         is_single_unique_value = (num_unique_per_col == 1)
 
         if is_single_unique_value.any():
             # get names of columns with one unique value
             # pylint: disable=unsubscriptable-object
             cols_with_single = is_single_unique_value[is_single_unique_value].index.to_list()
-            uniques = df.loc[:, cols_with_single].head(1)
+            uniques = pd.DataFrame({
+                column_name: [column.sort_values(kind='mergesort').values[0]]
```
matanper, line 77: isn't sort redundant?
```suggestion
                column_name: [column.dropna().values[0]]
```

yromanyshyn: note: 
I revert the change 
it will fail if to set `ignore_nan=False` and pass a data frame with only nans

matanper, line 77: so sorting will always put nulls at the bottom?

yromanyshyn: yes

matanper (approved):

ItayGabbay (approved):

PR #1525, opened 2022-05-29
Static vision model

### LINKED ISSUE #1364: [FEAT] support a df of predictions instead of providing a model
Currently, our checks and suites require a compatible model to work with. The 2 main functions that are used are the `predict` function that returns the predictions, and the inferring of `feature_importance`. It limits the package to work with a limited number of models and frameworks.

Instead of specifying a model, let the user pass a data frame full of predictions (and probably another parameter that specifies the feature_importances). This way we will be model and framework agnostic.

JKL98ISR:
closes #1364
also fixed batch length bug


deepchecks/vision/batch_wrapper.py
```diff
@@ -49,15 +50,28 @@ def labels(self):
...
     def predictions(self):
         """Return predictions for the batch, formatted in deepchecks format."""
         if self._predictions is None:
             dataset = self._context.get_data_by_kind(self._dataset_kind)
-            # Calling model will raise error if model was not given
-            model = self._context.model
-            self._context.assert_predictions_valid(self._dataset_kind)
-            self._predictions = dataset.infer_on_batch(self._batch, model, self._context.device)
+            if self._context.static_predictions is not None:
+                self._context.assert_predictions_valid(self._dataset_kind)
```
matanper, line 68: why do you have the same `self._context.assert_predictions_valid(self._dataset_kind)` twice?

deepchecks/vision/batch_wrapper.py
```diff
@@ -75,13 +89,19 @@ def __getitem__(self, index: int):
 
     def __len__(self):
         """Return length of batch."""
-        return len(self._batch)
+        dataset = self._context.get_data_by_kind(self._dataset_kind)
```
matanper, line 92: I'm not sure this is the best way to get batch length, maybe we should use `len(batch_to_images(batch))` 
@noamzbr WDYT?

matanper (approved):

deepchecks/vision/batch_wrapper.py
JKL98ISR: fixed comment bellow

deepchecks/vision/batch_wrapper.py
JKL98ISR: batch_to_images not always implemented.
fixed case of sampled data

ItayGabbay (approved):

deepchecks/vision/batch_wrapper.py
```diff
@@ -75,13 +90,19 @@ def __getitem__(self, index: int):
...
```
noamzbr, line 93: @JKL98ISR Can you explain the logic of this code? Possible also in comments. 
Seems to me like this returns the number of batches in the dataset.

PR #1528, opened 2022-05-29
Mixed data types examples

### LINKED ISSUE #1454: [FEAT] MixedDataTypes to display a few examples of each type
Current display:
image

Desired - to see a few unique examples from each type. currently have to inspect column in order to do it, e.g.:
image
would want the check to display it by itself

JKL98ISR:
closes #1454 

Screen Shot 2022-05-29 at 15 35 54


deepchecks/tabular/checks/data_integrity/mixed_data_types.py
```diff
@@ -79,7 +79,13 @@ def run_logic(self, context: Context, dataset_type: str = 'train') -> CheckResul
             result_dict[column_name] = mix
             if mix:
                 # Format percents for display
-                display_dict[column_name] = {k: format_percent(v) for k, v in mix.items()}
+                formated_mix = {}
+                formated_mix['strings'] = format_percent(mix['strings'])
```
matanper, line 83: If this is for display, than make it prettier, 'Strings', 'Numbers'... also drop the word unique, just "Strings examples"

matanper (approved):

PR #1530, opened 2022-05-29
Enable validation of empty predictions

noamzbr:
Find a non-empty label/prediction to validate, and finish validation with no error if non exist.

### DIFF
--- a/deepchecks/vision/detection_data.py
+++ b/deepchecks/vision/detection_data.py
@@ -144,8 +144,7 @@ class DetectionData(VisionData):
             raise ValidationError('Check requires object detection label to be a list with an entry for each '
                                   'sample')
         if len(labels) == 0:
-            raise ValidationError('Check requires object detection label to be a non-empty list, with one entry for '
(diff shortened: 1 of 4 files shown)


ItayGabbay (approved):

deepchecks/vision/detection_data.py
```diff
@@ -180,21 +187,28 @@ def validate_infered_batch_predictions(batch_predictions):
             raise ValidationError('Check requires detection predictions to be a list with an entry for each'
                                   ' sample')
         if len(batch_predictions) == 0:
-            raise ValidationError('Check requires detection predictions to be a non-empty list')
+            raise ValidationError('Check requires detection predictions to be a non-empty list, with one entry for '
```
Nadav-Barak, line 190: IMO "one entry per each sample" can be more confusing then helpful

Nadav-Barak (approved):
Good catch

JKL98ISR (approved):

shir22 (approved):

PR #1549, opened 2022-05-30
[Issue-1443] tabular Dataset repr 

### LINKED ISSUE #1443: [FEAT] Add printable representation of deepchecks Dataset
When just printing a dataset object, there is not output that gives any information about it's content
image

Design a default display (for jupyter notebook and for python console) that is similar to that of a pandas df, but different enough so people won't get confused and also contains info about column roles.

yromanyshyn:
resolves #1443 

screenshots:

-- plain --
image

-- plain (via tabulate library)--
image

-- html -- 
image


deepchecks/tabular/dataset.py
```diff
@@ -1061,6 +1065,172 @@ def datasets_share_date(cls, *datasets: 'Dataset') -> bool:
...
+    ]:
+        data = self.data
+        datetime_name = self.datetime_name
+        index_name = self.index_name
+        label_name = self.label_name
+        features = self.features
+        categorical_features = self.cat_features
+        numerical_features = self.numerical_features
+
+        columns = features[::]
```
matanper, line 1082: why columns = features? why not just the dataframe columns?

deepchecks/tabular/dataset.py
```diff
@@ -1061,6 +1065,172 @@ def datasets_share_date(cls, *datasets: 'Dataset') -> bool:
...
+
+        unamed_index_levels = {
+            level
+            for level, name in enumerate(data.index.names)
+            if name is None
+        }
+
+        if index_name in data.columns:
+            columns = [index_name, *columns]
+
```
matanper, line 1099: don't have blank lines between if/else

deepchecks/tabular/dataset.py
```diff
@@ -1061,6 +1065,172 @@ def datasets_share_date(cls, *datasets: 'Dataset') -> bool:
...
+                infer_dtype(data[feature_name], skipna=True),
+                kind
+            ])
+
+        columns_info = pd.DataFrame(
+            data=columns_info,
+            columns=['Column', 'DType', 'Kind'],
+        )
+
+        return columns, columns_info, data.loc[:, columns]
```
matanper, line 1163: why do you need `data.loc[:, columns]`? which columns will be filtered?

deepchecks/tabular/dataset.py
```diff
@@ -1061,6 +1065,172 @@ def datasets_share_date(cls, *datasets: 'Dataset') -> bool:
 
         return True
 
+    @property
+    def _dataset_description(self) -> t.Tuple[
```
matanper, line 1069: I'd pass here `max_rows_to_show` and in the start filter the data, since if it is a few millions in length, all dataframe operations here are expensive and un-necessary

deepchecks/tabular/dataset.py
```diff
@@ -1061,6 +1065,172 @@ def datasets_share_date(cls, *datasets: 'Dataset') -> bool:
...
+        t.List[t.Union[str, int]],
+        pd.DataFrame,
+        pd.DataFrame
+    ]:
+        data = self.data
+        datetime_name = self.datetime_name
+        index_name = self.index_name
```
matanper, line 1076: index_name can't be None?

deepchecks/tabular/dataset.py
```diff
@@ -1061,6 +1065,172 @@ def datasets_share_date(cls, *datasets: 'Dataset') -> bool:
...
+
+        else:
+            data = data.reset_index()
+            index_name = 'index'
+            columns = [index_name, *columns]
+
+        is_datetime_name_str = isinstance(datetime_name, str)
+        is_datetime_name_int = isinstance(index_name, int)
+
+        if datetime_name in columns:
```
matanper, line 1117: what's the difference between `datetime_name in columns` and `datetime_name in data.columns`?

matanper (commented):
Also: do we support multi-index handling? you spent a lot of code on it

deepchecks/tabular/dataset.py
yromanyshyn: sorry, did not get your question/
but, it can be None

image

deepchecks/tabular/dataset.py
yromanyshyn: as much as I know, pandas does not touch data during slicing/column selection
it operates only over indexes therefore it is so effective

but, yup, it is better to change this
will fix

deepchecks/tabular/dataset.py
yromanyshyn: 1. I am doing `df.reset_index` above and it will add new columns
2. in order to show columns in desired order

in general, this method task is to prepare a dataframe in a form
```
| index | datetime? | *other columns | label? |
------------------------------------------------
```

deepchecks/tabular/dataset.py
yromanyshyn: why?
I think code becomes more readable when there is one line between multiline `if/elif/else` branches

deepchecks/tabular/dataset.py
yromanyshyn: what if provided dataframe contains much more columns than there are features that are interesting for the user

image

deepchecks/tabular/dataset.py
matanper: I think it's all part of the same code block, and the blank lines separates it and make it harder to read for me

yromanyshyn:
@matanper 

> Also: do we support multi-index handling? you spent a lot of code on it

yes, `index_name` could be an integer indicating a level of a multiindex

deepchecks/tabular/dataset.py
matanper: > pandas does not touch data during slicing/column selection

right but when you do reset_index which adds new columns, it does copy the data

deepchecks/tabular/dataset.py
matanper: I'm not sure how you handle the case which index_name is none

deepchecks/tabular/dataset.py
yromanyshyn: -- `datetime_name in columns` --

user passed `datetime_name=<something> and set_datetime_from_dataframe_index=False`
therefore datetime column could be within `columns`

-- `datetime_name in data.columns` --

user passed `datetime_name=<something> and set_datetime_from_dataframe_index=True`
datetime column is an index or one of the levels of the `multiindex` but there are possibility that previously `data.reset_index` was called therefore datetime column now could be within `data.column`

deepchecks/tabular/dataset.py
yromanyshyn: I just use dataframe index
hm/ that is probably a wrong way/
probably better to remove this `else` branch?

```python
# def __dataset_description(): lines:1109:1113
    else:
        data = data.reset_index()
        index_name = 'index'
        columns = [index_name, *columns]

```

deepchecks/tabular/dataset.py
yromanyshyn: yes
`reset_index`, will produce a bunch of new columns, in case, if `multiindex` was used
and we need to get rid of them

deepchecks/tabular/dataset.py
matanper: why not just always do `reset_index()` right on the start? anyway you later filter it by columns

yromanyshyn: I rewrote that method, this is not actual anymore

deepchecks/tabular/dataset.py
yromanyshyn: fixed

deepchecks/tabular/dataset.py
yromanyshyn: fixed

deepchecks/tabular/dataset.py
```diff
@@ -1061,6 +1066,158 @@ def datasets_share_date(cls, *datasets: 'Dataset') -> bool:
 
         return True
 
+    def _get_dataset_description(
+        self,
+        max_rows_to_show: int = 25,
```
noamzbr, line 1071: would use 10

deepchecks/tabular/dataset.py
```diff
@@ -1061,6 +1066,158 @@ def datasets_share_date(cls, *datasets: 'Dataset') -> bool:
...
+            datetime_column_info = [datetime_name, infer_dtype(datetime_column, skipna=True), 'Datetime']
+            dataset_columns.append(datetime_name)
+            dataset_columns_info.append(datetime_column_info)
+        else:
+            datetime_name = None
+
+        label_name = label_column.name if label_column is not None else None
+
+        for feature_name in t.cast(t.List[str], features):
+            if feature_name in (index_name, datetime_name, label_name):
```
noamzbr, line 1121: I think it shouldn't be possible for a feature to be an index, date or label (we're supposed to check for it in Dataset construction)

deepchecks/tabular/dataset.py
```diff
@@ -1061,6 +1066,158 @@ def datasets_share_date(cls, *datasets: 'Dataset') -> bool:
...
+                continue
+
+            if feature_name in categorical_features:
+                kind = 'Categorical'
+            elif feature_name in numerical_features:
+                kind = 'Numerical'
+            else:
+                kind = 'Unknown'
```
noamzbr, line 1129:
```suggestion
                kind = 'Other Feature'
```
A feature that is non of the above is

deepchecks/tabular/dataset.py
```diff
@@ -1061,6 +1066,158 @@ def datasets_share_date(cls, *datasets: 'Dataset') -> bool:
...
+        label_name = label_column.name if label_column is not None else None
+
+        for feature_name in t.cast(t.List[str], features):
+            if feature_name in (index_name, datetime_name, label_name):
+                continue
+
+            if feature_name in categorical_features:
+                kind = 'Categorical'
+            elif feature_name in numerical_features:
+                kind = 'Numerical'
```
noamzbr, line 1127:
```suggestion
                kind = 'Numerical Feature'
```

deepchecks/tabular/dataset.py
noamzbr: If it's a topic of "we're showing at most 7 columns", I agree that we should first show the label, index, datetime, features, and only last the other columns. 
But if the dataframe has a total of 5 columns (e.g. 1 label, 3 features, 1 other) then we should show the other.

deepchecks/tabular/dataset.py
yromanyshyn: the constructor will raise an error in this case

```python
        if self._label_name in self.features:
            raise DeepchecksValueError(f'label column {self._label_name} can not be a feature column')

        if self._datetime_name in self.features:
            raise DeepchecksValueError(f'datetime column {self._datetime_name} can not be a feature column')

        if self._index_name in self.features:
            raise DeepchecksValueError(f'index column {self._index_name} can not be a feature column')
```

deepchecks/tabular/dataset.py
```diff
@@ -1061,6 +1065,172 @@ def datasets_share_date(cls, *datasets: 'Dataset') -> bool:
...
+        for feature_name in t.cast(t.List[str], columns):
+            if feature_name == index_name:
+                kind = 'Index'
+            elif feature_name == label_name:
+                kind = t.cast(str, self.label_type).replace('_', ' ').capitalize()
+            elif feature_name in categorical_features:
+                kind = 'Categorical'
+            elif feature_name in numerical_features:
+                kind = 'Numerical'
+            elif feature_name == datetime_name:
```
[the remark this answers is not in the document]
yromanyshyn: just `Text` or `Textual Feature`?

deepchecks/tabular/dataset.py
```diff
@@ -1061,6 +1066,158 @@ def datasets_share_date(cls, *datasets: 'Dataset') -> bool:
...
+            if len(columns_to_show) != len(dataset_columns):
+                columns_to_show = [*columns_to_show[:-1], '...', columns_to_show[-1]]
+                data['...'] = '...'
+
+        return dataset_columns_info, data.loc[:, columns_to_show].reset_index(drop=True)
+
+    def __repr__(
+        self,
+        max_columns_to_show: int = 7,
+        max_rows_to_show: int = 25,
```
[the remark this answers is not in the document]
yromanyshyn: fixed

deepchecks/tabular/dataset.py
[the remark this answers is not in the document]
ItayGabbay: Text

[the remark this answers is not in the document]
noamzbr: I've changed my original suggestion, think we should stick with "Other Feature". @ItayGabbay ?

deepchecks/tabular/dataset.py
ItayGabbay: Ok so as the expected behavior we will do the following:
First display Index, Datetime, Label, then features, and then if there is some space for more columns we will display other columns which are not features,

noamzbr (approved):

PR #1559, opened 2022-06-01
Some performance reduce-mixin fixes

JKL98ISR:
related to #1550

### DIFF
--- a/deepchecks/tabular/checks/model_evaluation/performance_report.py
+++ b/deepchecks/tabular/checks/model_evaluation/performance_report.py
@@ -9,7 +9,7 @@
 # ----------------------------------------------------------------------------
 #
 """Module containing performance report check."""
-from typing import Callable, Dict, TypeVar, cast
+from typing import Callable, Dict, TypeVar, Union, cast
 
 import pandas as pd
 import plotly.express as px
@@ -36,7 +36,9 @@ class PerformanceReport(TrainTestCheck, ReduceMixin):
     alternative_scorers : Dict[str, Callable], default: None
         An optional dictionary of scorer name to scorer functions.
         If none given, using default scorers
-
+    reduce: Union[Callable, str], default: 'mean'
+        An optional argument only used for the reduce_output function when using
+        non-average scorers.
     Notes
     -----
     Scorers are a convention of sklearn to evaluate a model.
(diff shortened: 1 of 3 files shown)


ItayGabbay (dismissed):

deepchecks/tabular/checks/model_evaluation/performance_report.py
```diff
@@ -150,6 +150,7 @@ def reduce_output(self, check_result: CheckResult) -> Dict[str, float]:
         """Return the values of the metrics for the test dataset in {metric: value} format."""
         df = check_result.value
         df = df[df['Dataset'] == DatasetKind.TEST.value]
+        df = df.groupby('Metric').mean().reset_index()
```
matanper, line 153: I don't think this makes sense to do a mean

JKL98ISR: we can't start making assumptions about params all need to be able to be passed as options...
half of our params coming from unknown sources like the alternative metrics

JKL98ISR: the default sklearn metrics behavior for metrics if you want a single value is mean.
"Calculate metrics for each label, and find their unweighted mean. This does not take label imbalance into account."

matanper: after talking with @noamzbr the main points are:
1. If someone wants an averaged metric (not per-class), then he should use averaged scorers.
2. Using per-class metrics is legit for other use cases, like monitoring the worst-performing class. Therefore in order to make it possible we can add `per_class_reduce_fn` which accepts a string of pre-defined values (min, max, median, std, etc...), and then in the per-class case, we apply this reduce

JKL98ISR: we need the reduced_output to work in every case of the check
also i commented on why the reduce function should be as the check sol did

JKL98ISR: how is the check's default behavior misconfigured?

> where? you posted a sentence from unknown source and context
what do you mean? the user should be able to give a reduce function as it is in sol's check

deepchecks/tabular/checks/model_evaluation/performance_report.py
```diff
@@ -150,6 +156,7 @@ def reduce_output(self, check_result: CheckResult) -> Dict[str, float]:
         """Return the values of the metrics for the test dataset in {metric: value} format."""
         df = check_result.value
         df = df[df['Dataset'] == DatasetKind.TEST.value]
+        df = df.groupby('Metric').aggregate(self.reduce).reset_index()
```
matanper, line 159: the aggregate function accepts strings and functions?

JKL98ISR: yas, even tested

matanper (approved):

PR #1568, opened 2022-06-02
Add display in python gui

matanper:
the libraries used to show html gui are quite heavy size-wise therefore does not add them to requirements


ItayGabbay (approved):

yromanyshyn:
@matanper maybe let's just save the html and open it in the default browser?
then it will not need additional dependencies like Qt

matanper:
> @matanper maybe let's just save the html and open it in the default browser? then it will not need additional dependencies like Qt

It was one of the options, but we don't want to open a browser

yromanyshyn (approved):

TheSolY (approved):
Works smoothly on Pycharm Professional

Issue #1453, opened 2022-05-17
[FEAT] Reconsider special character treatment when column is all special characters


shir22:
**Describe the bug**
Makes other checks (except Special Characters) fail, example:
"+" and "-" strings are considered as nulls - both in MixedNulls Check and in StringMismatch Check

**To Reproduce**
Run:
```
import pandas as pd
from deepchecks.tabular import Dataset
from deepchecks.tabular.suites import single_dataset_integrity

col_names = ["A" + str(i) for i in range(1,17)]
cat_features = ['A1', 'A4', 'A5', 'A6', 'A7', 'A9', 'A10', 'A12', 'A13', 'A16']
df = pd.read_csv('https://archive.ics.uci.edu/ml/machine-learning-databases/credit-screening/crx.data', sep=",", names=col_names) 
ds = Dataset(df, cat_features=cat_features)
si.run(ds)
```

Will result in:
Conditions Summary table:
![image](https://user-images.githubusercontent.com/33841818/168740389-1836f3e2-10a9-4586-b602-badb9a7814de.png)

Check Outputs:
![image](https://user-images.githubusercontent.com/33841818/168740474-8e1ad9dc-0f21-4f8a-811b-40f2ee2fa425.png)
![image](https://user-images.githubusercontent.com/33841818/168740492-d229efc8-3c4b-4d81-9e31-6b1bf082f1c8.png)

matanper:
If you also suggested a solution here I missed it. Both checks use the `string_baseform` function, which transforms '+' and '-' into an empty string. We can change the checks/the function to leave the string as-is in case the transform results in empty string

PR #1581, opened 2022-06-06
base form operates only if at least one alphanumeric char exist

Nadav-Barak:
Solves https://github.com/deepchecks/deepchecks/issues/1453.

### DIFF
--- a/deepchecks/utils/strings.py
+++ b/deepchecks/utils/strings.py
@@ -244,7 +244,10 @@ def get_random_string(n: int = 5):
 
 
 def string_baseform(string: Hashable) -> Hashable:
-    """Remove special characters from given string, leaving only a-z, A-Z, 0-9 characters.
+    """Normalize the string input to uniform form if possible.
+
+    If input is a string containing alphanumeric characters, removes all non-alphanumeric characters and convert
+    characters to lower form.
 
     Parameters
     ----------
@@ -253,11 +256,15 @@ def string_baseform(string: Hashable) -> Hashable:
     Returns
     -------
     str
-        string without special characters
+        original input if condition is not met or lower form alphanumeric characters of input.
     """
     if not isinstance(string, str):
         return string
-    return string.translate(DEL_MAP).lower()
+    lower_alphanumeric_form = string.translate(DEL_MAP).lower()
+    if len(lower_alphanumeric_form) > 0:
+        return lower_alphanumeric_form
+    else:
+        return string
 
 
 def is_string_column(column: pd.Series) -> bool:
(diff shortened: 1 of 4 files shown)


deepchecks/utils/strings.py
```diff
@@ -253,11 +256,15 @@ def string_baseform(string: Hashable) -> Hashable:
...
     -------
     str
-        string without special characters
+        original input if condition is not met or lower form alphanumeric characters of input.
     """
     if not isinstance(string, str):
         return string
-    return string.translate(DEL_MAP).lower()
+    lower_alphanumeric_form = string.translate(DEL_MAP).lower()
+    if len(lower_alphanumeric_form) > 0:
```
matanper, line 264: maybe we want this to be optional? does that fit in every check we use this?

tests/base/feature_importance_utils_test.py
```diff
@@ -55,7 +55,7 @@ def test_pipeline(iris_split_dataset_and_model_single_feature):
     _, test_ds, clf = iris_split_dataset_and_model_single_feature
     feature_importances, fi_type = calculate_feature_importance(clf, test_ds)
     print(feature_importances)
-    assert_that(feature_importances['sepal length (cm)'], equal_to(1))
+    assert_that(feature_importances['sepal length (cm)'], equal_to(1))  # pylint: disable=e1136
```
JKL98ISR, line 58: why we need that: # pylint: disable=e1136?

Nadav-Barak: Because pylint sucks and specifically they have a bug regarding this

tests/base/feature_importance_utils_test.py
```diff
@@ -55,7 +55,7 @@ def test_pipeline(iris_split_dataset_and_model_single_feature):
     _, test_ds, clf = iris_split_dataset_and_model_single_feature
     feature_importances, fi_type = calculate_feature_importance(clf, test_ds)
     print(feature_importances)
```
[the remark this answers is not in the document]
Nadav-Barak: :O
Removed

JKL98ISR:
1. why do we use this function in mixed_nulls?
2. look at the function `_get_special_samples` in string mismatch, pretty sure it will break it. there are checking if `len(string_baseform(sample))` is 0

deepchecks/utils/strings.py
Nadav-Barak: Will add a boolean

tests/tabular/checks/integrity/mixed_nulls_test.py
```diff
@@ -131,12 +131,12 @@ def test_mix_value_columns():
 
 def test_single_column_nulls_with_special_characters():
     # Arrange
-    data = {'col1': ['', '#@$', 'Nan!', '#nan', '<NaN>']}
+    data = {'col1': ['', 'value', 'Nan!', '#nan', '<NaN>']}
```
JKL98ISR, line 134: why not test the special char case?

PR #1594, opened 2022-06-09
Add passed function to suite result

### LINKED ISSUE #1589: [BUG] [DOC] Dataset.numerical_features typo
image

Should probably be list of numerical

### LINKED ISSUE #1591: [BUG] Missing pytest function from doc
**Describe the bug**
In the docs there is a section about "integration" with pytest. a SuiteResult should have " passed_conditions()" function to assert on. but none exist. Also, there is no clear way to find out in script (instead of a notebook) which tests  pass/fail/warn. and it feels like there should be.
best thing I manged to do was iterate over test_result.results ->[idx of test] -> category -> value
**To Reproduce**
Try to run a suite and then assert on the function from the docs

**Expected behavior**
The function that is mention in the docs should be real/ removed from the docs until such a time when it is ready

 - OS: linux
 - Python Version:3.7.1
 - Deepchecks Version:0.7.2

**Additional context**

matanper:
resolve #1591 
resolve #1589

### DIFF
--- a/deepchecks/core/check_result.py
+++ b/deepchecks/core/check_result.py
@@ -161,9 +161,9 @@ class CheckResult(BaseCheckResult):
         """Return if this check has display."""
         return bool(self.display)
 
-    def passed_conditions(self, fail_on_warning=True) -> bool:
+    def passed_conditions(self, fail_if_warning=True) -> bool:
         """Return if this check has no passing condition results."""
-        return all((r.is_pass(fail_on_warning) for r in self.conditions_results))
+        return all((r.is_pass(fail_if_warning) for r in self.conditions_results))
 
     @property
     def priority(self) -> int:
--- a/deepchecks/core/check_result.py
+++ b/deepchecks/core/check_result.py
@@ -49,7 +49,7 @@ if TYPE_CHECKING:
     from deepchecks.core.checks import BaseCheck
 
 
-__all__ = ['CheckResult', 'CheckFailure']
+__all__ = ['CheckResult', 'CheckFailure', 'BaseCheckResult']
 
 
 TDisplayCallable = Callable[[], None]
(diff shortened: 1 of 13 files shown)


JKL98ISR (approved):

deepchecks/core/check_result.py
```diff
@@ -161,9 +161,9 @@ def have_display(self) -> bool:
         """Return if this check has display."""
         return bool(self.display)
 
-    def passed_conditions(self) -> bool:
+    def passed_conditions(self, fail_on_warning=True) -> bool:
```
Nadav-Barak, line 164: IMO fail on warning should be default False

deepchecks/core/suite.py
```diff
@@ -290,19 +290,45 @@ def to_wandb(
...
         """
-        failures = {}
-        for res in self.results:
-            if isinstance(res, CheckFailure):
-                failures[res.header] = res
-        return failures
+        return self.select_results(self.failures)
+
+    def get_checks_not_passed(self, fail_on_warning=True) -> List[CheckResult]:
+        """Get all the check results that have not passing condition.
```
Nadav-Barak, line 304: Not sure about this devision, where do checks without conditions go?

Maybe do this in the same devision as the new suite output: pass, fail, didnt run, other(/ checks without conditions)

deepchecks/core/suite.py
```diff
@@ -290,19 +290,45 @@ def to_wandb(
...
+
+         Returns
+        -------
+        List[CheckResult]
+            All the check results in the suite that have failing conditions.
+        """
+        return [r for r in self.select_results(self.results_with_conditions)
+                if not r.passed_conditions(fail_on_warning)]
+
+    def passed(self, fail_on_warning=True) -> bool:
```
Nadav-Barak, line 319: Same, IMO default should be false

Nadav-Barak (commented):
See comment about check devision

deepchecks/core/check_result.py
matanper: @noamzbr and @shir22 thinks otherwise

deepchecks/core/suite.py
matanper: this division is just for the passed function, checks without condition are not relevant

shir22 (approved):

ItayGabbay (approved):

PR #1575, opened 2022-06-03
[Issue-1509] reimplemented `is_widgets_enabled` function

yromanyshyn:
resolves #1509 

NOTE: still testing, and writing tests

Issue #1509, opened 2022-05-25
[BUG] `deepchecks.utils.ipython.is_widgets_enabled` does not work properly 


yromanyshyn:
**Short summary**

regarding `deepchecks.utils.ipython.is_widgets_enabled` function

That is not the right way to verify whether widgets are enabled or not:
- there is always a possibility that a user had started the jupyter server with not default config path
- there are two extension types:
     + classical notebook extensions (managed by 'jupyter nbextension');
     + jupyterlab extensions (managed by 'jupyter labextension');

matanper:
Also in databricks it doesn't work properly:

![image](https://user-images.githubusercontent.com/9868530/170455661-9f98f5cf-6f7b-4f58-ab4a-bed0664b6ebc.png)

PR #1611, opened 2022-06-13
[suggestion] new suite/check result display api 

yromanyshyn:
this is an alternative solution to the issue #1509 (PR #1575)

as we saw there is no an easy way to determine whether ipywidgets use is possible because
- jupyter does not provide any api to obtain this kind of info
- all those environments/platforms (collab, databricks, kaggle, ...) are not unified and work differently

this PR suggests a more pragmatic solution: stop caring about that and provide a fallback method(s) for users

it adds the next methods to the check/suite result:

`show(as_widget: bool, **kwargs)`
displays result

`show_in_iframe(as_widget: bool, **kwargs)`
serializes result into independent html doc and displays within iframe

`show_in_window(as_widget: bool, **kwargs)`
serializes result into independent HTML doc and shows it in a seperate window

`show_not_interactive()`
displays result without ipywidgets and also transforms Plotly figures into images


matanper:
if the iframe always works when the user will want the non interactive option?

yromanyshyn:
@matanper 

> if the iframe always works when a user will want the non-interactive option?

I have tested iframe in sagemaker, kaggle, databricks, nbclassic, jupyterlab, vscode
it works and looks good

but I cannot say in what env with what restrictions/limitations user will use deepchecks library
so I think it is a good idea to provide this option

>we still want to avoid a situation the user doesn't see any display and doesn't know why, so I guess the default should be iframe

hm..., yup, why not
then `show` will use iframe by default and `show_in_iframe` is not needed anymore

deepchecks/core/display.py
```diff
@@ -0,0 +1,368 @@
...
+                    full_html=True,
+                    include_requirejs=True,
+                    include_plotlyjs=True,
+                    is_for_iframe_with_srcdoc=True,
+                    **kwargs
+                )),
+                raw=True
+            )
+
+    # def show_in_iframe(
```
matanper, line 121: why is that in comment?

deepchecks/core/display.py
```diff
@@ -0,0 +1,368 @@
...
+            web.setGeometry(0, 0, 1200, 1200)
+
+            html_out = io.StringIO()
+            result.save_as_html(html_out)
+            web.setHtml(html_out.getvalue())
+            web.show()
+
+            sys.exit(app.exec_())
+        except BaseException:  # pylint: disable=broad-except
+            print(
```
JKL98ISR, line 244: use get_logger function

deepchecks/core/serialization/check_result/html.py
```diff
@@ -255,8 +297,18 @@ def handle_callable(cls, item, index, **kwargs) -> str:
...
+    def handle_figure(
+        cls,
+        item: BaseFigure,
+        index: int,
+        plotly_to_image: bool = False,
+        **kwargs
+    ) -> str:
         """Handle plotly figure item."""
+        if plotly_to_image is True:
+            img = item.to_image(format='png', engine='auto')
```
JKL98ISR, line 309: use jpg, much smaller

deepchecks/core/display.py
```diff
@@ -0,0 +1,368 @@
...
+
+    @property
+    @abc.abstractmethod
+    def html_serializer(self) -> HtmlSerializer[t.Any]:
+        """Return HtmlSerializer instance."""
+        raise NotImplementedError()
+
+    def show(
+        self,
+        as_widget: bool = True,
```
matanper, line 59: If as_widgets is false, doesn't it mean we need to use the HTML serializer?

deepchecks/core/display.py
yromanyshyn: fixed

deepchecks/core/serialization/check_result/html.py
yromanyshyn: ok, fixed

deepchecks/core/display.py
yromanyshyn: not in the case of the `show` method 
here we are using `IPythonSerializer`

take a look at how it deals with plotly figures 
generally saying, when we use `IPythonSerializer` 
we delegate display responsibility to the plotly figure itself

but in the case of `show_in_iframe` method we `HtmlSerializer`

deepchecks/core/display.py
yromanyshyn: I thought that we going to use iframe by default, therefore, commented it

uncommented (bring it back)

deepchecks/core/display.py
matanper: So If we got the IFrame I wonder if it's smart to continue support also the IPython serializer

yromanyshyn: nope, we need it, I mean `IPythonSerializer`

shortly:
- we cannot use `HtmlSerializer` in the case of `show(as_widget=False)`, because we do not know how plotly figures should be serialized in the current context (therefore we use `IPythonSerializer` and  delegate this task to the plotly itself)
- we cannot use `IPythonSerializer` in the case of `show_in_iframe(as_widget=False) / show_in_window(as_widget=False)`, becasue we cannot use `IPythonSerializer` to produce html document

matanper: yea I meant we will just drop the as_widgets, we will always do it as widgets, unless it's non-interactive

ItayGabbay (approved):

PR #1633, opened 2022-06-15
Add new custom check templates

matanper:
[Custom Check Templates — gallery-boilerplate documentation.pdf](https://github.com/deepchecks/deepchecks/files/8909284/Custom.Check.Templates.gallery-boilerplate.documentation.pdf)

### DIFF
--- a/docs/source/user-guide/tabular/custom_check_templates.rst
+++ b/docs/source/user-guide/tabular/custom_check_templates.rst
@@ -19,6 +19,8 @@ Templates:
 
 Single Dataset Check
 --------------------------
+Check type for cases when running on a single dataset and optional model, for example integrity checks. When in suite
+if 2 datasets are supplied it will run on both independently.
 
 .. code-block::
 
@@ -39,7 +41,7 @@ Single Dataset Check
       def run_logic(self, context: Context, dataset_kind: DatasetKind) -> CheckResult:
           # Get the dataset by its type (train/test)
           dataset: Dataset = context.get_data_by_kind(dataset_kind)
-          # Get the model (if needed)
+          # Get the model (optional, if needed for check logic)
           model = context.model
           # Get from the dataset the data
           data: pd.DataFrame = dataset.data
@@ -78,6 +80,7 @@ Single Dataset Check
 
 Train Test Check
 -----------------
+Check type for cases when running on two datasets and optional model, for example drift checks.
 
 .. code-block::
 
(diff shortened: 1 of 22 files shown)


shir22:
Code examples look great and clear.

Main comment: explanation at top should be a more structured, elaborate and clear.
Both higher level context (why & when is this guide useful?) and also what will we see (why are there different check types? what's the structure that will be demonstrated - result, display, optional condition, etc.) with links where relevant. 
One specific note - would refrain from using the word "bootstrap" (I think it's very dev-oriented).

Another thought - not clear how to use checks with model + dataset... (all examples had either: one dataset, two datasets, or model only...). Also - in one dataset - what is the difference if it's train/test? (why does the run_logic have DatasetKind?)

Should we add "see more" with links to api reference of relevant parts? e.g. Context, DatasetKind, and the types of checks that we're implementing here)

docs/source/user-guide/tabular/custom_check_templates.rst
```diff
@@ -0,0 +1,185 @@
...
+Templates:
+
+* `Single Dataset Check <#single-dataset-check>`__
+* `Train Test Check <#train-test-check>`__
+* `Model Only Check <#model-only-check>`__
+
+
+Single Dataset Check
+--------------------------
+
```
shir22, line 22: Can we add one/two-sentence descriptions for each of these?
(extracted from api reference or written directly here)
For each of the three types

docs/source/user-guide/tabular/custom_check_templates.rst
```diff
@@ -0,0 +1,185 @@
...
+      # OPTIONAL: we can add different properties in the init
+      def __init__(self, prop_a: str, prop_b: str, **kwargs):
+          super().__init__(**kwargs)
+          self.prop_a = prop_a
+          self.prop_b = prop_b
+
+      def run_logic(self, context: Context, dataset_kind: DatasetKind) -> CheckResult:
+          # Get the dataset by its type (train/test)
+          dataset: Dataset = context.get_data_by_kind(dataset_kind)
+          # Get the model (if needed)
```
shir22, line 42:
```suggestion
          # Get the model (optional, if needed for check logic)
```

docs/source/user-guide/tabular/custom_check_templates.rst
```diff
@@ -0,0 +1,185 @@
...
+      def __init__(self, prop_a: str, prop_b: str, **kwargs):
+          super().__init__(**kwargs)
+          self.prop_a = prop_a
+          self.prop_b = prop_b
+
+      def run_logic(self, context: Context) -> CheckResult:
+          # Get the 2 datasets
+          train_dataset: Dataset = context.train
+          test_dataset: Dataset = context.test
+          # Get the model (if needed)
```
shir22, line 103:
```suggestion
          # Get the model (optional, if needed for check logic)
```

shir22 (approved):

PR #1659, opened 2022-06-19
Kaggle display fix

ItayGabbay:
Fixing the display on Kaggle.

---

### DIFF
--- a/deepchecks/core/display.py
+++ b/deepchecks/core/display.py
@@ -94,7 +94,7 @@ class DisplayableResult(abc.ABC):
 
         if is_kaggle_env():
             self.show_in_iframe()
-        if is_colab_env() and as_widget is True:
+        elif is_colab_env() and as_widget is True:
             widget = self.widget_serializer.serialize(**kwargs)
             content = widget_to_html_string(widget, title=get_result_name(self))
             display_html(content, raw=True)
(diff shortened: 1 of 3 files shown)


matanper (approved):

deepchecks/core/display.py
```diff
@@ -92,7 +92,9 @@ def _repr_html_(self):
 
             return TempSphinx()
 
-        if is_colab_env() and as_widget is True:
+        if is_kaggle_env():
```
yromanyshyn, line 95: you need then change it to this

```
if is_kaggle_env() and is_databricks_env() and is_sagemaker_env():
    self.show_in_iframe(as_widget=as_widget, unique_id=unique_id, **kwargs)
```

yromanyshyn (changes requested):

deepchecks/core/display.py
ItayGabbay: Do we have a mechanism to check if we're on sagemaker?

yromanyshyn: plus, I think it is worth showing a warning, something like 
`this envs do not support ipywidgets (or have troubles with thems) therefore we will use iframe`

yromanyshyn (approved):

deepchecks/core/display.py
yromanyshyn: sorry, did not see you prev comment, no we do not have

PR #1667, opened 2022-06-20
[Issue-1561] `features_importance` parameter deprecation

### LINKED ISSUE #1561: [BUG] sometimes we refer to feature_importance as features_importance
**Describe the bug**
sometimes we refer to feature_importance as features_importance.
it is in the context/suite/check params but should be feature_importance instead. (maybe also in other places).

we need to rename it and add deprecation warnings for features_importance

yromanyshyn:
resolves #1561

### DIFF
--- a/deepchecks/tabular/context.py
+++ b/deepchecks/tabular/context.py
@@ -186,7 +186,11 @@ class Context:
                  y_proba_test: np.ndarray = None,
                  features_importance: pd.Series = None,  # TODO: deprecated, should be removed
                  ):
-        feature_importance = features_importance
+        feature_importance = (
+            features_importance  # in case if user used deprecated parameter
+            if feature_importance is None
+            else feature_importance
+        )
         # Validations
         if train is None and test is None and model is None:
             raise DeepchecksValueError('At least one dataset (or model) must be passed to the method!')
--- a/deepchecks/tabular/context.py
+++ b/deepchecks/tabular/context.py
@@ -145,7 +145,8 @@ class Context:
         The name of the model
     features_importance: pd.Series , default: None
         pass manual features importance
-        (parameter is deprecated, use 'feature_importance' instead)
+        .. deprecated:: 0.8.1
+            Use 'feature_importance' instead.
     feature_importance: pd.Series , default: None
         pass manual features importance
     feature_importance_force_permutation : bool , default: False
--- a/deepchecks/tabular/context.py
+++ b/deepchecks/tabular/context.py
@@ -22,6 +22,7 @@ from deepchecks.tabular.dataset import Dataset
 from deepchecks.tabular.utils.task_type import TaskType
 from deepchecks.tabular.utils.validation import (ensure_predictions_proba, ensure_predictions_shape,
                                                  model_type_validation, validate_model)
+from deepchecks.utils.decorators import deprecate_kwarg
 from deepchecks.utils.features import calculate_feature_importance_or_none
 from deepchecks.utils.logger import get_logger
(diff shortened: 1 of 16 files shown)


deepchecks/tabular/context.py
```diff
@@ -144,6 +145,9 @@ class Context:
         The name of the model
     features_importance: pd.Series , default: None
         pass manual features importance
+        (parameter is deprecated, use 'feature_importance' instead)
+    feature_importance: pd.Series , default: None
+        pass manual features importance
```
ItayGabbay, line 150: I think you can use the .. deprecated directive

deepchecks/tabular/context.py
```diff
@@ -274,28 +285,42 @@ def task_type(self) -> TaskType:
 
     @property
     def features_importance(self) -> t.Optional[pd.Series]:
-        """Return features importance, or None if not possible."""
+        """Return feature importance, or None if not possible."""
+        # TODO: remove in future
+        get_logger().warning('"features_importance" property is deprecated use "feature_importance" instead')
+        return self.feature_importance
```
ItayGabbay, line 291: Why not raise the DeprecationWarning?

deepchecks/tabular/context.py
```diff
@@ -274,28 +285,42 @@ def task_type(self) -> TaskType:
...
             self._calculated_importance = True
 
-        return self._features_importance
+        return self._feature_importance
 
     @property
     def features_importance_type(self) -> t.Optional[str]:
+        """Return feature importance type if feature importance is available, else None."""
+        # TODO: remove in future
+        get_logger().warning('"features_importance_type" property is deprecated use "feature_importance_type" instead')
```
ItayGabbay, line 315: Same here. why not raise a DeprecationWarning?

deepchecks/utils/decorators.py
```diff
@@ -0,0 +1,97 @@
...
+            old_arg_value = kwargs.pop(old_arg_name, None)
+
+            if old_arg_value is not None:
+                if new_arg_name is None:
+                    msg = (
+                        f'the {repr(old_arg_name)} keyword is deprecated and '
+                        'will be removed in a future version. Please take '
+                        f'steps to stop the use of {repr(old_arg_name)}'
+                    )
+                    warnings.warn(msg, FutureWarning, stacklevel=stacklevel)
```
ItayGabbay, line 63: FutureWarning? Or DeprecationWarning?

ItayGabbay: Also, how does it work with our verbosity level?

deepchecks/tabular/context.py
yromanyshyn: do you mean something like this?
```python
warnings.warn(<message>, DeprecationWarning)
```
it will not work with our `verbosity level` API,
is not it?

or do you mean something like this? 

```python
 raise DeprecationWarning(<message>)
```
this will raise an error

deepchecks/utils/decorators.py
yromanyshyn: changed to `get_logger().warning(<msg>)`
and `warnings.warn` does not work with our `verbosity level` functionality

ItayGabbay: I think we should be consistent with the package's verbosity level

ItayGabbay: so great

deepchecks/tabular/context.py
yromanyshyn: fixed

PR #1680, opened 2022-06-24
[Issue-1560] fix effect of display control parameters on `CheckResult.value`

### LINKED ISSUE #1560: [BUG] some checks change the value according to n_show_top instead of only the display
**Describe the bug**
some checks change the value according to n_show_top instead of only the display 

**To Reproduce**
for now we know it happens in the Conflicting Labels Check  but it may be in other checks too

**Expected behavior**
it should affect only the display

**Additional context**
We need to check all checks that are using this param and make sure n_show_top works as intended and fix otherwise.
related to #1514

yromanyshyn:
resolves #1560

### DIFF
--- a/deepchecks/vision/checks/model_evaluation/image_segment_performance.py
+++ b/deepchecks/vision/checks/model_evaluation/image_segment_performance.py
@@ -162,16 +162,24 @@ class ImageSegmentPerformance(SingleDatasetCheck):
                 result_value[property_name].append(single_bin)
 
         display_df = pd.DataFrame(display_data)
+        
         if display_df.empty:
             return CheckResult(value=dict(result_value))
+        
         first_metric = display_df['Metric'][0]
+        
         if self.alternative_metrics is None:
             display_df = display_df[display_df['Metric'] == first_metric]
-        top_properties = display_df[display_df['Metric'] == first_metric] \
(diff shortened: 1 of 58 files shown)


matanper (approved):

deepchecks/vision/checks/model_evaluation/image_segment_performance.py
```diff
@@ -162,16 +162,24 @@ def compute(self, context: Context, dataset_kind: DatasetKind) -> CheckResult:
                 result_value[property_name].append(single_bin)
 
         display_df = pd.DataFrame(display_data)
```
JKL98ISR, line 164: can you write a similiar todo as in:
> # TODO: this for loop should be below `if context.with_display:` branch

yromanyshyn:
@ItayGabbay @matanper @JKL98ISR @noamzbr 
Question:

whether parameters like `class_list_to_show`, `show_only`, and `n_to_show` should have an effect on the `ClassPerformance` conditions? (and similarry in other similar checks)

if yes, then we need to store `full` and `filtered` result dataframes in the `CheckResult.value` /

ItayGabbay:
Hmmm in my opinion the conditions shouldn't be affected by the `n_to_show`. I'd leave the current logic as is and open it for discussion later

PR #1681, opened 2022-06-26
Vision benchmark

### LINKED ISSUE #1626: [FEAT] Profiling the vision package
Following #1566 (which is profiling the tabular package), we should know what is taking the most time & memory in the vision package, and how we can improve that

1. See if the profiling used in #1566 fits also the vision.
2. Summarize profiling of all vision checks in presentable forms
3. For checks which are slow/heavy try to find the cause

JKL98ISR:
closes #1626


.github/workflows/benchmark-history.yml
```diff
@@ -40,4 +40,26 @@ jobs:
       run: make freeze
     - name: Run benchmarks
       run: |
-        .github/scripts/benchmark-history.sh
\ No newline at end of file
+        .github/scripts/benchmark-history.sh benchmarks/tabular
+
+  benchmark-history-vision:
```
ItayGabbay, line 45: Why separate to different jobs..? Can we somehow do it in the same script?

.github/workflows/benchmark.yml
```diff
@@ -34,4 +34,22 @@ jobs:
       run: make freeze
     - name: Run benchmarks
       run: |
-        .github/scripts/benchmark.sh
\ No newline at end of file
+        .github/scripts/benchmark.sh benchmarks/tabular
+  
+  benchmark-vision:
```
ItayGabbay, line 39: Same comment here

ItayGabbay (commented):
Great PR!

.github/workflows/benchmark-history.yml
JKL98ISR: but why? it looks better like that
you can do everything in one big job

.github/workflows/benchmark.yml
JKL98ISR: Same comment here

.github/workflows/benchmark-history.yml
ItayGabbay: Because: 
1. It will require more resources from GitHub
2. I don't know how it will look in history. did you check?

ItayGabbay (approved):

PR #1694, opened 2022-06-28
[issue-1514] conflicting labels check fix

### LINKED ISSUE #1514: [FEAT][Tabular][Checks] Allow access to all checks data not only top showing
I was running the Conflicting Labels Check on my dataset. which worked great and showed me that there were 476 records with conflicting labels, but it only showed me the 5 of them (which is what I would expect since that's the default)
Then I wanted to fix the issue by observing the record and removing them, But there wasn't any way to access all of them.
the only way I figured out how to do it is by running the test again with n_to_show=(the number the message showed me)

I would like the test to return all relevant data to get to the problem faster.  it may show me only n_to_show, but when ill access the test object I should get all indexes of conflicting Labels.

I did notice I'm able to access the data the test returns, but it only returns the n_to_show. so just return everything and only **show** the relevant rows per the argument
**Additional context**

yromanyshyn:
resolves #1514

### DIFF
--- a/deepchecks/tabular/checks/data_integrity/conflicting_labels.py
+++ b/deepchecks/tabular/checks/data_integrity/conflicting_labels.py
@@ -65,11 +65,11 @@ class ConflictingLabels(SingleDatasetCheck):
         dataset = context.get_data_by_kind(dataset_kind)
         context.assert_classification_task()
         dataset.assert_label()
-        features = dataset.features
-        label_name = dataset.label_name
         with_display = context.with_display
 
         dataset = dataset.select(self.columns, self.ignore_columns, keep_label=True)
+        features = dataset.features
+        label_name = dataset.label_name
 
         # HACK: pandas have bug with groupby on category dtypes, so until it fixed, change dtypes manually
         df = dataset.data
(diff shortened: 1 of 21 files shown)


deepchecks/tabular/checks/data_integrity/conflicting_labels.py
```diff
@@ -59,50 +65,67 @@ def run_logic(self, context: Context, dataset_kind) -> CheckResult:
...
-        return CheckResult(value=percent_ambiguous, display=display)
+            samples.append(group_df.loc[:, [label_name, *features]].copy())
+
+            if with_display is True:
+                display_sample = dict(group_df[features].iloc[0])
+                ambiguous_labels = tuple(sorted(group_df[label_name].unique()))
+                display_sample[ambiguous_label_name] = ambiguous_labels
+                display_samples.append(display_sample)
+
+        if not (with_display is True and len(display_samples) > 0):
```
JKL98ISR, line 110: can just be len(display_samples) == 0, because it won't append anything if with_display was false 
and if with_display is used once no need to make a variable

JKL98ISR:
add to the run_logic doc that it returns a dict now with number of samples too

yromanyshyn:
@JKL98ISR 

>add to the run_logic doc that it returns a dict now with number of samples too

fixed

PR #1699, opened 2022-06-30
1604 docs metrics guide

TheSolY:
Deepchecks presents: An informative and engaging Metrics Guide! Read for free on our website!


docs/source/user-guide/general/metrics_guide.rst
```diff
@@ -0,0 +1,259 @@
...
+*   `Jaccard Index <https://en.wikipedia.org/wiki/Jaccard_index>`__ / IoU - the ratio between the overlapping area of
+    predicted and the actual bounding box and the union of their areas. The most intuitive metric for object detection.
+*   mAP - mean average precision. The mean of the average precision per class over the classes. Calculated at a certain
+    threshold, usually 0.5. Commonly used for benchmarking object detection models. A good explanation of how it is
+    calculated can be found
+    `here <https://towardsdatascience.com/breaking-down-mean-average-precision-map-ae462f623a52>`__.
+*   mAR - mean average recall. The mean of the average precision per class over the classes. Summed over the threshold
+    range [0.5, 1].
+
+How can I use metrics with Deepchecks?
```
Nadav-Barak, line 174: Would extend this section, mention all metrics currently supported by deepchecks and their deepchecks str name (for sklearn methods it is as mentioned here https://scikit-learn.org/stable/modules/model_evaluation.html)
Not sure the default are consistent across all check so maybe avoid saying what are the defaults
Would add example how to supply a metric to a specific check (as a list of strings to the alternative_scorers argument)

docs/source/user-guide/general/metrics_guide.rst
```diff
@@ -0,0 +1,259 @@
...
+`source <https://commons.wikimedia.org/w/index.php?curid=94134880>`__
+
+*   `F-1 <https://en.wikipedia.org/wiki/F-score>`__ - combines the precision and recall into one metric by taking their
+    harmonic mean. More robust to class imbalance than accuracy.
+*   `AUC <https://developers.google.com/machine-learning/crash-course/classification/roc-and-auc>`__  - The area under
+    the curve of the ROC graph. For the previous classification metrics, a threshold on the model output was chosen
+    above which the sample is classified as positive and below as negative. The AUC takes into account all possible
+    thresholds.
+
+Object Detection
```
Nadav-Barak, line 163: Would explain how FP, TP, FN are calculated for object detection (what JKL did)
Also add them as metric options

TheSolY: @Nadav-Barak explained in the link, WDYM as metric options?

docs/source/user-guide/general/metrics_guide.rst
TheSolY: @Nadav-Barak do we have this kind of example somewhere in the docs? This sklearn link is already there under "Alternative Metrics" (line 225)

docs/source/user-guide/general/metrics_guide.rst
```diff
@@ -0,0 +1,259 @@
...
+
+Tabular:
+
+Classification, binary or multiclass averaged over the classes:
+
+*   Accuracy
+*   Precision
+*   Recall
+
+Multiclass classification per class:
```
Nadav-Barak, line 193: Describe the different averaging options for those metrics (macro,micro etc..)
Also mention their string name such as "f1_micro"

Nadav-Barak (changes requested):

docs/source/user-guide/general/metrics_guide.rst
Nadav-Barak: IMO I would explain them specifically here in a different paragraph as they can be regarded as metric on their on

docs/source/user-guide/general/metrics_guide.rst
Nadav-Barak: I think it should be an integral part here, we should describe all the metrics we support and their str value the user need to pass in our check to get those metrics
It is a big requirement, so better if you approve @shir22

TheSolY: @Nadav-Barak, agreed to wait until the implementation of str and Sklearn metrics for vision is complete

docs/source/user-guide/general/metrics_guide.rst
TheSolY: @Nadav-Barak I think it's too much information, we want to keep it short and simple

shir22:
Main comment - Structure + Splitting into two
=====================================

This should be split into two parts: first part (up to - "How can I use metrics with Deepchecks") should be in a separate page, "Intro to Model Performance Metrics" (or any other title, making it clear that this is content for juniors...)

Second part (from "How can I use metrics with Deepchecks?") - is a very good basis as a "Using Metrics with Deepchecks" Guide. Should be tweaked a bit, such that a user can super easily understand: how do I use other metrics, and which code template can I use as a basis... Thus - it should be structured with that main purpose in mind.
Specifically - adding a general explanation (about many of the model evaluation checks using metrics, that they all have defaults that are task-type dependant, and that they can be customized etc.)
Then the explanation about alternative metrics - incl. code examples (inline and not in the end) of how to customize + the relevant links.
Then the part about "these are the defaults used" - and this can be linked from the top.

Tactical Comment - Linking from checks that have alternative_scorers
------------------------------------------------------------------------

API Reference and check examples should of all relevant checks should include link to this guide (the "second part" one), which was its main purpose.. thus - this should be included here (otherwise - opened in a separate issue which should be completed in upcoming milestone)

Tatical Comment - Image Captions
------------------------------------

They don't look good.
1. Is there a way to have them as actual captions (and not as text underneath)? (I think they should be indented along with image directive?)
2. Consider dropping or shortening where possible.

docs/source/user-guide/general/metrics_guide.rst
shir22: I don't agree about keeping it short and simple... the main target of this guide is so the user will be able to understand how to use the alternative_metrics

docs/source/user-guide/general/metrics_guide.rst
shir22: @TheSolY let's make sure there is an issue up to update this accordingly?

shir22:
If want to override alternative_scorers in a given suite - for all checks, is this possible? (using run()? or only in check initialization?)
If so - should be also demonstrated

docs/source/user-guide/general/metrics_guide.rst
TheSolY: @shir22 done https://github.com/deepchecks/deepchecks/issues/1705

docs/source/user-guide/general/metrics_guide.rst
Nadav-Barak: I think it is important, if a user pass f1 score to a multilabel task our suite will fail as he must specify an averaging option

PR #1717, opened 2022-07-04
[suggestion] html progress bar

yromanyshyn:
`tqdm` library in order to display a progress bar in notebooks uses `ipywidgets` library, 
this is a problem because as we know:
- in order to work `ipywidgets` needs additional jupyter extensions
- there is no unified or any good way to determine whether those extensions are enabled/installed and whether `ipywidgets` will work or not

I suggest to start use HTML [progress](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/progress) tag in order to display a progress bar

what has been done:
- added `HtmlProgressBar` class
- modified `create_progress_bar` function
- removed all functionality that tries to determine whether `ipywidgets` use is possible

https://user-images.githubusercontent.com/71635444/177122504-d9e11244-24e9-4a8d-bd17-ed01db9e5a8f.mp4


ItayGabbay:
@JKL98ISR what is your opinion on that?

ItayGabbay (approved):

PR #1718, opened 2022-07-04
Ignore NaNs in FeatureFeatureCorrelation

TheSolY:
Minor fix to the encoding, the rest is the default behavior of pandas corr. 
In the future we should implement a function to encode categorical dataframe with the options to keep NaNs or replace the with an int.


deepchecks/tabular/checks/data_integrity/feature_feature_correlation.py
```diff
@@ -81,6 +82,8 @@ def run_logic(self, context: Context, dataset_kind) -> CheckResult:
         num_features = [f for f in dataset.numerical_features if f in df.columns]
         cat_features = [f for f in dataset.cat_features if f in df.columns]
         encoded_cat_data = df.loc[:, cat_features].apply(lambda x: pd.factorize(x)[0])
+        # NaNs are encoded as -1, replace back to NaN
+        encoded_cat_data.replace(-1, np.NaN, inplace=True)
```
noamzbr, line 86: Better use pd.factorize own's `na_sentinel=None` parameter

tests/tabular/checks/integrity/feature_feature_correlation_test.py
```diff
@@ -78,9 +78,7 @@ def test_feature_feature_correlation_pass_condition(adult_no_split):
 def test_feature_feature_correlation_fail_condition(adult_no_split):
     threshold = 0.5
     num_pairs = 3
-    high_pairs = [('age', 'marital-status'), ('education-num', 'occupation'), ('edu'
-                                                                               ''
-                                                                               'cation', 'education-num'),
+    high_pairs = [('age', 'marital-status'), ('education-num', 'occupation'), ('education', 'education-num'),
```
noamzbr, line 81: Should add tests for the nan case

TheSolY: @noamzbr There's already a test for that

deepchecks/tabular/checks/data_integrity/feature_feature_correlation.py
TheSolY: @noamzbr It replaces NaN with max cat + 1 (different number for each column), doesn't keep it as None, so it's better to replace all with -1 (the default) and then replace back to NaN

noamzbr: Right, my bad

tests/tabular/checks/integrity/feature_feature_correlation_test.py
JKL98ISR: where?

TheSolY: @JKL98ISR the test is called test_feature_feature_correlation_with_mixed_data

JKL98ISR: so how did nothing change?

TheSolY: @JKL98ISR because it was not a bug to begin with, just a decision which of the options we should go for- both return a legit answer. I don't think we should check the actual values.

JKL98ISR (approved):

PR #1736, opened 2022-07-06
Suite/check conf serialize

JKL98ISR:


deepchecks/core/checks.py
```diff
@@ -160,6 +167,62 @@ def metadata(self, with_doc_link: bool = False) -> CheckMetadata:
             summary=get_docs_summary(self, with_doc_link)
         )
 
+    def config(self, with_module=True) -> CheckConfig:
+        """Return check config (conditions' config not yet supported).
+
+        Parameters
+        ----------
+        with_module : bool, default True
```
matanper, line 175: when we will set it to False?

deepchecks/core/checks.py
```diff
@@ -160,6 +167,62 @@ def metadata(self, with_doc_link: bool = False) -> CheckMetadata:
...
+        DeepchecksValueError
+            if the CheckConfig doesn't include the module name and module_name is None.
+        """
+        module_name = conf.get('module_name', module_name)
+        if module_name is None:
+            raise DeepchecksValueError(
+                'Check Config must include module name or it should be passed to the function.')
+        is_deepchecks = module_name.startswith('deepchecks.')
+        if is_deepchecks:
+            module_name += '.checks'
```
matanper, line 221: why?

deepchecks/core/checks.py
```diff
@@ -160,6 +167,62 @@ def metadata(self, with_doc_link: bool = False) -> CheckMetadata:
...
+            whethere to include the module path or not
+
+        Returns
+        -------
+        CheckConfig
+            includes the checks class name, params, and module if with_module was True
+        """
+        module_name = self.__module__
+        if module_name.startswith('deepchecks.'):
+            module_name = 'deepchecks.' + module_name.split('.')[1]
```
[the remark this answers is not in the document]
JKL98ISR: so the module path will be just deepchecks.tabular or vision and we won't import checks from a deprecated place

deepchecks/core/checks.py
JKL98ISR: if we're doing it for our own suites we won't need the module

JKL98ISR: as it'll be set in the suite

deepchecks/core/checks.py
JKL98ISR: also that way we can use the module as in the suite

deepchecks/core/checks.py
```diff
@@ -160,6 +167,62 @@ def metadata(self, with_doc_link: bool = False) -> CheckMetadata:
...
+        """
+        module_name = conf.get('module_name', module_name)
+        if module_name is None:
+            raise DeepchecksValueError(
+                'Check Config must include module name or it should be passed to the function.')
+        is_deepchecks = module_name.startswith('deepchecks.')
+        if is_deepchecks:
+            module_name += '.checks'
+        module = importlib.import_module(module_name)
+        return getattr(module, conf['class_name'])(**conf['params'])
```
yromanyshyn, line 223: it is worth verifying whether all checks store their init params, IMO, not all do this/

JKL98ISR: wdym store?

JKL98ISR: the params are taken from the init function and I'm calling the init

JKL98ISR: i don't use any param used in the init itself, it is by the signature

JKL98ISR: oh, but the validation needs to be in the param function then

JKL98ISR: if a parameter isn't stored it is a problem we can't fix rn, we relied on this function and the behavior remains the same in my pr

yromanyshyn: 1. yes, it is true for user checks, but do our checks store all init params, and under what name? did you check that? that was my initial question)
2. IMO, it is worth modifying `initvars` because you will not get default values from the init signature if init params were not assigned to the objected, this could be changed

JKL98ISR: still not related to this pr particularly, we might want to reconsider hot that function work, but should be a new issue.

yromanyshyn: point 1.

`MultiModelPerformanceReport` stores parameter `alternative_scorers` under name `user_scorers`

```python
    def __init__(self, alternative_scorers: Dict[str, Callable] = None, **kwargs):
        super().__init__(**kwargs)
        self.user_scorers = alternative_scorers
```

you will not get this parameter by calling `initvars`

yromanyshyn: same for `SegmentPerformance`
note: I know that callable serialization is currently not supported by this method, but `alternative_scorer` also could be of type `Dict[str, str]` or `Tuple[str, str]`

deepchecks/core/checks.py
matanper: so add a comment "remove this after removing .checks from the library imports"

deepchecks/core/checks.py
matanper: Also we have parameters which stores default functions which is also problematic no?
Image segment performance
```
        if image_properties:
            self.image_properties = validate_properties(image_properties)
        else:
            self.image_properties = default_image_properties
```

JKL98ISR: still should be a new issue, it is a bug about how we get the params

deepchecks/core/checks.py
matanper: but why do you care about having it twice to simplify the code?

deepchecks/core/checks.py
JKL98ISR: https://github.com/deepchecks/deepchecks/issues/1739

ItayGabbay (approved):

Issue #1700, opened 2022-07-03
[BUG] Outlier sample detection exception


ItayGabbay:
**Describe the bug**
When scrolling through the docs I saw this error in the integrity suite:
![image](https://user-images.githubusercontent.com/20860465/177029968-6282acf6-675f-486f-90ca-f692743832b3.png)

https://docs.deepchecks.com/stable/user-guide/tabular/auto_tutorials/plot_quick_data_integrity.html

Nadav-Barak:
DOD: Reproduce and fix

PR #1741, opened 2022-07-07
handle single value columns gower dist

Nadav-Barak:
Resolves https://github.com/deepchecks/deepchecks/issues/1700

### DIFF
--- a/deepchecks/utils/gower_distance.py
+++ b/deepchecks/utils/gower_distance.py
@@ -11,7 +11,6 @@
 """Module for calculating distance matrix via Gower method."""
 import numpy as np
 import pandas as pd
-from sklearn.preprocessing import OrdinalEncoder
 
 from deepchecks.utils.array_math import fast_sum_by_row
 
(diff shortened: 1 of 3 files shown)


deepchecks/utils/gower_distance.py
```diff
@@ -84,13 +83,11 @@ def calculate_nearest_neighbours_distances(cat_data: pd.DataFrame, numeric_data:
...
-    enc = OrdinalEncoder()
-    if not cat_data.empty:
-        cat_data = cat_data.apply(lambda x: pd.factorize(x)[0])
-    cat_data = enc.fit_transform(cat_data)
+    cat_data = np.asarray(cat_data.apply(lambda x: pd.factorize(x)[0])) if not cat_data.empty else np.asarray(cat_data)
     # handle numerical - calculate ranges per feature and fill numerical nan to minus np.inf
-    numeric_data = np.asarray(numeric_data.fillna(value=np.nan))
+    numeric_data = np.asarray(numeric_data.fillna(value=np.nan)).astype('float64')
     numeric_feature_ranges = np.nanmax(numeric_data, axis=0) - np.nanmin(numeric_data, axis=0)
+    numeric_feature_ranges = np.where(numeric_feature_ranges == 0, 1, numeric_feature_ranges)
```
nirhutnik, line 90: Why range = 1 for those without range? I understand it probably causes a problem, but:
1. This is used later in comparison to other numeric features, no? So the values matter. If I understand correctly, if a user has a feature with a constant value (let's say 3) and another on the range (0,0.0001) then the first will have range=1 and the other range=0.0001, and because you do use these values to scale the features for gower distance, this creates an issue... no?

In general, why not drop columns with a constant value? they have no meaning distance-wise.

Nadav-Barak: We can drop them but it doesn't make a lot of difference run time vise, since the numeric feature data distance is vectorized (the main constrain on run time is the loop). logically, gower score is based on all columns (can see in the test added) without uniform column we will get higher outlier score (imaging a case where in the sample the column is uniform but not over all data).
is the case of range 0, in mean that in the gower calculation there will be 0 / 0 operation, now it will be 0 / 1 and overall will not effect the outlier calculations scores (as the distance per feature between samples with same value should be 0)

Nadav-Barak: This is used later in comparison to other numeric features, no? - it does not

ItayGabbay (approved):

PR #1745, opened 2022-07-10
Scorers string

### LINKED ISSUE #1630: Refactor alternative metric mechanism to support new metrics and string metrics in vision
There are several basic metrics that are currently not supported and we should implement, those include:
Classification: Sensitivity (recall), [specificity](https://www.jigsawacademy.com/sensitivity-vs-specificity-in-logistic-regression/), FPR, FNR, Precision-recall AUC
Object detection: in other issue - https://github.com/deepchecks/deepchecks/issues/1634

Implement the metrics internally, make sure those metrics are implemented support the DeepcheckScorer class api. In addition keep in mind those metrics should also be applicable to vision classification tasks.
Add unit tests to make sure those metrics return the correct value + work properly with the DeepcheckScorer API

JKL98ISR:
closes #1630

### DIFF
--- a/deepchecks/tabular/metric_utils/metrics.py
+++ b/deepchecks/tabular/metric_utils/metrics.py
@@ -9,7 +9,6 @@
 # ----------------------------------------------------------------------------
 #
 """Utils module containing utilities for checks working with metrics."""
-import copy
 import typing as t
 from numbers import Number
 
@@ -35,7 +34,6 @@ __all__ = [
     'DEFAULT_MULTICLASS_SCORERS',
     'MULTICLASS_SCORERS_NON_AVERAGE',
     'DeepcheckScorer',
-    'get_gain',
     'init_validate_scorers',
     'get_default_scorers'
 ]
@@ -49,8 +47,8 @@ DEFAULT_BINARY_SCORERS = {
 
 DEFAULT_MULTICLASS_SCORERS = {
     'Accuracy': 'accuracy',
-    'Precision - Macro Average': 'precision_macro',
-    'Recall - Macro Average': 'recall_macro',
+    'Precision - Macro Average': make_scorer(precision_score, average='macro', zero_division=0),
+    'Recall - Macro Average': make_scorer(recall_score, average='macro', zero_division=0),
 }
 
 MULTICLASS_SCORERS_NON_AVERAGE = {
@@ -254,7 +252,7 @@ def get_default_scorers(model_type, class_avg: bool = True):
         return DEFAULT_SCORERS_DICT[model_type]
 
 
-def init_validate_scorers(scorers: t.Mapping[str, t.Union[str, t.Callable]],
+def init_validate_scorers(scorers: t.Union[t.Mapping[str, t.Union[str, t.Callable]], t.List[str]],
                           model: BasicModel,
                           dataset: 'tabular.Dataset',
                           model_type: TaskType,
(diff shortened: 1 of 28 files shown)


deepchecks/tabular/metric_utils/metrics.py
```diff
@@ -0,0 +1,279 @@
...
+        scorer name
+    """
+
+    def __init__(self, scorer: t.Union[str, t.Callable], name: str):
+        self.name = name
+        if isinstance(scorer, str):
+            formated_scorer_name = scorer.lower().replace('sensitivity', 'recall').replace(' ', '_')
+            if formated_scorer_name in _func_dict:
+                self.scorer = _func_dict[formated_scorer_name]
+            else:
```
Nadav-Barak, line 101: would add code comment - this uses sklearn string to scorer mechanism https://scikit-learn.org/stable/modules/model_evaluation.html

deepchecks/tabular/metric_utils/metrics.py
```diff
@@ -0,0 +1,279 @@
...
+    name : str
+        scorer name
+    """
+
+    def __init__(self, scorer: t.Union[str, t.Callable], name: str):
+        self.name = name
+        if isinstance(scorer, str):
+            formated_scorer_name = scorer.lower().replace('sensitivity', 'recall').replace(' ', '_')
+            if formated_scorer_name in _func_dict:
+                self.scorer = _func_dict[formated_scorer_name]
```
Nadav-Barak, line 100: in those case,  the name displayed in graph different (without the per_class)

Nadav-Barak: i.e change self.name

deepchecks/tabular/metric_utils/metrics.py
```diff
@@ -0,0 +1,279 @@
...
+}
+
+DEFAULT_SCORERS_DICT = {
+    TaskType.BINARY: DEFAULT_BINARY_SCORERS,
+    TaskType.MULTICLASS: DEFAULT_MULTICLASS_SCORERS,
+    TaskType.REGRESSION: DEFAULT_REGRESSION_SCORERS
+}
+
+
+_func_dict = {
```
Nadav-Barak, line 73: Additional thing needed here - all the regression stuff + 'Accuracy'

deepchecks/tabular/metric_utils/metrics.py
```diff
@@ -0,0 +1,279 @@
...
+
+    Parameters
+    ----------
+    scorer : t.Union[str, t.Callable]
+        sklearn scorer name or callable
+    name : str
+        scorer name
+    """
+
+    def __init__(self, scorer: t.Union[str, t.Callable], name: str):
```
Nadav-Barak, line 95: Additional request - add a class field indicating whether the scorer returns a float or a list of float + a public method is_scalar_scorer

JKL98ISR: not now

deepchecks/tabular/metric_utils/metrics.py
JKL98ISR: yes

deepchecks/tabular/metric_utils/metrics.py
JKL98ISR: this is already there, need to add to vision

Nadav-Barak (approved):
Add test to see at least one of the checks works with getting a list of str to alternative_scorers
+ see what happens if you pass str that do not exist in sklearn and in ours

deepchecks/vision/metrics_utils/metrics.py
```diff
@@ -45,18 +49,58 @@ def get_default_object_detection_scorers() -> t.Dict[str, Metric]:
     }
 
 
+_func_naming_dict = {
+    'precision': 'Precision',
+    'recall': 'Recall',
```
noamzbr, line 54: I think this are the per class options, right? If so, perhaps for consistency we should call these precision_per_class etc
@Nadav-Barak WDYT?

JKL98ISR: doesn't matter much tbh
default behavior will remain the same
don't mind changing them but not sure if it will make much sense cuz all the name will be longer and all the threshold metrics names won't correspond the the name in that dict

deepchecks/vision/metrics_utils/metrics.py
```diff
@@ -45,18 +49,58 @@ def get_default_object_detection_scorers() -> t.Dict[str, Metric]:
...
+    'sensitivity': 'Recall',
+    'average_precision': 'Average Precision',
+    'average_recall': 'Average Recall',
```
Nadav-Barak, line 57: If not too messy, also add mean_average_precision and mean_average_recall options

JKL98ISR: kinda is

yromanyshyn (approved):

PR #1746, opened 2022-07-10
Separate gower distance function

nirhutnik:
Changed gower distance API - change function name, change inputs to better suite our package, and add functionality to choose which samples the distance will be calculated for.


Nadav-Barak (approved):
So american of you

JKL98ISR (approved):

PR #1792, opened 2022-07-18
Nb/feat/update custom check guide

### LINKED ISSUE #1617: [DOCS] custom check guide is outdated
- [x] Update the guide to reflect current state
- [x] Add more info about the context
- [ ] Possibly change its location to be under "customizations" + make it clear this is a tabular check

noamzbr:
#### Reference Issues/PRs
Resolved #1617 

Things done

1. Elaborate on Context and Batch
2. Link to guides from customization
3. Link to guides from the check galleries


docs/source/user-guide/general/customizations/plot_create_a_custom_check.py
```diff
@@ -0,0 +1,17 @@
...
+Create a Custom Check
+=====================
+
+It is possible to extend deepchecks by implementing custom checks. This
+enables you to have your own logic of metrics or validation, or even just
+to display your own graph using deepchecks' suite.
+
+While some aspects for writing a check are similar when writing checks for the
+``deepchecks.tabular`` ``and deepchecks.vision`` subpackages, there are major differences
+that warrant a separate guide for each:
```
TheSolY, line 12: IMO this paragraph is more confusing than helpful, we should just say something like: here is the guide for tabular (tabular link) and here is the guide for vision (vision link)

matanper (approved):

TheSolY (commented):
I think that we need to add an empty template to these pages https://docs.deepchecks.com/stable/user-guide/vision/custom_check_templates.html, 
https://docs.deepchecks.com/stable/user-guide/tabular/custom_check_templates.html
The examples are too cluttered with comments ad example items. An empty check class with the necessary methods (empty)  that the user can copy paste

matanper:
The PR is nice but I think we also need:
1. When to use different parameters: init vs run vs condition params
2. More info about possible display options, for example we have the Tabs now, we should expose it
3. Move the "base check types" to be in the start 
4. Add information how to write condition

noamzbr:
Comments added to future issue
https://github.com/deepchecks/deepchecks/issues/1796

PR #1794, opened 2022-07-19
small fixes to drift and gower distance

nirhutnik:
Small fixes I encountered - 

1. Gower distance returns iloc indexes and not original indexes
2. categorical drift comments quote the needed value for "cramer_v" wrong. Also, added the ability to accept slight differences (using lower case instead of upper, and "cramers_v" instead of "cramer_v")


noamzbr (approved):

nirhutnik:
(didn't merge changes to gower distance)

PR #1821, opened 2022-07-25
vision properties refactor preliminary 

TheSolY:
Functionality is complete, still left to adjust to unittests and documentation after this PR is closed


deepchecks/vision/batch_wrapper.py
```diff
@@ -94,6 +96,23 @@ def __len__(self):
         dataset = self._context.get_data_by_kind(self._dataset_kind)
         return len(list(dataset.data_loader.batch_sampler)[self.batch_index])
 
+    def vision_properties(self, raw_data: List, properties_list: List[Dict], input_type=PropertiesInputType):
```
nirhutnik, line 99:
```suggestion
    def vision_properties(self, raw_data: List, properties_list: List[Dict], input_type: PropertiesInputType):
```
I guess you meant this?

deepchecks/vision/batch_wrapper.py
```diff
@@ -94,6 +96,23 @@ def __len__(self):
...
+        # if there are no cached properties at all, calculate all the properties on the list,
+        # else calculate only those that were not yet calculated.
+        dataset = self._context.get_data_by_kind(self._dataset_kind)
+        if self._vision_properties_cache[input_type.value] is None:
+            self._vision_properties_cache[input_type.value] = dataset.calc_properties(raw_data, properties_list)
+        else:
+            properties_to_calc = [p for p in properties_list if p['name'] not in
+                                  self._vision_properties_cache[input_type.value].keys()]
+            self._vision_properties_cache[input_type.value].update(
+                                                                dataset.calc_properties(raw_data, properties_to_calc))
```
nirhutnik, line 111: What happens if we want to run them on a sample, and not on all of the data? Seems the cache is not per index of the data, right?

deepchecks/vision/checks/data_integrity/abstract_property_outliers.py
```diff
@@ -57,51 +58,44 @@ class AbstractPropertyOutliers(SingleDatasetCheck):
         Two percentiles which define the IQR range
     iqr_scale: float, default: 1.5
         The scale to multiply the IQR range for the outliers detection
+    properties_precalculated: bool, default: False
```
nirhutnik, line 61: name is a bit weird. I suggest maybe "are_properties_precalculated" or just "precalculated_properties"

deepchecks/vision/checks/data_integrity/abstract_property_outliers.py
```diff
@@ -57,51 +58,44 @@ class AbstractPropertyOutliers(SingleDatasetCheck):
...
         # Take either alternative properties if defined or default properties defined by the child class
-        if self.user_properties is not None:
-            self._properties_funcs = self.user_properties
-        else:
-            self._properties_funcs = self.get_default_properties(data)
-            # Filter out properties that have class_id as output_type
-            self._properties_funcs = [p for p in self._properties_funcs if p['output_type'] != 'class_id']
+        self.properties_list = self.properties_list if self.properties_list else self.get_default_properties(data)
+        if any(p['output_type'] == 'class_id' for p in self.properties_list):
+            warnings.warn('Properties that have class_id as output_type will be skipped.')
```
nirhutnik, line 89: Why? Because they don't make sense in the context of outliers?

deepchecks/vision/checks/model_evaluation/image_segment_performance.py
```diff
@@ -86,18 +83,14 @@ def initialize_run(self, context: Context, dataset_kind: DatasetKind):
 
     def update(self, context: Context, batch: Batch, dataset_kind: DatasetKind):
         """Update the bins by the image properties."""
-        images = batch.images
         predictions = [tens.detach() for tens in batch.predictions]
         labels = [tens.detach() for tens in batch.labels]
+        properties_results = batch.vision_properties(batch.images, self.image_properties, PropertiesInputType.IMAGES)
```
nirhutnik, line 88: So, now your validation of properties happens in the initialize_run instead of the __init__. I think that this is not the right behavior - the validation should happen when its defined, if possible.
Why not validate then?

deepchecks/vision/checks/train_test_validation/property_label_correlation_change.py
```diff
@@ -118,12 +116,16 @@ def __init__(
 
     def update(self, context: Context, batch: Batch, dataset_kind: DatasetKind):
         """Calculate image properties for train or test batches."""
+
+        dataset = context.get_data_by_kind(dataset_kind)
+        # properties = batch.vision_properties
```
nirhutnik, line 121: left a comment here

deepchecks/vision/checks/train_test_validation/property_label_correlation_change.py
```diff
@@ -118,12 +116,16 @@ def __init__(
...
+
         if dataset_kind == DatasetKind.TRAIN:
-            dataset = context.train
-            properties = self._train_properties
+            # dataset = context.train
+            properties_results = self._train_properties
         else:
-            dataset = context.test
-            properties = self._test_properties
+            # dataset = context.test
```
nirhutnik, line 127: and here and above

deepchecks/vision/batch_wrapper.py
TheSolY: @nirhutnik we decided together with @ItayGabbay that we cache on all the data and filter later, you can see ImagePropertyDrift for example

deepchecks/vision/checks/data_integrity/abstract_property_outliers.py
TheSolY: @nirhutnik This parameter was removed, now removed from docstring as well

deepchecks/vision/checks/data_integrity/abstract_property_outliers.py
TheSolY: Yes, the outliers on the class id are meaningless because it's only encoding of non-ordinal data

deepchecks/vision/batch_wrapper.py
```diff
@@ -94,6 +96,23 @@ def __len__(self):
         dataset = self._context.get_data_by_kind(self._dataset_kind)
         return len(list(dataset.data_loader.batch_sampler)[self.batch_index])
 
+    def vision_properties(self, raw_data: List, properties_list: List[Dict], input_type=PropertiesInputType):
+        """Calculate and cache the properties for the batch according to the propety input type"""
+        properties_list = validate_properties(properties_list)
+        # if there are no cached properties at all, calculate all the properties on the list,
+        # else calculate only those that were not yet calculated.
+        dataset = self._context.get_data_by_kind(self._dataset_kind)
+        if self._vision_properties_cache[input_type.value] is None:
```
JKL98ISR, line 105: no need for that if else if you will do from_keys([..], [])

deepchecks/vision/checks/model_evaluation/image_segment_performance.py
TheSolY: As I see it, the general validations are the job of the batch/context and not the check

deepchecks/vision/batch_wrapper.py
JKL98ISR, line 99:
```suggestion
    def vision_properties(self, raw_data: List, properties_list: List[Dict], input_type: PropertiesInputType = PropertiesInputType):
```

JKL98ISR:
I don't see how the properties can be set from the context as the predictions can

deepchecks/vision/batch_wrapper.py
```diff
@@ -94,6 +96,23 @@ def __len__(self):
...
+    def vision_properties(self, raw_data: List, properties_list: List[Dict], input_type: PropertiesInputType):
+        """Calculate and cache the properties for the batch according to the propety input type"""
+        properties_list = validate_properties(properties_list)
+        # if there are no cached properties at all, calculate all the properties on the list,
+        # else calculate only those that were not yet calculated.
+        dataset = self._context.get_data_by_kind(self._dataset_kind)
+        if self._vision_properties_cache[input_type.value] is None:
+            self._vision_properties_cache[input_type.value] = dataset.calc_properties(raw_data, properties_list)
+        else:
+            properties_to_calc = [p for p in properties_list if p['name'] not in
```
noamzbr, line 108: Not an important comment, but sounds like you can use only this line if you initialized _vision_properties_cache to be an empty dict per input type rather than None

deepchecks/vision/checks/data_integrity/abstract_property_outliers.py
```diff
@@ -60,48 +61,39 @@ class AbstractPropertyOutliers(SingleDatasetCheck):
     """
 
     def __init__(self,
-                 properties: t.List[t.Dict[str, t.Any]] = None,
+                 properties_list: t.List[t.Dict[str, t.Any]] = None,
+                 property_input_type: PropertiesInputType = PropertiesInputType.OTHER,
```
noamzbr, line 65: Maybe don't put a default value / comment that this must be replaced in the inheriting checks?

deepchecks/vision/checks/data_integrity/image_property_outliers.py
```diff
@@ -47,11 +48,13 @@ class ImagePropertyOutliers(AbstractPropertyOutliers):
 
     def __init__(self,
                  image_properties: t.List[t.Dict[str, t.Any]] = None,
+                 property_input_type: PropertiesInputType = PropertiesInputType.IMAGES,
```
noamzbr, line 51: Need to update also the docstrings in the inheriting checks. Missing docstring for this parameter

deepchecks/vision/checks/data_integrity/label_property_outliers.py
```diff
@@ -50,11 +51,13 @@ class LabelPropertyOutliers(AbstractPropertyOutliers):
 
     def __init__(self,
                  label_properties: t.List[t.Dict[str, t.Any]] = None,
+                 property_input_type: PropertiesInputType = PropertiesInputType.LABELS,
```
noamzbr, line 54: same here need docstring

deepchecks/vision/checks/model_evaluation/image_segment_performance.py
noamzbr: I tend to agree with @nirhutnik. It is more elegant the way you did it (no need to call validation per check init), but the fact is that properties **can** be validated in the init, and from a user experience perspective it's better - If my properties are broken, I'll know that when creating the check rather than when running it.

deepchecks/vision/vision_data.py
```diff
@@ -224,6 +225,25 @@ def batch_to_images(self, batch) -> Sequence[np.ndarray]:
         """
         raise DeepchecksNotImplementedError('batch_to_images() must be implemented in a subclass')
 
+    def calc_properties(self, raw_data: List = None, properties_list: List[Dict] = None) -> Dict[str, list]:
```
noamzbr, line 228: Here lies my biggest concern - how do override this method in a monitoring context to retrieve the appropriate properties? Because this method receives the raw data, it has no idea what is the index of the image, and so no way to query the properties for the right image from the DB...
@TheSolY @ItayGabbay Am I right about this?

noamzbr (commented):
My largest concern is with the comment I wrote for the calc_properties method, if I'm wrong there than this looks good. Other then that, the property user guide need to be updated to reflect these changes.

deepchecks/vision/vision_data.py
TheSolY: @noamzbr raw data can be any kind of list, not necessarily images, so it can also be a list of precalculated results with a single property that is the identity function.  As to managing indices in the batch_wrapper- I assumed we can implement it similar to _do_static_pred, am I wrong here?

deepchecks/vision/checks/data_integrity/abstract_property_outliers.py
TheSolY: @noamzbr why?

PR #1822, opened 2022-07-25
FI new function for users

### LINKED ISSUE #1757: [FEAT] Enable explicit control of FI calculation
We want the user to be able to use our feature importance calculation mechanism easily in two scenarios:
1. Passing calculated feature importance to a check/suite 
2. Calculating feature importance and exporting the FI Series to an external source

Currently if the user wants to control the FI calculation he can either pass a FI Series he calculated elsewhere, or modify one of the following parameters of the run() method:
```
        feature_importance_force_permutation: bool = False,
        feature_importance_timeout: int = 120,
```
We want the option for an intermediate flow - use something equivalent to our `calculate_feature_importance` method (perhaps with the permutation_kwargs explicitly stated), get an FI Series that can be inspected, and then pass it to the run() feature_importance arg.

Also add to docs

This creates a "two ways to do one thing" situation, so need to consider if there is a way to simplify.

*Additional alternative*: create a sub package with SHAP that can calculate way better FI compare to permutation importance.

nirhutnik:
Closes #1757 

Added function calculate_feature_importance in new file tabular/user_utils.py for users to use.
Changed the warnings for force_permutation=True so they will fail the process and we won't continue to built-in feature importance.
Added paragraph in feature_importance.rst.

NOTE!
Changed behavior of calculate_feature_importance in features.py - Now if force_permutation=True, and it fails to calculate permutation importance, the function raises an error (previous behavior skipped and tried model built-in importance, which is not intuitive and doesn't make sense IMO).

### DIFF
--- a/deepchecks/tabular/user_utils.py
+++ b/deepchecks/tabular/user_utils.py
@@ -15,21 +15,62 @@ import pandas as pd
 
 from deepchecks import tabular
 from deepchecks.core.errors import DeepchecksValueError
+from deepchecks.tabular.metric_utils import DeepcheckScorer
 from deepchecks.utils.features import calculate_feature_importance as _calculate_feature_importance
 
 __all__ = [
-    'calculate_feature_importance'
+    'feature_importance'
 ]
 
 
-def calculate_feature_importance(
+def feature_importance(
         model: t.Any,
         dataset: t.Union['tabular.Dataset', pd.DataFrame],
-        permutation_kwargs: t.Dict[str, t.Any] = None,
+        n_repeats: int = 30,
+        mask_high_variance_features: bool = False,
+        n_samples: int = 10_000,
+        alternative_scorer: t.Optional[DeepcheckScorer] = None,
+        random_state: int = 42
 ) -> pd.Series:
-    permutation_kwargs = permutation_kwargs or {}
-    permutation_kwargs['skip_messages'] = permutation_kwargs.get('skip_messages', True)
-    permutation_kwargs['timeout'] = permutation_kwargs.get('timeout', None)
+    """
+    Calculate feature importance outside of check and suite.
+
+    As many checks and suites use feature importance, it is rec
+
+    Parameters
+    ----------
+    model: t.Any
+        a fitted model
+    dataset: t.Union['tabular.Dataset', pd.DataFrame]
+        dataset used to fit the model
+    n_repeats: int, default: 30
+        Number of times to permute a feature
+    mask_high_variance_features : bool , default: False
+        If true, features for which calculated permutation importance values
+        varied greatly would be returned has having 0 feature importance
+    n_samples: int, default: 10_000
+        The number of samples to draw from X to compute feature importance
+        in each repeat (without replacement).
+    alternative_scorer: t.Optional[DeepcheckScorer], default: None
+        Scorer to use for evaluation of the model performance in the permutation_importance function. If not defined,
+        the default deepchecks scorers are used.
+    random_state: int, default: 42
+        Random seed for permutation importance calculation.
+
+    Returns
+    -------
+    pd.Series:
+        feature importance normalized to 0-1 indexed by feature names
+    """
+    permutation_kwargs = {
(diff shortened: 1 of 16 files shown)


deepchecks/tabular/user_utils.py
```diff
@@ -0,0 +1,41 @@
...
+def calculate_feature_importance(
+        model: t.Any,
+        dataset: t.Union['tabular.Dataset', pd.DataFrame],
+        permutation_kwargs: t.Dict[str, t.Any] = None,
+) -> pd.Series:
+    permutation_kwargs = permutation_kwargs or {}
+    permutation_kwargs['skip_messages'] = permutation_kwargs.get('skip_messages', True)
+    permutation_kwargs['timeout'] = permutation_kwargs.get('timeout', None)
+
+    if isinstance(dataset, pd.DataFrame):
```
Nadav-Barak, line 34: I this is the case, why are you allowing dataset to be t.Union['tabular.Dataset', pd.DataFrame]?

deepchecks/tabular/user_utils.py
```diff
@@ -0,0 +1,41 @@
...
+
+__all__ = [
+    'calculate_feature_importance'
+]
+
+
+def calculate_feature_importance(
+        model: t.Any,
+        dataset: t.Union['tabular.Dataset', pd.DataFrame],
+        permutation_kwargs: t.Dict[str, t.Any] = None,
```
Nadav-Barak, line 28: I really dont like the permutation_kwargs for external methods, consider accepting the different parameters separately.

tests/tabular/test_user_utils.py
```diff
@@ -0,0 +1,38 @@
...
+#
+"""Test user utils"""
+from hamcrest import assert_that, close_to, raises, calling
+from sklearn.neural_network import MLPClassifier
+
+from deepchecks.core.errors import DeepchecksValueError
+from deepchecks.tabular.user_utils import calculate_feature_importance
+
+
+def test_calculate_importance_when_no_builtin(iris_split_dataset_and_model):
```
Nadav-Barak, line 19: Add a test were you supply acceptable kwargs

deepchecks/tabular/user_utils.py
```diff
@@ -0,0 +1,41 @@
...
+) -> pd.Series:
+    permutation_kwargs = permutation_kwargs or {}
```
Nadav-Barak, line 30: doc string

JKL98ISR:
can you give a better name for the pr?

deepchecks/utils/features.py
```diff
@@ -150,15 +150,12 @@ def calculate_feature_importance(
 
     if force_permutation:
         if isinstance(dataset, pd.DataFrame):
-            permutation_failure = 'Cannot calculate permutation feature importance on a pandas Dataframe, using ' \
```
JKL98ISR, line 153: why remove support for df?

nirhutnik: It's removing the behavior that if you mark "force_permutation" then it later looks for built-in feature importance. Instead, it fails.

JKL98ISR (approved):

tests/tabular/test_feature_importance_user_utils.py
```diff
@@ -0,0 +1,49 @@
...
+# ----------------------------------------------------------------------------
+#
+"""Test user utils"""
+from hamcrest import assert_that, close_to, raises, calling
+
+from deepchecks.core.errors import DeepchecksValueError
+from deepchecks.tabular.feature_importance import calculate_feature_importance
+
+
+def test_calculate_importance(iris_split_dataset_and_model):
```
Nadav-Barak, line 18: Minor and optional: Maybe in one of the test try using an alternative scorer

Nadav-Barak (approved):
Looks good!
Additional suggestion: maybe show case this option in some places in docs (specifically plots that calculate FI multiple times), see comment in board issue

ItayGabbay (approved):

PR #1864, opened 2022-08-01
1706 static properties

TheSolY:
Static (precalculated) properties can be passed to the checks' run similarly to static predictions. 
The properties are also cached in the Batch object after the first they're calculated to save time in suites.


deepchecks/vision/checks/data_integrity/abstract_property_outliers.py
```diff
@@ -50,6 +52,8 @@ class AbstractPropertyOutliers(SingleDatasetCheck):
         - 'categorical' - for discrete, non-ordinal outputs. These can still be numbers,
           but these numbers do not have inherent value.
         For more on image / label properties, see the :ref:`property guide </user-guide/vision/vision_properties.rst>`
+    property_input_type: PropertiesInputType, default: PropertiesInputType.OTHER
```
noamzbr, line 55: Now that I think about it - in what case will the user want to modify the property_input_type of the check? (of the inheriting check, not this abstract class).
I mean, for example the image property outliers always passes the images to the property, so modifying this argument will just lead to the properties not to be listed under then right section...

deepchecks/vision/utils/vision_properties.py
```diff
@@ -0,0 +1,142 @@
...
+        warnings.warn(
+            f'Property Warnings:\n+ {concatenated_warnings}',
+            category=DeprecationWarning
+        )
+
+    return properties
+
+
+# pylint: disable=invalid-name
+STATIC_PROPERTIES_FORMAT = Dict[int, Dict[PropertiesInputType, Dict[str, Any]]]
```
noamzbr, line 122: Just making sure - dataset indices are always int?

tests/vision/checks/model_evaluation/train_test_prediction_drift_test.py
```diff
@@ -250,33 +250,3 @@ def test_condition_pass(mnist_dataset_train, mock_trained_mnist, device):
     ))
 
 
-def test_with_drift_object_detection_defected_alternative_measurements():
```
noamzbr, line 253: Why delete these tests?

tests/vision/checks/train_test_validation/image_property_drift_test.py
```diff
@@ -66,24 +66,6 @@ def test_image_property_drift_check_limit_classes_illegal(coco_train_visiondata,
     )
 
 
-def test_image_property_drift_initialization_with_empty_list_of_image_properties():
```
noamzbr, line 69: Same, why delete these tests?

tests/vision/checks/train_test_validation/train_test_label_drift_test.py
```diff
@@ -266,32 +266,3 @@ def prop(labels):
         )
         }
     ))
-
-
-def test_with_drift_object_detection_defected_alternative_properties():
```
noamzbr, line 271: Also here

tests/vision/utils_tests/properties_utils_tests.py
```diff
@@ -9,113 +9,92 @@
...
-            rf"instead got hello-world")
+            r"\+ Property #1: dictionary must include keys \('name', 'method', 'output_type'\)\. "
+            r"Next keys are missed \['name'\]")
     )
 
-
-# =====================================================
-
-
-def test_label_prediction_properties_validation():
```
noamzbr, line 76: Why delete these?

deepchecks/vision/utils/vision_properties.py
```diff
@@ -0,0 +1,142 @@
...
+
+
+class PropertiesInputType(Enum):
+    """Enum containing supported task types."""
+
+    IMAGES = 'images'
+    BBOXES = 'bounding_boxes'
+    LABELS = 'labels'
+    PREDICTIONS = 'predictions'
+    OTHER = 'other'
```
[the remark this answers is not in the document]
TheSolY: @noamzbr a. Whatever the user wants- segmentation mask, a combination of image and label, metadata, etc.
b. I used it as a default for AbstractPropertyOutliers

tests/vision/checks/model_evaluation/train_test_prediction_drift_test.py
TheSolY: @noamzbr the validation moved to the batch_wrapper so the tests moved to properties_utils_test

deepchecks/vision/utils/vision_properties.py
TheSolY: yes

deepchecks/vision/checks/data_integrity/abstract_property_outliers.py
TheSolY: @noamzbr I don't have an example use case but IMO it's neater that this class is in the same format of the child classes

deepchecks/vision/utils/vision_properties.py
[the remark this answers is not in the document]
noamzbr: But for each of these the check needs to know what to pass to the property...

tests/vision/checks/model_evaluation/train_test_prediction_drift_test.py
noamzbr: But I don't see that properties_utils_test tests things like alternative measurements

TheSolY: test_validate_properties_with_incorrect_property_dict_structure
test_validate_properties_with_bad_name_field
test_validate_properties_with_incorrect_property_output_type

TheSolY: actually, most of the tests there test different types of defected alternative measurements

deepchecks/vision/utils/vision_properties.py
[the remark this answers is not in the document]
TheSolY: @noamzbr IMO it's good to give the user this kind of flexibility but if you believe it's critical I'll delete this option

deepchecks/vision/checks/data_integrity/abstract_property_outliers.py
```diff
@@ -50,6 +52,8 @@ class AbstractPropertyOutliers(SingleDatasetCheck):
         - 'categorical' - for discrete, non-ordinal outputs. These can still be numbers,
```
Nadav-Barak, line 52: Semi related question, why are properties not just a dict of name: callable?
More specifically would love help understanding the usage of output_type
@yromanyshyn (git say you originally added this)

tests/vision/base/test_static_properties.py
```diff
@@ -0,0 +1,97 @@
...
+                                    indexes})
+
+        else:
+            static_prop = None
+        static_props.append(static_prop)
+    train_prop, tests_prop = static_props
+    return train_prop, tests_prop
+
+
+def test_image_properties_outliers(mnist_dataset_train, mnist_dataset_test):
```
Nadav-Barak, line 67: Can add a test that compares the results of passing static prediction to run vs passing a callable to init an check the results are the same

deepchecks/vision/checks/data_integrity/abstract_property_outliers.py
noamzbr: Actually IMO it's @nirhutnik

noamzbr (approved):

deepchecks/vision/checks/data_integrity/abstract_property_outliers.py
nirhutnik: output_type is for the display - 
The display need to know if to plot a KDE graph (numerical) or barchart (categorical). The "class_id" type exists so that we can map the label values (usually integers) to the actual names of the class (using VisionData._label_map)

Nadav-Barak: Didn't understand :/ lets talk tomorrow before I start working on NLP properties

nirhutnik: We display the distribution of each property in checks like PropertyDrift.
For numerical properties, we show a continuous graph:
image

For categorical properties, we show a bar graph:
image

Also, notice that it shows the correct class name in the x-axis.

The only way for the check to know which type of graph to show (and which type of drift to calculate) is by telling it which type of variable it is (numerical/categorical).

Better explanation?

ItayGabbay (approved):

PR #1884, opened 2022-08-08
Fix dead links for quickstarts

dnth:

This PR fixes 3 dead links (Quickstarts) on the readme page.


ItayGabbay:
@dnth Thanks for your contribution! good catch.
You got us in a middle of a release process so the links in the readme were updated already but did not exist in the docs yet 😅 
This will be fixed in a few moments since a new docs version is on its way.
Closing this because of that, but thanks again!

PR #1885, opened 2022-08-08
[Issue-1739] Suite/check instances serialization/deserialization to/from config

### LINKED ISSUE #1739: [BUG] need to reconsider how we get check param
**Describe the bug**
we get the check param by the variables defined by the check and not by what has been passed to it
also some params are functions or classes (such as metric) which can cause problems in serialization
 
**Expected behavior**
get the params the check was initialized with

**Screenshots**
Screen Shot 2022-07-07 at 10 26 42

yromanyshyn:
resolves #1739 

to summarise

this PR is the continuation of this [issue](https://github.com/deepchecks/mon/issues/20) (initial [PR](https://github.com/deepchecks/deepchecks/pull/1736))

previously we added `config` and `from_config` methods to the `BaseCheck` and `BaseSuite` types. 
The purpose of those methods is to be able:
- to transform (`flatten`) check/suite instance into a dictionary of simple (JSON/YAML serialisable) values that represent it`s state
- to restore check/suite instance from a dictionary of values

default implementations of those methods will make the next assumptions:

**point 1.**

check type `__init__` method signature represents it's instances internal state. This means that each parameter passed to the `__init__` is stored within an instance under the same name, like

```python
class Foo:
   def __init__(self, x, y, c):
       self.x = x
       self.y = y
       self.c = c
```

It is true for most of our checks but not for all (example `IndexTrainTestLeakage`, `SegmentPerformance`, ...)
To not do any implicit magical stuff the simplest solution for those checks for which the statement  above is not true is to override the `config()` method

**point 2.**

`BaseCheck.config` and `BaseCheck.from_config` make an assumption that the internal state of a check is represented by simple JSON/YAML serializable values, moreover, they expect that serialization/deserialization will not make those values to lose type information that might cause a restored check instance to fail. Example:

```python
t = (1,2,3)
s = json.dumps((1,2,3))

# now it is a list and if the check behaviour
# dependent on the fact that it was a tuple
# then recreated check will probably fail
l = json.loads(s)
```

Again, if the above is not true for some checks then they must override the `config` and `from_config` methods

**Notes:**

Initially, I thought about using `jsonpickle` for config serialization/deserialization but it is not secure and might execute arbitrary code during deserialization, therefore its usage within the monitoring package is unacceptable.

Currently, I do not see a way to support secure serialization/deserialization of  complex values like lambdas (see `SegmentPerformance.alternative_scorer`) In cases like this check type must override the `config` and `from_config` methods, and either raise an exception or replace those values with some default one.


yromanyshyn:
@ItayGabbay  @JKL98ISR 

UPDATE:

I overrode the `config` method for those check types that require it.
Currently, only for those which implement the `reduce_output` method, if you 
do not have any remarks about the solution I will do the same for others check types

deepchecks/core/common.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+def importable_name(obj: t.Any) -> str:
+    kind = type(obj) if not isinstance(obj, type) else obj
+    name = kind.__qualname__
+    module = kind.__module__
+    return f'{name}.{module}'
+
+
+def import_type(full_name: str, base: t.Optional[t.Type[t.Any]] = None):
```
matanper, line 19: why do we need "base" if we have full name?

deepchecks/vision/checks/model_evaluation/train_test_prediction_drift.py
```diff
@@ -229,6 +229,30 @@ def compute(self, context: Context) -> CheckResult:
...
+                if 'method' not in prop or not prop['method']:
+                    raise ValueError('Each prediction property is expected to contain not emtpy "method" key')
+                if callable(prop['method']):
+                    type_name = type(self).__name__
+                    raise ValueError(
+                        f'Serialization of "{type_name}" check instance is not supported '
+                        'if custom user defined properties were passed to the "prediction_properties" '
+                        f'parameter during instance initialization. Property name: {prop["name"]}'
+                    )
+        return self._prepare_config(
```
matanper, line 244: why not just call the parent `config` if the validation passed?

deepchecks/__init__.py
```diff
@@ -9,13 +9,19 @@
...
 import warnings
 from importlib._bootstrap import _init_module_attrs
 
+try:
+    from importlib.metadata import version
+except ImportError:
+    from importlib_metadata import version
+    
+# NOTE: it is here, before other import, in order to omit circular import error
+__version__ = version('deepchecks')
```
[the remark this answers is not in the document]
yromanyshyn: hm/

for users, probably not, because `version` function expects that the given package was installed as a library (via pip)

it might fail for you if you have not done `pip install -e .` and expects that `deepchecks` will be in the `sys.path`

deepchecks/core/checks.py
```diff
@@ -48,8 +52,8 @@ class CheckMetadata(TypedDict):
 
 
 class CheckConfig(TypedDict):
-    class_name: str
-    module_name: Optional[str]
+    kind: str
+    version: NotRequired[str]
```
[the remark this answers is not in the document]
yromanyshyn: please take a look at the spec that added it - [link](https://peps.python.org/pep-0655/)

deepchecks/core/common.py
yromanyshyn: it has totally different purpose/
it helps to verify that what we had imported is actually a subtype of `BaseCheck`/`BaseSuite`

deepchecks/vision/checks/model_evaluation/train_test_prediction_drift.py
yromanyshyn: _**point 1**_ from the description
the `prediction_properties` parameter is stored under a different name within the instance

deepchecks/core/checks.py
```diff
@@ -168,37 +172,83 @@ def metadata(self, with_doc_link: bool = False) -> CheckMetadata:
...
+    def from_json(
+        self,
+        conf: str,
+        version_unmatch: Union[L['raise'], L['warn'], None] = 'warn'
+    ) -> Self:
+        """Deserialize check instance from JSON string."""
+        check_conf = json.loads(conf)
+        return self.from_config(check_conf, version_unmatch=version_unmatch)
+
+    def _prepare_config(
```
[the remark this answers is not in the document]
yromanyshyn: to separate config preparation logic from parameter values preparation logic

I do not want the check types that override `config` method to import `deepchecks.core.common` module and to handle other config dictionary values

deepchecks/vision/checks/model_evaluation/train_test_prediction_drift.py
matanper: maybe we should change that

yromanyshyn: what exactly?

matanper: this
> the prediction_properties parameter is stored under a different name within the instance

deepchecks/__init__.py
```diff
@@ -9,13 +9,19 @@
...
-import pathlib
 import sys
 import types
 import warnings
 from importlib._bootstrap import _init_module_attrs
 
+try:
+    from importlib.metadata import version
+except ImportError:
+    from importlib_metadata import version
```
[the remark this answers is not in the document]
yromanyshyn: added

deepchecks/vision/checks/model_evaluation/train_test_prediction_drift.py
yromanyshyn: actually, somebody else did that and I did not notice after merge with the main

fixed everything else related

matanper: so we can just call the parent `config`?

matanper: oh you've done it, so I don't think the `_prepare_config` is needed anymore, but won't insist on this

matanper (approved):

matanper (approved):

JKL98ISR:
can you just add a screenshot/print of an example of the new config structure with a check?

yromanyshyn:
@JKL98ISR 
> can you just add a screenshot/print of an example of the new config structure with a check?

image

image

image

JKL98ISR:
why you changed the format? now the check name is very hard to access, I prefer the old format where the name and module are seperated. also you remove the option that was before so that our checks won't use the full path

yromanyshyn:
> I prefer the old format where the name and module are seperated.

how are these two different?
why 2 keys are more convenient than one?

```
# first
"name": "IsSingleValue",
"module": "deepchecks.tabular.checks.data_integrity.is_single_value",

# second
"kind": "deepchecks.tabular.checks.data_integrity.is_single_value.IsSingleValue"
```

> also you remove the option that was before so that our checks won't use the full path

yeap, that a good point/
lets go even furthere and allow users to use `tabular.<check-name>` and `vision.<check-name>`?

so there will be two possibilities:
- full path to a check - `deepchecks.tabular.checks.data_integrity.is_single_value.IsSingleValue`
- short check path - `tabular.IsSingleValue` / `vision.ImagePropertyOutliers`

how do you think?

yromanyshyn:
@JKL98ISR 
> lets go even furthere and allow users to use tabular.<check-name> and vision.<check-name>?

**UPDATE:**

I suggest that shortcuts to be available only for users when they manually write/modify a config file
but our `config()` implementation still will be producing full-path names

deepchecks/__init__.py
```diff
@@ -9,13 +9,23 @@
...
 import sys
 import types
 import warnings
 from importlib._bootstrap import _init_module_attrs
 
+try:
+    from importlib.metadata import version
+except ImportError:
+    # NOTE:
+    # 'importlib.metadata' was added to python>=3.8
```
noamzbr, line 21: So maybe we need to add this library to our requirements?

JKL98ISR:
2 values are more comfortable in simple usage, no need to keep conventions and no need to fully understand the config.
name and module are more intuitive (also easier for UI proposes).

deepchecks/tabular/checks/model_evaluation/boosting_overfit.py
```diff
@@ -229,6 +232,19 @@ def condition(result: dict):
...
         return self.add_condition(name, condition)
 
+    def config(self, include_version: bool = True) -> 'CheckConfig':
+        """Return check instance config."""
+        if self.alternative_scorer is not None:
+            for k, v in self.alternative_scorer.items():
+                if not isinstance(v, str):
+                    name = type(self).__name__
+                    raise ValueError(
+                        f'Serialization of "{name}" check instance is not supported '
```
noamzbr, line 242: Who sees this message? In case it's the user, I suggest making it clearer what was wrong and what they should be doing. For example "Only recognized string scorers are allowed when sterilizing checks. For a list of recognized strings please refer to https://docs.deepchecks.com/en/stable/user-guide/general/metrics_guide.html#list-of-supported-strings"

deepchecks/__init__.py
yromanyshyn: I added 
`importlib_metadata>=1.4; python_version < '3.8'`

deepchecks/tabular/checks/model_evaluation/boosting_overfit.py
yromanyshyn: I am not against rephrasing error messages, so yeap will do that.

But I am not sure about links in the error message/ User always can do `help(CheckType)` to get a docstring of a check without going anywhere

yromanyshyn:
@JKL98ISR 

> 2 values are more comfortable in simple usage, no need to keep conventions and no need to fully understand the config.
name and module are more intuitive (also easier for UI proposes).

as for me no need for two keys
but it is not so important - ok, I redid that part back

deepchecks/tabular/checks/model_evaluation/boosting_overfit.py
matanper: @yromanyshyn notice the link that @noamzbr sent is not tied directly to the check, so `help(CheckType)` won't get you there that easily

yromanyshyn: I rephrased error messages and added references to the docs

ItayGabbay (approved):
Approving the changed in .github folder

noamzbr (approved):

PR #1919, opened 2022-08-17
Vision Properties v2

JKL98ISR:
fixed a bug which causes static properties to always use only the first batch
made property calculation to use index
made label/prediction properties not mandatory if given static properties

### DIFF
--- a/deepchecks/vision/batch_wrapper.py
+++ b/deepchecks/vision/batch_wrapper.py
@@ -11,12 +11,10 @@
 """Contains code for BatchWrapper."""
 from operator import itemgetter
 from typing import TYPE_CHECKING, Any, Callable, Dict, Iterable, List, Tuple, TypeVar, cast
-import numpy as np
 
 import torch
 
 from deepchecks.core import DatasetKind
-from deepchecks.core.errors import DeepchecksProcessError
 from deepchecks.vision.task_type import TaskType
 from deepchecks.vision.utils.image_functions import crop_image
 from deepchecks.vision.utils.vision_properties import (PropertiesInputType, calc_vision_properties,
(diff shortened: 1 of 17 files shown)


deepchecks/vision/batch_wrapper.py
```diff
@@ -103,32 +103,52 @@ def _do_static_prop(self):
...
+            for label in labels:
+                label = label.cpu().detach().numpy()
+                bbox = label[1:]
+                # make sure image is not out of bounds
+                if round(bbox[2]) + min(round(bbox[0]), 0) <= 0 or round(bbox[3]) <= 0 + min(round(bbox[1]), 0):
+                    continue
+                imgs.append(crop_image(img, *bbox))
+        return imgs
+
+    def _get_relevant_data(self, input_type: PropertiesInputType):
```
nirhutnik, line 122: name of function too general. Maybe add "for properties"?

deepchecks/vision/context.py
```diff
@@ -132,9 +132,9 @@ def __init__(
                                                          [train_properties, test_properties]):
                 if dataset is not None:
                     try:
-                        props = itemgetter(*list(dataset.data_loader.batch_sampler)[0])(properties)
+                        # TODO: add validation
```
nirhutnik, line 135: Well did you add? :)

deepchecks/vision/utils/vision_properties.py
```diff
@@ -138,6 +139,9 @@ def static_prop_to_cache_format(static_props: STATIC_PROPERTIES_FORMAT) -> PROPE
 
     for input_type in input_types:
         for prop_name in list(static_props[indices[0]][input_type].keys()):
-            props_cache[input_type][prop_name] = [static_props[index][input_type][prop_name] for index in indices]
+            prop_vals = [static_props[index][input_type][prop_name] for index in indices]
+            if input_type == PropertiesInputType.BBOXES.value:
```
nirhutnik, line 143: Add an inline comment that explains why for bboxes we flatten the output.

nirhutnik: Also, didn't you change it to "partial_images"?

nirhutnik (approved):

ItayGabbay (approved):

PR #1920, opened 2022-08-17
feature label correlation change display fixed

TheSolY:
Switched between the train and test colors and set the font to a fixed size

### DIFF
--- a/deepchecks/utils/plot.py
+++ b/deepchecks/utils/plot.py
@@ -21,8 +21,8 @@ __all__ = ['create_colorbar_barchart_for_check', 'shifted_color_map',
            'create_confusion_matrix_figure', 'colors', 'hex_to_rgba']
 
 
-colors = {'Train': '#00008b',  # dark blue
-          'Test': '#69b3a2',
+colors = {'Train': '#69b3a2',  # green
+          'Test': '#00008b',   # dark blue
           'Baseline': '#b287a3',
           'Generated': '#2191FB'}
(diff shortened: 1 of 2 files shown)


TheSolY:
Example for before and after the fix 
Before:
image
After:
image

deepchecks/utils/plot.py
```diff
@@ -21,8 +21,8 @@
            'create_confusion_matrix_figure', 'colors', 'hex_to_rgba']
 
 
-colors = {'Train': '#00008b',  # dark blue
-          'Test': '#69b3a2',
+colors = {'Train': '#69b3a2',  # green
+          'Test': '#00008b',   # dark blue
```
nirhutnik, line 25: Why?
Seems confusing to users already using the package...

TheSolY: @nirhutnik because the "(" is seen more clearly on the green background where the text is wider than the bar (see the middle pair of bars int he example)

nirhutnik: But this changes all of the graphs... Seems Like an excessive solution. Maybe just using a lighter shade of blue?

shir22: We talked about it in the daily, no?

TheSolY: @nirhutnik the color definitions are global anyway

noamzbr:
Decided to resolve issue with https://github.com/deepchecks/deepchecks/pull/1924 instead

PR #1950, opened 2022-08-25
make unused feature a single dataset check

### LINKED ISSUE #1927: [BUG] unused features check shouldn't be TrainTestCheck
**Describe the bug**
It only uses one dataset - old code

Screen Shot 2022-08-18 at 17 05 51

we just need to make it a single dataset check

JKL98ISR:
closes #1927

### DIFF
--- a/deepchecks/tabular/feature_importance.py
+++ b/deepchecks/tabular/feature_importance.py
@@ -47,7 +47,7 @@ def calculate_feature_importance(
     >>> from deepchecks.tabular.feature_importance import calculate_feature_importance
     >>> from deepchecks.tabular.datasets.classification.iris import load_data, load_fitted_model
     >>> from deepchecks.tabular.checks import UnusedFeatures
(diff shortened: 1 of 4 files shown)


matanper (approved):

deepchecks/tabular/feature_importance.py
```diff
@@ -50,8 +50,7 @@ def calculate_feature_importance(
     >>> iris_train_dataset, iris_test_dataset = load_data()
```
nirhutnik, line 50:
```suggestion
    >>> _, iris_test_dataset = load_data()
```

nirhutnik (approved):

ItayGabbay (approved):

deepchecks/tabular/feature_importance.py
JKL98ISR: it is used

Issue #1820, opened 2022-07-25
[FEAT] Help user avoid misclassification of task type


shir22:
**Is your feature request related to a problem? Please describe.**
When a Dataset is created, task type is inferred using the label.
Sometimes this is done incorrectly, e.g. see #1814 (inferred as multiclass classification instead of regression), and may result in weird result values.

**Describe the solution you'd like**
1. For multiclass of integers, add a warning when inferring task type that we found x values of the label and interpreted them as multiclass, but they could be regression as well, with a code snippet of how to define this.
2. For checks that use the task type and the prediction value (such as the Train Test Prediction Drift, but not limited ot it) add a validation that the prediction value indeed fit the defined task type, in if they don't - check should return an error or warning, with explanation how to define task type to avoid this issue. (otherwise we'll be displaying meaningless results on predictions... whether it's drift calculation, model performance metrics, etc.)

DanBasson:
@nirhutnik can you please assign this issue to me?

nirhutnik:
> @nirhutnik can you please assign this issue to me?

@DanBasson Done :)

PR #1953, opened 2022-08-27
feature: Task type warning for multiclass label containing integer

DanBasson:
Reference Issues/PRs
Fixes https://github.com/deepchecks/deepchecks/issues/1820

What does this implement/fix? Explain your changes.
This address the first section in the issue.
I've added a warning when inferring task type as multiclass if we found more than 10 unique values in the label and
the label contains only integers.

Any other comments?
i would appreciate your comments.
if i'm in the right direction, will move to the second section.

### DIFF
--- a/deepchecks/tabular/checks/model_evaluation/train_test_prediction_drift.py
+++ b/deepchecks/tabular/checks/model_evaluation/train_test_prediction_drift.py
@@ -21,6 +21,7 @@ from deepchecks.core.errors import DeepchecksValueError
 from deepchecks.core.reduce_classes import ReduceMixin
 from deepchecks.tabular import Context, TrainTestCheck
 from deepchecks.tabular.utils.task_type import TaskType
+from deepchecks.utils.logger import get_logger
 from deepchecks.utils.distribution.drift import (SUPPORTED_CATEGORICAL_METHODS, SUPPORTED_NUMERIC_METHODS,
                                                  calc_drift_and_plot, get_drift_plot_sidenote)
 
@@ -172,6 +173,13 @@ class TrainTestPredictionDrift(TrainTestCheck, ReduceMixin):
                 train_prediction = train_prediction[:, [1]]
                 test_prediction = test_prediction[:, [1]]
         else:
+            if context.task_type == TaskType.MULTICLASS:
+                get_logger().warning(
+                    'deepchecks infers label to be multiclass and auto plot class prediction drift. '
+                    'It is recommended to plot prediction probabilities instead, in order to '
+                    'produce meaningful results. To do so, add "drift_mode=\'proba\'" '
+                    'to TrainTestPredictionDrift check'
+                )
             train_prediction = np.array(model.predict(train_dataset.features_columns)).reshape((-1, 1))
             test_prediction = np.array(model.predict(test_dataset.features_columns)).reshape((-1, 1))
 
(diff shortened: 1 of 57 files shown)


deepchecks/tabular/dataset.py
```diff
@@ -535,9 +535,17 @@ def train_test_split(self: TDataset,
         return self.copy(train_df), self.copy(test_df)
 
     @staticmethod
-    def _infer_label_type(label_col: pd.Series):
+    def _infer_label_type(label_col: pd.Series, infer_task_type_from_label_nunique: int = 10):
```
nirhutnik, line 538: As this attribute is used for raising a warning only, there's not need for it to be a parameter of the function.

deepchecks/tabular/dataset.py
```diff
@@ -535,9 +535,17 @@ def train_test_split(self: TDataset,
...
         if is_categorical(label_col, max_categorical_ratio=0.05):
             if label_col.nunique(dropna=True) > 2:
+                if infer_dtype(label_col) == 'integer' \
+                        and label_col.nunique() > infer_task_type_from_label_nunique:
+                    get_logger().warning(
+                        'Attributes such as "label_type" are not mandatory, but in a case of ordinal integers, '
```
nirhutnik, line 544: Would rephrase the warning here to be clearer on what happened, and how to fix it.
Suggesting something like:
"Integer label has many unique values. In this case, deepchecks automatically infers label to be multiclass. If your label is a regression label and not multiclass, please use "label_type='regression_label'" when initializing your Dataset."

DanBasson:
Hi @nirhutnik
Do you know why am i getting this failures?

nirhutnik:
@DanBasson yes, one was a small PR test that checks if the PR has a label. Solved that for you by adding one :)
The other test that fails, though it's not mandatory, is the coverages test that checks if this change causes less coverage. If you'll add a test for that warning you're adding, it would be ok

deepchecks/tabular/checks/model_evaluation/train_test_prediction_drift.py
```diff
@@ -172,6 +173,13 @@ def run_logic(self, context: Context) -> CheckResult:
                 train_prediction = train_prediction[:, [1]]
                 test_prediction = test_prediction[:, [1]]
         else:
+            if context.task_type == TaskType.MULTICLASS:
+                get_logger().warning(
+                    'deepchecks infers label to be multiclass and auto plot class prediction drift. '
+                    'It is recommended to plot prediction probabilities instead, in order to '
+                    'produce meaningful results. To do so, add "drift_mode=\'proba\'" '
+                    'to TrainTestPredictionDrift check'
```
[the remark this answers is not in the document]
nirhutnik: I know that in #1820 there was a request for a validation of some sort, but this seems unnecessary, and I tried to discuss this with others and we can't currently find how this could cause confusion that the other warning you added doesn't cover.

deepchecks/tabular/dataset.py
```diff
@@ -546,6 +546,13 @@ def train_test_split(self: TDataset,
     def _infer_label_type(label_col: pd.Series):
         if is_categorical(label_col, max_categorical_ratio=0.05):
             if label_col.nunique(dropna=True) > 2:
+                if infer_dtype(label_col) == 'integer' \
+                        and label_col.nunique() > 10:
```
nirhutnik, line 550: I think that 10 is too large. There's an issue here as some problems (discrete ordinal features) can actually be tackled both as regression and as multiclass (e.g. 1-5 stars in reviews).
So maybe >=5?

deepchecks/tabular/checks/model_evaluation/train_test_prediction_drift.py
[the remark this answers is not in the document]
DanBasson: i agree, will change it

deepchecks/tabular/dataset.py
DanBasson: sure

nirhutnik (approved):

ItayGabbay (approved):

ItayGabbay:
@DanBasson thanks for your contribution!

PR #1955, opened 2022-08-28
Changed gower distance to use actual given samples instead of only indices.

nirhutnik:
Changed gower distance to use actual given samples instead of only indices.

### DIFF
--- a/deepchecks/utils/gower_distance.py
+++ b/deepchecks/utils/gower_distance.py
@@ -56,7 +56,7 @@ def gower_matrix(data: np.ndarray, cat_features: np.array) -> np.ndarray:
 
 
 def calculate_nearest_neighbors_distances(data: pd.DataFrame, cat_cols: List[Hashable], numeric_cols: List[Hashable],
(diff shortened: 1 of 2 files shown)


Nadav-Barak (approved):

deepchecks/utils/gower_distance.py
```diff
@@ -77,27 +77,26 @@ def calculate_nearest_neighbors_distances(data: pd.DataFrame, cat_cols: List[Has
     num_neighbors: int
         Number of neighbors to return. For example, for n=2 for each sample returns the distances to the two closest
         samples in the dataset.
-    indices_to_calc_neighbors_for: List[Any], default None
-        List of indexes in data to calculate its nearest neighbors. If None, calculates for all given data samples.
+    samples_to_calc_neighbors_for: List[Any], default None
```
JKL98ISR, line 80: update that it is a dataframe now

JKL98ISR (approved):
kinda weird to have this parameter imo if it is never used in the code but looks like it works

Issue #1942, opened 2022-08-23
Unify the API for passing number of samples to the tabular checks


TheSolY:
Similar to what we did for Vision: https://github.com/deepchecks/deepchecks/issues/1863
Currently it's sometime passed to the init, for example in feature_feature_correlation has a parameter n_samples, and sometime to the run.
Should be passed only to the run in on all of the tabular checks.

noamzbr:
Something we need to address - currently different check have a different default number of samples (and rightly so, integrity checks can and should run on more samples). This will go away if we implement this.

noamzbr:
Decision: Add n_samples init argument to all tabular checks

noamzbr:
Closed by https://github.com/deepchecks/deepchecks/issues/2013

PR #1958, opened 2022-08-28
1900 nlp tf idf drift check PR draft

TheSolY:
basic functionality

### DIFF
--- a/deepchecks/nlp/checks/train_test_validation/keyword_frequency_drift.py
+++ b/deepchecks/nlp/checks/train_test_validation/keyword_frequency_drift.py
@@ -95,23 +95,26 @@ class KeywordFrequencyDrift(TrainTestCheck):
         else:
             raise DeepchecksValueError('top_n_method must be one of: top_diff, top_freq or a list of keywords')
 
-        self.top_n_words = np.take(np.array(vocab), top_n_idxs)
-        self.top_n_diffs = np.take(word_freq_diff, top_n_idxs)
+        top_n_words = np.take(np.array(vocab), top_n_idxs)
+        top_n_diffs = np.take(word_freq_diff, top_n_idxs)
 
         if context.with_display:
             train_to_show = max_train_freqs[top_n_idxs]
             test_to_show = max_test_freqs[top_n_idxs]
-            display = word_counts_drift_plot(train_to_show, test_to_show, self.top_n_words)
+            display = word_counts_drift_plot(train_to_show, test_to_show, top_n_words)
         else:
             display = None
-        return CheckResult(value=drift_score, display=display, header='Keyword Frequency Drift')
+
+        result = {'drift_score': drift_score, 'top_n_diffs': dict(zip(top_n_words, top_n_diffs))}
+        return CheckResult(value=result, display=display, header='Keyword Frequency Drift')
 
 
     def add_condition_drift_score_less_than(self, threshold: float):
         """
         Add condition - require drift score to be less than the threshold.
         """
-        def condition(drift_score) -> ConditionResult:
+        def condition(value) -> ConditionResult:
+            drift_score = value['drift_score']
             if drift_score < threshold:
                 details = f'The drift score {format_number(drift_score)} is less than the threshold {format_number(threshold)}'
                 return ConditionResult(ConditionCategory.FAIL, details)
@@ -119,3 +122,10 @@ class KeywordFrequencyDrift(TrainTestCheck):
                 details = f'The drift score {format_number(drift_score)} is not less than the threshold {format_number(threshold)}'
                 return ConditionResult(ConditionCategory.FAIL, details)
         return self.add_condition(f'Drift Score is Less Than {format_number(threshold)}', condition)
+
+    def add_condition_top_n_differences_less_than(self, threshold: float):
+        """
+        Add condition - require the absolute differences between the counts of train and the test to be less than the
+        threshold for all of the top n keywords.
+        """
+        pass
--- a/deepchecks/nlp/checks/train_test_validation/keyword_frequency_drift.py
+++ b/deepchecks/nlp/checks/train_test_validation/keyword_frequency_drift.py
@@ -9,10 +9,14 @@
 # ----------------------------------------------------------------------------
 #
 """Module containing the keyword drift check."""
+from deepchecks import CheckResult
+from deepchecks.core import DatasetKind
 from deepchecks.nlp.base_checks import TrainTestBaseCheck
-from typing import Union, List
+from typing import Union, List, Any
+from deepchecks.nlp.context import Context
 
 __all__ = ['KeywordFrequencyDrift']
+
 
 class KeywordFrequencyDrift(TrainTestBaseCheck):
     """
@@ -40,3 +44,9 @@ class KeywordFrequencyDrift(TrainTestBaseCheck):
         self.top_n_to_show = top_n_to_show
         self.top_n_method = top_n_method
         self.drift_method = drift_method
+
+    def run_logic(self, context: Context) -> CheckResult:
+        pass
+    
+
+
--- a/deepchecks/nlp/checks/train_test_validation/keyword_frequency_drift.py
+++ b/deepchecks/nlp/checks/train_test_validation/keyword_frequency_drift.py
@@ -81,19 +81,19 @@ class KeywordFrequencyDrift(TrainTestCheck):
         all_data = tokenized_train + tokenized_test
 
         vectorizer = TfidfVectorizer(input='content', strip_accents='ascii', tokenizer=_identity_tokenizer, min_df=2,
-                                     preprocessor=_identity_tokenizer, binary=True)
+                                     preprocessor=_identity_tokenizer)
         vectorizer.fit(all_data)
         train_freqs = vectorizer.transform(tokenized_train)
-        mean_train_freqs = np.array(train_freqs.mean(axis=0)).reshape(-1)
+        max_train_freqs = np.array(train_freqs.max(axis=0).todense()).reshape(-1)
 
         test_freqs = vectorizer.transform(tokenized_test)
-        mean_test_freqs = np.array(test_freqs.mean(axis=0)).reshape(-1)
-        word_freq_diff = np.abs(mean_train_freqs - mean_test_freqs)
+        max_test_freqs = np.array(test_freqs.max(axis=0).todense()).reshape(-1)
+        word_freq_diff = np.abs(max_train_freqs - max_test_freqs)
 
-        mean_train_counts = mean_train_freqs * test_dataset.n_samples
-        mean_test_counts = mean_test_freqs * test_dataset.n_samples
+        max_train_counts = max_train_freqs * test_dataset.n_samples
+        max_test_counts = max_test_freqs * test_dataset.n_samples
 
-        drift_score = self.drift_method(mean_train_counts, mean_test_counts, from_freqs=True)
+        drift_score = self.drift_method(max_train_counts, max_test_counts, from_freqs=True)
         vocab = vectorizer.get_feature_names_out()
 
         if isinstance(self.top_n_method, List):
@@ -101,7 +101,7 @@ class KeywordFrequencyDrift(TrainTestCheck):
         elif self.top_n_method == 'top_diff':
             top_n_idxs = np.argsort(word_freq_diff)[-self.top_n_to_show:]
         elif self.top_n_method == 'top_freq':
-            max_freqs = np.maximum(mean_train_freqs, mean_test_freqs)
+            max_freqs = np.maximum(max_train_freqs, max_test_freqs)
             top_n_idxs = np.argsort(max_freqs, )[-self.top_n_to_show:]
         else:
             raise DeepchecksValueError('top_n_method must be one of: top_diff, top_freq or a list of keywords')
@@ -111,8 +111,8 @@ class KeywordFrequencyDrift(TrainTestCheck):
 
         if context.with_display:
             dataset_names = (train_dataset.name, test_dataset.name)
-            train_to_show = mean_train_freqs[top_n_idxs]
-            test_to_show = mean_test_freqs[top_n_idxs]
+            train_to_show = max_train_freqs[top_n_idxs]
+            test_to_show = max_test_freqs[top_n_idxs]
             display = word_counts_drift_plot(train_to_show, test_to_show, top_n_words, dataset_names)
         else:
             display = None
--- a/deepchecks/nlp/checks/train_test_validation/keyword_frequency_drift.py
+++ b/deepchecks/nlp/checks/train_test_validation/keyword_frequency_drift.py
@@ -12,7 +12,8 @@
 from deepchecks import CheckResult
 from deepchecks.core import DatasetKind
 from deepchecks.core.errors import DeepchecksValueError
-from deepchecks.nlp.base_checks import TrainTestBaseCheck
+from deepchecks.utils.distribution.drift import word_counts_drift_plot
+from deepchecks.nlp.base_checks import TrainTestCheck
 from deepchecks.utils.distribution.drift import cramers_v, psi
 from typing import Union, List, Any
 from deepchecks.nlp.context import Context
@@ -23,7 +24,7 @@ import numpy as np
 __all__ = ['KeywordFrequencyDrift']
 
 
-class KeywordFrequencyDrift(TrainTestBaseCheck):
+class KeywordFrequencyDrift(TrainTestCheck):
     """
     Computes the keywords' frequencies drift between the train and the test datasets.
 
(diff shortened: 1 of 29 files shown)


TheSolY:
image

deepchecks/nlp/checks/train_test_validation/keyword_frequency_drift.py
```diff
@@ -0,0 +1,142 @@
...
+from sklearn.feature_extraction.text import TfidfVectorizer
+from nltk.stem.porter import PorterStemmer
+import numpy as np
+
+__all__ = ['KeywordFrequencyDrift']
+
+
+class KeywordFrequencyDrift(TrainTestCheck):
+    """
+    Computes the keywords' frequencies drift between the train and the test datasets.
```
noamzbr, line 30: We'd probably want to expand the description here a bit

deepchecks/nlp/checks/train_test_validation/keyword_frequency_drift.py
```diff
@@ -0,0 +1,142 @@
...
+
+    Parameters
+    ----------
+    top_n_to_show: int, default: 20
+        How many words will be displayed in the graph.
+    top_n_method: Union[str, List[str]], default: 'top_diff'
+        Decides which method will be used to select the top n words to show. Possible values are:
+        "top_diff" for the top difference between train and test, "top_freq" for the top absolute frequencies, or a list
```
noamzbr, line 38: Can format that like line 93 in `train_test_prediction_drift.py`, with bullet points per option

deepchecks/nlp/checks/train_test_validation/keyword_frequency_drift.py
```diff
@@ -0,0 +1,142 @@
...
+        of keywords in which case the words on the list are shown and the "top_n_to_show" parameter is ignored.
```
noamzbr, line 39: IMO we should have two independent arguments - one controlling top_n_method, and one enabling to add pre-selected keywords. My thinking here is that the user may want to monitor some keywords, but it still interested in watching other words that have changed dramatically

deepchecks/nlp/checks/train_test_validation/keyword_frequency_drift.py
```diff
@@ -0,0 +1,142 @@
...
+        super().__init__(**kwargs)
+        self.top_n_to_show = top_n_to_show
+        self.top_n_method = top_n_method
+
+        if drift_method == 'PSI':
+            self.drift_method = psi
+        elif drift_method == 'cramer_v':
+            self.drift_method = cramers_v
+        else:
+            raise DeepchecksValueError('drift_method must be one of: PSI, cramer_v')
```
noamzbr, line 59:
```suggestion
            raise DeepchecksValueError(f'drift_method must be one of: PSI, cramer_v, found {drift_method}')
```

deepchecks/nlp/checks/train_test_validation/keyword_frequency_drift.py
```diff
@@ -0,0 +1,142 @@
...
+        self.stem_func = PorterStemmer().stem
+        self.token_pattern = r'[a-z]{2,}'
+
+        self.top_n_words = None
```
noamzbr, line 63: Think these two are not used anywhere

deepchecks/nlp/checks/train_test_validation/keyword_frequency_drift.py
```diff
@@ -0,0 +1,142 @@
...
+from deepchecks.core import DatasetKind
+from deepchecks.utils.strings import format_number
+from deepchecks.core.errors import DeepchecksValueError
+from deepchecks.utils.distribution.drift import word_counts_drift_plot
+from deepchecks.nlp.base_checks import TrainTestCheck
+from deepchecks.utils.distribution.drift import cramers_v, psi
+from typing import Union, List, Any
+from deepchecks.nlp.context import Context
+from sklearn.feature_extraction.text import TfidfVectorizer
+from nltk.stem.porter import PorterStemmer
```
noamzbr, line 22: add nltk to `nlp-requirements.txt` (if we decide to use it eventually)

deepchecks/nlp/checks/train_test_validation/keyword_frequency_drift.py
```diff
@@ -0,0 +1,142 @@
...
+        all_data = list(train_dataset.text) + list(test_dataset.text)
+
+        vectorizer = TfidfVectorizer(input='content', strip_accents='ascii', preprocessor=self.stem_func,
+                                     token_pattern=self.token_pattern)
+        vectorizer.fit(all_data)
+        train_freqs = vectorizer.transform(train_dataset.text)
+        max_train_freqs = np.array(train_freqs.max(axis=0).todense()).reshape(-1)
+
+        test_freqs = vectorizer.transform(test_dataset.text)
+        max_test_freqs = np.array(test_freqs.max(axis=0).todense()).reshape(-1)
```
noamzbr, line 79: Why look at the max tfidf value over all samples, and not the mean one?

deepchecks/nlp/checks/train_test_validation/keyword_frequency_drift.py
```diff
@@ -0,0 +1,142 @@
...
+        top_n_diffs = np.take(word_freq_diff, top_n_idxs)
+
+        if context.with_display:
+            train_to_show = max_train_freqs[top_n_idxs]
+            test_to_show = max_test_freqs[top_n_idxs]
+            display = word_counts_drift_plot(train_to_show, test_to_show, top_n_words)
+        else:
+            display = None
+
+        result = {'drift_score': drift_score, 'top_n_diffs': dict(zip(top_n_words, top_n_diffs))}
```
noamzbr, line 108: Feel like it would be preferable to get the differences (not the np.abs of differences) here, if the user wants to use the results somehow.

deepchecks/nlp/context.py
```diff
@@ -289,8 +289,8 @@ def __init__(
             if test_dataset.has_label() and train_dataset.has_label() and not \
                     TextData.datasets_share_task_type(train_dataset, test_dataset):
                 raise DatasetValidationError('train_dataset and test_dataset must share the same label and task type')
-            if train_dataset.name == test_dataset.name:
-                raise DatasetValidationError('train_dataset and test_dataset must have different names')
+            # if train_dataset.name == test_dataset.name:
```
noamzbr, line 292: Agree, you can just delete that instead of commenting out.

deepchecks/utils/distribution/drift.py
```diff
@@ -118,7 +125,8 @@ def cramers_v(dist1: Union[np.ndarray, pd.Series], dist2: Union[np.ndarray, pd.S
 
 
 def psi(dist1: Union[np.ndarray, pd.Series], dist2: Union[np.ndarray, pd.Series],
-        min_category_size_ratio: float = 0, max_num_categories: int = None, sort_by: str = 'dist1') -> float:
+        min_category_size_ratio: float = 0, max_num_categories: int = None, sort_by: str = 'dist1',
+        from_freqs: bool = True) -> float:
```
noamzbr, line 129: Add docstring for `from_freqs` here as well

deepchecks/utils/distribution/drift.py
```diff
@@ -89,14 +91,19 @@ def cramers_v(dist1: Union[np.ndarray, pd.Series], dist2: Union[np.ndarray, pd.S
         - 'dist2': Sort by the largest dist2 categories.
         - 'difference': Sort by the largest difference between categories.
         > Note that this parameter has no effect if max_num_categories = None or there are not enough unique categories.
+    from_freqs: bool, default: False
```
noamzbr, line 94: IMO from_freq is not the right name. For me it implies that the values are already normalized 0-1 or something. Actually if I understood correctly you're saying it already comes as counts rather than raw values, so maybe `from_value_counts` or something along these lines?

tests/nlp/nlp_conftest.py
```diff
@@ -36,3 +39,28 @@ def text_multilabel_classification_dataset_mock():
     return TextData(['I think therefore I am', 'I am therefore I think', 'I am'],
                     [[0, 0, 1], [1, 1, 0], [0, 1, 0]],
                     task_type='text_classification')
+
+
+@pytest.fixture(scope='session')
+def movie_reviews_data():
```
noamzbr, line 45: How heavy is that dataset? Maybe worth considering using a subset of it

deepchecks/utils/distribution/plot.py
```diff
@@ -362,3 +362,27 @@ def sort_int_and_strings(a, b):
     )
 
     return traces, yaxis_layout
+
+def word_counts_bar_traces(
```
noamzbr, line 366: What's the difference between this and the regular categorical drift plot?

deepchecks/nlp/checks/train_test_validation/keyword_frequency_drift.py
```diff
@@ -0,0 +1,142 @@
...
+        else:
+            raise DeepchecksValueError('top_n_method must be one of: top_diff, top_freq or a list of keywords')
+
+        top_n_words = np.take(np.array(vocab), top_n_idxs)
+        top_n_diffs = np.take(word_freq_diff, top_n_idxs)
+
+        if context.with_display:
+            train_to_show = max_train_freqs[top_n_idxs]
+            test_to_show = max_test_freqs[top_n_idxs]
+            display = word_counts_drift_plot(train_to_show, test_to_show, top_n_words)
```
noamzbr, line 104: I think we should add here some text explaining what the checks does, same as we do in https://docs.deepchecks.com/stable/checks_gallery/tabular/train_test_validation/plot_train_test_feature_drift.html and other complicated checks. Explain what is a drift score (can copy the paragraph from the regular drift check), that we're showing the difference in TF-IDF and what it means.

deepchecks/nlp/checks/train_test_validation/keyword_frequency_drift.py
```diff
@@ -0,0 +1,142 @@
...
+        Decides which method will be used to select the top n words to show. Possible values are:
+        "top_diff" for the top difference between train and test, "top_freq" for the top absolute frequencies, or a list
+        of keywords in which case the words on the list are shown and the "top_n_to_show" parameter is ignored.
+    drift_method: str, default: "cramer_v"
+        Decides which method will be used for drift calculation. Possible values are:
+        "cramer_v" for Cramer's V, "PSI" for Population Stability Index (PSI).
+    """
+
+    def __init__(self,
+                 top_n_to_show: int = 20,
```
noamzbr, line 46: Think we also need n_samples and random_state here, no need to run on all the data
Unless this will become obsolete by https://github.com/deepchecks/deepchecks/issues/1942 and then please ignore my comment

deepchecks/nlp/checks/train_test_validation/keyword_frequency_drift.py
```diff
@@ -0,0 +1,142 @@
...
+        vectorizer.fit(all_data)
+        train_freqs = vectorizer.transform(train_dataset.text)
+        max_train_freqs = np.array(train_freqs.max(axis=0).todense()).reshape(-1)
+
+        test_freqs = vectorizer.transform(test_dataset.text)
+        max_test_freqs = np.array(test_freqs.max(axis=0).todense()).reshape(-1)
+        word_freq_diff = np.abs(max_train_freqs - max_test_freqs)
+
+        max_train_counts = max_train_freqs * test_dataset.n_samples
+        max_test_counts = max_test_freqs * test_dataset.n_samples
```
noamzbr, line 83: This promises a scale of 0 to 1?

noamzbr (commented):
Looks good! Main issue for now is making sure the stemmer runs fast, the rest of the comments can wait

deepchecks/utils/distribution/plot.py
TheSolY: @noamzbr the regular categorical accepts a categorical series and calculates the counts and the train-test alignment as well, here the data is already aligned and in frequencies

deepchecks/nlp/checks/train_test_validation/keyword_frequency_drift.py
TheSolY: The opposite - this makes it in total counts instead of fraction, the format that both Cramer's V and PSI expect.

deepchecks/nlp/checks/train_test_validation/keyword_frequency_drift.py
TheSolY: @noamzbr agreed, for the final version

deepchecks/nlp/checks/train_test_validation/keyword_frequency_drift.py
TheSolY: In that case, how would you suggest to combine between the keywords and the top n?

deepchecks/nlp/checks/train_test_validation/keyword_frequency_drift.py
TheSolY: I'll remove them after setting the final implementation

deepchecks/nlp/context.py
noamzbr: On a second thought, I rely on the fact that dataset have different names to retrieve the appropriate prediction indexes from the prediction and proba vectors, so I do need the names not to be the same

TheSolY: @noamzbr the name is None for both before the context assigns "Train" or "test" by default so they're the same at the point of the validation if a name is not specified.

deepchecks/utils/distribution/drift.py
TheSolY: When doing statistic tests the commonly used term is frequencies https://en.wikipedia.org/wiki/Frequency_(statistics)

deepchecks/nlp/checks/train_test_validation/keyword_frequency_drift.py
TheSolY: we need this for the condition

deepchecks/nlp/checks/train_test_validation/keyword_frequency_drift.py
noamzbr: I suggest that the keywords will be "on top" of the words that have top diff / top values in TFIDF. This suggestion has the drawback of the possibility of having to many words for the display (if the user adds 10 keywords, they'll have to reduce the n_to_show to get a normal display) . Open to other suggestions.

deepchecks/utils/distribution/plot.py
noamzbr: So maybe let's put this plot in the drift.py file? Not a must, only if you think we may use this elsewhere

deepchecks/nlp/checks/train_test_validation/keyword_frequency_drift.py
noamzbr: Oh right, and the plot is of the version before multiplying by n_samples?

TheSolY: correct

deepchecks/nlp/checks/train_test_validation/keyword_frequency_drift.py
noamzbr: condition can compute the abs internally

deepchecks/nlp/text_data.py
```diff
@@ -87,17 +87,22 @@ def __init__(
...
             if label is not None:
-                raise DeepchecksValueError('task_type must be set when label is provided')
-            self._task_type = None
+                if isinstance(label, t.Sequence):
+                    if pd.notnull(label).any():
+                        raise DeepchecksValueError('task_type must be set when label is provided')
+                else:
+                    raise DeepchecksValueError('task_type must be set when label is provided')
+
+            self._task_type = TaskType.OTHER
```
noamzbr, line 98: Did you add OTHER to the enum? Can't find it here

tests/nlp/nlp_conftest.py
```diff
@@ -36,3 +39,31 @@ def text_multilabel_classification_dataset_mock():
...
+    """Dataset of single sentence samples."""
+    sentences = [' '.join(x) for x in movie_reviews.sents()]
+    random.seed(42)
+    train_data = TextData(random.choices(sentences, k=10000))
```
noamzbr, line 49: Why use 10K here and not 1K like the two other fixtures?

deepchecks/nlp/text_data.py
TheSolY: It was already there, I didn't need to add

tests/nlp/nlp_conftest.py
TheSolY: @noamzbr because there's not enough overlap in the random split to show that it's more similar than split by label with less then 10k

noamzbr (approved):

PR #1977, opened 2022-09-04
Small scorer update

JKL98ISR:
splitted  scorer dict to multiclass/binary

### DIFF
--- a/deepchecks/tabular/checks/model_evaluation/single_dataset_performance.py
+++ b/deepchecks/tabular/checks/model_evaluation/single_dataset_performance.py
@@ -54,7 +54,7 @@ class SingleDatasetPerformance(SingleDatasetCheck, ReduceMetricClassMixin):
         scorers = context.get_scorers(self.scorers, use_avg_defaults=False)
 
         results = []
-        classes = dataset.classes
+        classes = model.classes_
         label = cast(pd.Series, dataset.label_col)
         n_samples = label.groupby(label).count()
         for scorer in scorers:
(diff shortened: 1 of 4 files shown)


deepchecks/tabular/context.py
```diff
@@ -103,6 +105,8 @@ def __init__(self,
 
         self.predictions = pd.concat(predictions, axis=0) if predictions else None
         self.probas = pd.concat(probas, axis=0) if probas else None
+        self.classes_ = None if self.probas is None else sorted(set(self.predictions))
```
Nadav-Barak, line 108: Add assert if self.classes_ not the same size as self.probas

deepchecks/tabular/context.py
```diff
@@ -58,6 +59,7 @@ class _DummyModel:
     feature_df_list: t.List[pd.DataFrame]
     predictions: pd.DataFrame
     proba: pd.DataFrame
+    classes_: t.List[t.Hashable]
```
Nadav-Barak, line 62: Check for places where we create dummy models and pass the classes argument when available

deepchecks/tabular/checks/model_evaluation/single_dataset_performance.py
```diff
@@ -63,7 +63,7 @@ def run_logic(self, context: Context, dataset_kind) -> CheckResult:
                 results.append([pd.NA, scorer.name, scorer_value, len(label)])
             else:
                 results.extend(
-                    [[class_name, scorer.name, class_score, n_samples[class_name]]
+                    [[class_name, scorer.name, class_score, n_samples.get(class_name, pd.NA)]
```
Nadav-Barak, line 66: If there are no samples for this label then num samples should be 0

deepchecks/tabular/checks/model_evaluation/single_dataset_performance.py
```diff
@@ -54,7 +54,7 @@ def run_logic(self, context: Context, dataset_kind) -> CheckResult:
         scorers = context.get_scorers(self.scorers, use_avg_defaults=False)
 
         results = []
-        classes = dataset.classes
+        classes = model.classes_ if hasattr(model, 'classes_') else dataset.classes
```
Nadav-Barak, line 57: Is this a hot fix? if not then it should be done in the context level not inside a specific check

Nadav-Barak (changes requested):

Nadav-Barak (approved):

PR #1983, opened 2022-09-06
[Issue-1712] results selection by kind

### LINKED ISSUE #1712: [FEAT] Have checks in suite results addressable by name and not by their index
If suite is updated (e.g. in new version) - sometimes indices change, requiring manual change of logic.
Also - less convenient, if want to investigate more results of specific checks, have to manually iterate and search for relevant check.

Have checks addressable by name and not by index.
Have to consider convenient solution given that checks aren't necessarily unique in suite. 

**Additional context**
Request given as feedback from community by @DL1992

yromanyshyn:
resolves #1712

### DIFF
--- a/deepchecks/core/suite.py
+++ b/deepchecks/core/suite.py
@@ -115,6 +115,47 @@ class SuiteResult(DisplayableResult):
                 output.append(result)
         return output
 
+    def select_results_by_check_kind(
+        self,
+        kind: Union[Type[BaseCheck], Sequence[Type[BaseCheck]]]
+    ) -> List['check_types.BaseCheckResult']:
+        """Select results by check type(s).
+
+        Parameters
+        ==========
+        kind : Union[Type[BaseCheck], Sequence[Type[BaseCheck]]]
+            a check type(s), results of which to return
+
+        Returns
+        =======
+        List['check_types.BaseCheckResult']:
+            list of results and failures
+
+        Examples
+        ========
+        >> from deepchecks.tabular.suites import data_integrity
+        >> from deepchecks.tabular.checks import IsSingleValue
+        >>
+        >> suite_result = data_integrity(...)
+        >> check_results = suite_result.select_results_by_check_kind(IsSingleValue)
+        >>
+        """
+        if isinstance(kind, Sequence):
+            if len(kind) == 0:
+                raise ValueError("Empty sequence is not allowed")
+            kind = tuple(kind)
+        elif isinstance(kind, type) and issubclass(kind, BaseCheck):
+            kind = (kind,)
+        else:
+            t = type(kind).__name__
+            raise TypeError(f'Unsupported type of "kind" parameter - {t}')
+
+        return [
+            r
+            for r in self.results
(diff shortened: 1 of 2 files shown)


deepchecks/core/suite.py
```diff
@@ -115,6 +115,47 @@ def select_results(self, idx: Set[int]) -> List[Union[
                 output.append(result)
         return output
 
+    def select_results_by_check_kind(
```
Nadav-Barak, line 118: kind is not vary clear, Would rename to type.
Consider also allowing passing the check kind as str

yromanyshyn: > kind is not vary clear, Would rename to type.

renamed

> Consider also allowing passing the check kind as str

it is always possible to make a mistake in a string, so I do not see how it is better/

tests/base/check_suite_test.py
```diff
@@ -223,3 +223,41 @@ def test_config():
     conf_suite_mod = BaseSuite.from_config(suite_mod)
     assert_that(conf_suite_mod.name, equal_to('Model Evaluation Suite'))
     assert_that(conf_suite_mod.checks.values(), has_length(check_amount))
+
+
+def test_results_selection_by_check_type():
```
Nadav-Barak, line 228: Would also add a test in which you perform suite.run then accessing one of the results
Can also verify what happen if you try to get a check result for a check that do not exist in the suite / check run that failed

deepchecks/core/suite.py
```diff
@@ -115,6 +115,47 @@ def select_results(self, idx: Set[int]) -> List[Union[
...
+            if len(check_type) == 0:
+                raise ValueError('Empty sequence is not allowed')
+            check_type = tuple(check_type)
+        elif isinstance(check_type, type) and issubclass(check_type, BaseCheck):
+            check_type = (check_type,)
+        else:
+            t = type(check_type).__name__
+            raise TypeError(f'Unsupported type of "check_type" parameter - {t}')
+
+        return [
```
Nadav-Barak, line 153: Would convert into regular for and add warning / exception for when:
1. the check requested is not part of the suite
2. the check requested run failed for some reason (not sure if a check result is created in this case)

Nadav-Barak (approved):
Optional and can be in a separate issue: Add a similar method to access the check itself not its result (for purposes of updating its conditions / removing it from suite)

deepchecks/core/suite.py
yromanyshyn: 1. added warning
2. `CheckFailure` instances are created for failed checks and a resulting list will include thems, IMO, it is bad idea to raise an exception for them/

tests/base/check_suite_test.py
yromanyshyn: sorry for delays

> Would also add a test in which you perform suite.run then accessing one of the results

added

> Can also verify what happen if you try to get a check result for a check that do not exist in the suite 

you will get an empty list

> check run that failed

`CheckFailure` is also considered as a result, so a resulting list will include instances of it

PR #1987, opened 2022-09-06
top_5 aggregation method + bug fix in percent_of_nulls.py

Nadav-Barak:
Bonus PR, no issue


noamzbr (approved):
1. Add test for percent null with FI
2. Consider having the reduce docstring inside a variable, similar to additional_context_params

PR #1994, opened 2022-09-08
Added duplicate column validation

### LINKED ISSUE #1909: [FEAT] Add validation in tabular dataset creation that there are no columns with identical names
A pandas dataframe can have columns with the same name, but for the operation of the deepchecks package that can lead to unexpected bugs.

Validate that this is not the case in the dataset init

kishore-s-gitaa:
Closes #1909.


deepchecks/tabular/dataset.py
```diff
@@ -126,6 +127,23 @@ def __init__(
...
+        dataset_columns = self._data.columns
+        columns_frequency_counter = Counter(dataset_columns)
+
+        # Checking for duplicate columns
+        duplicated_columns = []
+        for column, frequency in columns_frequency_counter.items():
+            if frequency > 1:
+                duplicated_columns.append(column)
+                
+        if len(duplicated_columns) > 1:
```
noamzbr, line 140: Otherwise it'll fail when there is only one duplicate (:
```suggestion
        if len(duplicated_columns) >= 1:
```

deepchecks/tabular/dataset.py
```diff
@@ -126,6 +127,23 @@ def __init__(
...
+            raise DeepchecksValueError(
+                f"Data has {len(duplicated_columns)} duplicate columns. "
+                "Change the duplicate column names or remove them from the data. "
+                f"Duplicated columns: {duplicated_columns}"
```
noamzbr, line 144:
```suggestion
                f"Duplicated column names: {duplicated_columns}"
```

noamzbr (changes requested):
Great work @kishore-s-gitaa! Thanks a lot for your contribution 🚀 

I've requested changes because there was a small bug in the case when there is only a single duplicate column. Also, as we've seen there could be subtle bugs here - I'd appreciate if you could add a simple test to verify this functionality to `dataset_test.py`.

kishore-s-15:
> Great work @kishore-s-gitaa! Thanks a lot for your contribution rocket
> 
> I've requested changes because there was a small bug in the case when there is only a single duplicate column. Also, as we've seen there could be subtle bugs here - I'd appreciate if you could add a simple test to verify this functionality to `dataset_test.py`.

Sure @noamzbr, Thanks for the code review. I'll work on the requested changes. I'll add a test for validating this feature (even I was thinking of asking you the same) 😀

deepchecks/tabular/dataset.py
kishore-s-15: Ok, I'll make the change.

deepchecks/tabular/dataset.py
kishore-s-15: Sure, I'll change the terminology used.

kishore-s-15:
@noamzbr I've made the requested changes and added a test case for the duplicate column name validation.

noamzbr (approved):
Great, approved once the test issues and pylint issues are resolved.

tests/base/dataset_test.py
```diff
@@ -1049,3 +1049,33 @@ def test_cat_features_warning(iris, caplog):
...
+    """
+
+    # Duplicate column names
+    columns = ["sepal length (cm)", "sepal width (cm)", "sepal length (cm)",
+               "sepal width (cm)", "target"]
+    
+    duplicated_columns = ["sepal length (cm)", "sepal width (cm)"]
+
+    # data now has duplicate column names
+    iris.columns = columns
```
noamzbr, line 1065: The tests are failing because you're modifying the iris fixture in place. Please use a copy of the iris dataset `iris_copy = iris.copy()` and modify it's columns instead.

kishore-s-15: > Great, approved once the test issues and pylint issues are resolved.

Sure, I'll work on these changes.

noamzbr:
@kishore-s-gitaa Thanks a lot for your contribution!

kishore-s-15:
Thanks @noamzbr 👍🏻

PR #1995, opened 2022-09-08
Deprecate old modules

### LINKED ISSUE #1479: Deprecate old check modules
In 0.7.0 we refactored the checks into new modules to match the suites. In 0.8.0 we're ought to deprecate the support for importing from the old modules.

noamzbr:
#### Reference Issues/PRs

Closes #1479 

All deprecated but the ability to import tabular from deepchecks directly.


deepchecks/checks.py
```diff
@@ -34,7 +34,6 @@
     'SpecialCharacters',
     'StringLengthOutOfBounds',
```
Nadav-Barak, line 35: Isnt all of this file to be deprecated?

nirhutnik (approved):
I think some of the drift tests had a test for the deprecation warnings, did you search for the relevant strings in all of the project?

deepchecks/checks.py
noamzbr: yes but not now. We decided to keep the ability to import tabular stuff directly from deepcheck until 0.10.0

ItayGabbay (approved):

PR #1998, opened 2022-09-10
Changed owners:

nirhutnik:
* Added Noam to approve docs
* Changed that devs are only responsible for tests/base instead of tests/ folder. This way, most changes in checks will not require devs (and can be either algo or devs).

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -6,7 +6,7 @@
 
 *       @deepchecks/deepchecks-team
 
-/docs/ @ItayGabbay @shir22
+/docs/ @ItayGabbay @shir22 @noambzr
 /docs/source/examples/ @deepchecks/deepchecks-team
 
 .github @ItayGabbay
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -6,7 +6,7 @@
 
 *       @deepchecks/deepchecks-team
 
-/docs/ @ItayGabbay @shir22
+/docs/ @ItayGabbay @shir22 @noambzr
```
ItayGabbay, line 9: It doesn't work. I think it's noamzbr?

ItayGabbay (changes requested):

.github/CODEOWNERS
nirhutnik: You're right, thanks!

.github/CODEOWNERS
```diff
@@ -15,7 +15,7 @@ README.rst @shir22
 
 requirements/ @ItayGabbay
 deepchecks/core/ @deepchecks/dev-team
-tests/ @deepchecks/dev-team
+tests/base @deepchecks/dev-team
```
noamzbr, line 18: Add also tests/serialization

nirhutnik: Fixed

noamzbr (approved):

ItayGabbay (approved):

PR #2004, opened 2022-09-12
1939 semantic segmentation metrics

TheSolY:
MeanDice runtime improvement
MeanIoU implemented for segmentation
test for MeanIoU and MeanDice


deepchecks/vision/metrics_utils/semantic_segmentation_metrics.py
```diff
@@ -37,26 +48,78 @@ def update(self, output: Tuple[torch.Tensor, torch.Tensor]):
...
+
+            for class_id in [int(x) for x in torch.unique(y[i])]:
+                self._evals[class_id]['iou'] += iou_per_class[class_id]
+                self._evals[class_id]['count'] += 1
+
+    def compute(self):
+        """Compute metric value."""
+        sorted_classes = [int(class_id) for class_id in sorted(self._evals.keys())]
+        max_class = max(sorted_classes)
+        scores_per_class = np.empty(max_class + 1) * np.nan
```
JKL98ISR, line 119: why not -np.ones like in detections? and then if not class is found put -1 as score?

noamzbr (commented):
Looks good! We should add MeanIoU to scorers.py `semantic_segmentation_dict` so that the user will be about to use it. 

I think we should also add to both these metrics an averaging_method argument (similarly toObjectDetectionTpFpFn) so that the user will be able to request also micro/macro aggregations. In classification this is handled by falling back on sklearn metrics, but here we don't have this option. If this is a lot of work let's open a separate PR, but it may be easy given that the current implementation already gives us per-class.

deepchecks/vision/metrics_utils/semantic_segmentation_metrics.py
nirhutnik: Because then you got -1 appearing in the graphs... the metric_results_to_df function ignores np.nan and that's the wanted behavior

deepchecks/vision/metrics_utils/scorers.py
```diff
@@ -93,7 +93,12 @@ def get_default_semantic_segmentation_scorers() -> t.Dict[str, Metric]:
 }
 
 semantic_segmentation_dict = {
-    'dice': MeanDice
+    'dice': MeanDice,
```
noamzbr, line 96: Think this should be 'dice_per_class' to comply with out naming convention.

noamzbr (approved):

PR #2006, opened 2022-09-12
use deepchecks metrics for precision recall f1

noamzbr:
#### Reference Issues/PRs

Closes https://github.com/deepchecks/private-issues/issues/73 (For now, we'll probably want to keep improving and extend improvements to other metrics)

Uses the metrics implemented in https://github.com/deepchecks/deepchecks-metrics to reduce the runtime of precision/recall/f1 by around %60

Does not cover regression metrics and roc_auc for example.

### DIFF
--- a/deepchecks/tabular/checks/model_evaluation/single_dataset_performance.py
+++ b/deepchecks/tabular/checks/model_evaluation/single_dataset_performance.py
@@ -56,7 +56,10 @@ class SingleDatasetPerformance(SingleDatasetCheck, ReduceMetricClassMixin):
         results = []
         classes = dataset.classes
         label = cast(pd.Series, dataset.label_col)
-        n_samples = label.groupby(label).count()
+        if context.with_display:
+            n_samples = label.groupby(label).count()
+        else:
+            n_samples = dict(zip(classes, [None] * len(classes)))
         for scorer in scorers:
             scorer_value = scorer(model, dataset)
             if isinstance(scorer_value, Number):
--- a/deepchecks/tabular/metric_utils/scorers.py
+++ b/deepchecks/tabular/metric_utils/scorers.py
@@ -16,11 +16,11 @@ from numbers import Number
 import numpy as np
 import pandas as pd
 from sklearn.base import BaseEstimator, ClassifierMixin
-from sklearn.metrics import (get_scorer, make_scorer, mean_absolute_error, mean_squared_error)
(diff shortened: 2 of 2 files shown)


ItayGabbay (approved):

deepchecks/tabular/checks/model_evaluation/single_dataset_performance.py
```diff
@@ -56,7 +56,10 @@ def run_logic(self, context: Context, dataset_kind) -> CheckResult:
         results = []
         classes = dataset.classes
         label = cast(pd.Series, dataset.label_col)
-        n_samples = label.groupby(label).count()
+        if context.with_display:
+            n_samples = label.groupby(label).count()
+        else:
+            n_samples = dict(zip(classes, [None] * len(classes)))
```
Nadav-Barak, line 62: weird that n_samples is a dict, consider changing the name

deepchecks/tabular/checks/model_evaluation/single_dataset_performance.py
```diff
@@ -56,7 +56,10 @@ def run_logic(self, context: Context, dataset_kind) -> CheckResult:
         results = []
         classes = dataset.classes
         label = cast(pd.Series, dataset.label_col)
-        n_samples = label.groupby(label).count()
+        if context.with_display:
+            n_samples = label.groupby(label).count()
```
Nadav-Barak, line 60: can use Counter

deepchecks/tabular/metric_utils/scorers.py
```diff
@@ -105,9 +109,16 @@
 multiclass_scorers_dict = {
     'accuracy': get_scorer('accuracy'),
```
Nadav-Barak, line 110: Loved how you kept the order

Nadav-Barak:
Also, https://github.com/deepchecks/deepchecks-metrics does not exist

deepchecks/tabular/checks/model_evaluation/single_dataset_performance.py
noamzbr: It's a series in the with_display==True case, but a dict has the same api

deepchecks/tabular/checks/model_evaluation/single_dataset_performance.py
noamzbr: if it's the with_display case then efficiency isn't that important

PR #2021, opened 2022-09-19
[issue-2017] ImagePropertyDrift serialization fix

### LINKED ISSUE #2017: [BUG] ImagePropertyDrift is not serializable
It is not serializable therefore cannot be used in monitoring.
Big problem since vision use cases include this check

yromanyshyn:
resolves #2017


JKL98ISR:
pushed #2059 instead

PR #2057, opened 2022-09-28
Default vision properties improvements 

TheSolY:
default label and prediction properties accept a list in addition to the previous tensor only to comply with calc_vision_properties.
Implemented calc_default_image_properties to be used instead of calc_vision_properties on the default image properties to speed up this use case (x2 speedup).

### DIFF
--- a/deepchecks/vision/utils/image_properties.py
+++ b/deepchecks/vision/utils/image_properties.py
@@ -9,7 +9,7 @@
 # ----------------------------------------------------------------------------
 #
 """Module containing the image formatter class for the vision module."""
-from typing import List, Tuple
+from typing import List, Tuple, Dict
 
 import numpy as np
 from skimage.color import rgb2gray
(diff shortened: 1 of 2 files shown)


deepchecks/vision/utils/image_properties.py
```diff
@@ -122,6 +132,25 @@ def get_dimension(img) -> int:
...
 
+def calc_default_image_properties(batch: List[np.ndarray]) -> Dict:
+    """Speed up the calculation for the default image properties by sharing common actions."""
+    results_dict = {}
+    sizes_array = _sizes_array(batch)
+    results_dict['Aspect Ratio'] = list(sizes_array[:, 0] / sizes_array[:, 1])
+    results_dict['Area'] = list(sizes_array[:, 0] * sizes_array[:, 1])
+
+    grayscale_images = [rgb2gray(img) if _is_grayscale(img) else img for img in batch]
+    results_dict['Brightness'] = [image.mean() for image in grayscale_images]
```
noamzbr, line 143: Why not use the existing code for brightness etc? We're still iterating on the images here, so we're not saving any time here except with the _rgb_relative_intensity_mean_array

TheSolY: We're saving time by calling _is_grayscale once for both brightness and RMS

PR #2085, opened 2022-10-01
[Doc] fix tabular class imbalance

### LINKED ISSUE #1607: [FEAT]Class Imbalance check
Add a data integrity (single dataset) check for class imbalance, relevant for both tabular and vision
The check calculate the ratio between the class frequencies, for a multiclass case, the ratio between the most prevalent and the least prevalent.
This check should have a condition to alert if the ratio is above a certain threshold.

DanBasson:
#### Reference Issues/PRs

Fixes #1607

minor fix in doc for tabular class imbalance - class imbalance description and "Generate data" link.

### DIFF
--- a/docs/source/checks/tabular/data_integrity/plot_class_imbalance.py
+++ b/docs/source/checks/tabular/data_integrity/plot_class_imbalance.py
@@ -3,32 +3,32 @@
 .. _plot_tabular_class_imbalance:
 
 Class Imbalance
-***************
+****************
 
 This notebook provides an overview for using and understanding the Class Imbalance check.
 
 **Structure:**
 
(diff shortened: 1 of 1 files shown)


docs/source/checks/tabular/data_integrity/plot_class_imbalance.py
```diff
@@ -3,32 +3,32 @@
 .. _plot_tabular_class_imbalance:
 
 Class Imbalance
-***************
+****************
```
noamzbr, line 6:
```suggestion
***************
```
Why add another `*`? It should match the length of the title, and it does.

noamzbr (approved):

PR #2091, opened 2022-10-09
2027 single dataset check select dataset

TheSolY:
SingleDatasetCheck inside the suite can now run on only train, only test, or both with the parameter run_single_dataset.


deepchecks/vision/suite.py
```diff
@@ -158,21 +164,29 @@ def _update_loop(
...
         # SingleDatasetChecks have different handling, need to initialize them here (to have them ready for different
         # dataset kind)
+
         for idx, check in single_dataset_checks.items():
-            try:
-                check.initialize_run(context, dataset_kind=dataset_kind)
-            except Exception as exp:
-                results[idx] = CheckFailure(check, exp, type_suffix)
+            if flag_skip_single:
+                results[idx] = -1
```
Nadav-Barak, line 184: Is this necessary? if not possible to do it in another way then  would add a comment in the code why this is the case. (we will not remember)

Nadav-Barak (approved):

PR #2101, opened 2022-10-23
1743 feat pass scorer list to model evaluation suite

TheSolY:
API unified for all the relevant checks in both tabular and vision.
Passed over the documentation and made sure that no adjustments were required.


noamzbr (commented):
Looks good. One thing I think that we should add - an example of using that in the model evaluation quickstart, mention the fact that params can be modified for all checks in Create a Custom Suite guide and last mention that scorers can be determined that way in the metrics guide

noamzbr (approved):

PR #2111, opened 2022-10-27
1852 single dataset static predictions

TheSolY:
Pass pre-calculated predictions for single dataset checks to predictions for vision and to y_pred and y_proba for tabular instead of train_predictions etc.

### DIFF
--- a/deepchecks/vision/base_checks.py
+++ b/deepchecks/vision/base_checks.py
@@ -9,6 +9,7 @@
 # ----------------------------------------------------------------------------
 #
 """Module for vision base checks."""
+import warnings
 from typing import Any, Dict, Mapping, Optional, Sequence, Union
 
 import torch
@@ -17,6 +18,7 @@ from torch import nn
 
 from deepchecks.core.check_result import CheckResult
 from deepchecks.core.checks import DatasetKind, ModelOnlyBaseCheck, SingleDatasetBaseCheck, TrainTestBaseCheck
+from deepchecks.core.errors import DeepchecksValueError
 from deepchecks.utils.ipython import ProgressBarGroup
 from deepchecks.vision import deprecation_warnings  # pylint: disable=unused-import # noqa: F401
 from deepchecks.vision._shared_docs import docstrings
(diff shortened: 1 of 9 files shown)


deepchecks/vision/base_checks.py
```diff
@@ -49,10 +51,12 @@ def run(
         random_state: int = 42,
         n_samples: Optional[int] = 10_000,
         with_display: bool = True,
+        predictions: Optional[Dict[int, Union[Sequence[torch.Tensor], torch.Tensor]]] = None,
```
JKL98ISR, line 54: add properties too

docs/source/user-guide/tabular/supported_models.rst
```diff
@@ -121,6 +121,15 @@ ROC Curve and the AUC metric) will not run.
 Running Deepchecks with Pre-computed Predictions
 ------------------------------------------------
 
+It is possible to pass pre-computed predictions to the checks, this can be useful for example when the model is not
```
Nadav-Barak, line 124: Not sure this additional text is required, this info is IMO covered in the text above

docs/source/user-guide/tabular/supported_models.rst
```diff
@@ -121,6 +121,15 @@ ROC Curve and the AUC metric) will not run.
...
+available, only the results previously saved from it.
+The pre-computed predictions should be passed to suite/check's ``run`` method in the appropriate format.
+The parameters to pass are ``y_pred`` and ``y_proba`` for single dataset checks or ``y_pred_train`` and
+``y_proba_train`` and ``y_pred_test`` and ``y_proba_test`` for checks that use both datasets.
+
+Code Example
```
Nadav-Barak, line 130: the additional title is redundant in my opinion

JKL98ISR (approved):

docs/source/user-guide/tabular/supported_models.rst
noamzbr: Tend to agree. What is the added value of this paragraph?

noamzbr (commented):
Looks good, agree with @Nadav-Barak though that the extra text in Running Deepchecks with Pre-computed Predictions seems unnecessary

noamzbr (approved):
Minor ammendment

Issue #1772, opened 2022-07-14
[FEAT][Tabular] Unify Regression Error Distribution and Regression Systematic Error


Nadav-Barak:
**Is your feature request related to a problem? Please describe.**
Both checks convey the same information

**Describe the solution you'd like**
Deprecate the Regression Systematic Error which is less clear 
Add relevant information (mean error + std) to the Regression Error Distribution display + result value.
Also update the example plot and add a relevant condition - max_bias_condition.

PR #2112, opened 2022-10-27
regression_error_distribution makeover

Nadav-Barak:
Resolves https://github.com/deepchecks/deepchecks/issues/1772

### DIFF
--- a/deepchecks/tabular/checks/model_evaluation/regression_error_distribution.py
+++ b/deepchecks/tabular/checks/model_evaluation/regression_error_distribution.py
@@ -9,9 +9,13 @@
 # ----------------------------------------------------------------------------
 #
 """The regression_error_distribution check module."""
+from typing import Dict
+
+import numpy as np
 import pandas as pd
 import plotly.express as px
 from scipy.stats import kurtosis
+from sklearn.metrics import mean_squared_error
 
 from deepchecks.core import CheckResult, ConditionCategory, ConditionResult
 from deepchecks.tabular import Context, SingleDatasetCheck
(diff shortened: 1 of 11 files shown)


deepchecks/tabular/checks/model_evaluation/regression_error_distribution.py
```diff
@@ -21,12 +25,13 @@
 
 
 class RegressionErrorDistribution(SingleDatasetCheck):
-    """Check regression error distribution.
+    """Check different aspects and properties of the regression error distribution.
```
noamzbr, line 28:
```suggestion
    """Check for systematic error and abnormal shape in the regression error distribution.
```

docs/source/checks/tabular/model_evaluation/plot_regression_error_distribution.py
```diff
@@ -9,75 +9,64 @@
...
 
 
 What is the Regression Error Distribution check?
 ==================================================
 The ``RegressionErrorDistribution`` check shows the distribution of the regression error,
-and enables to set conditions on the distribution's kurtosis. Kurtosis is a measure of the shape
-of the distribution, helping us understand if the distribution is significantly "wider" from
-the normal distribution, which may imply a certain cause of error deforming the normal shape.
-
+and enables to set conditions on several of the distribution parameters including systematic error and Kurtosis values.
```
noamzbr, line 19: It's not including these, it's only these

deepchecks/tabular/checks/model_evaluation/regression_systematic_error.py
```diff
@@ -37,6 +39,8 @@ def __init__(
         random_state: int = 42,
         **kwargs
     ):
+        warnings.warn('RegressionSystematicError check is deprecated and will be removed in future version,'
```
noamzbr, line 42: I know we didn't use to do that, but I think it's worth while adding a notice about deprecation (and a link to the new updated better check) in the doc for this check

noamzbr (approved):

PR #2115, opened 2022-10-28
Update columns_info.py

alphabetagamer:
#### Reference Issues/PRs

top n columns was not being passed to column importance sorter dict.  column_importance_sorter_dict automatically picks 10 as default value inside.

#### Any other comments?

---

### DIFF
--- a/deepchecks/tabular/checks/data_integrity/columns_info.py
+++ b/deepchecks/tabular/checks/data_integrity/columns_info.py
@@ -42,7 +42,7 @@ class ColumnsInfo(SingleDatasetCheck):
         """
         dataset = context.get_data_by_kind(dataset_kind)
         columns_info = dataset.columns_info
-        columns_info = column_importance_sorter_dict(columns_info, dataset, context.feature_importance)
+        columns_info = column_importance_sorter_dict(columns_info, dataset, context.feature_importance, self.n_top_columns)
 
         columns_info_to_display = (
             columns_info
--- a/deepchecks/tabular/checks/data_integrity/columns_info.py
+++ b/deepchecks/tabular/checks/data_integrity/columns_info.py
@@ -43,14 +43,7 @@ class ColumnsInfo(SingleDatasetCheck):
         dataset = context.get_data_by_kind(dataset_kind)
         columns_info = dataset.columns_info
         columns_info = column_importance_sorter_dict(columns_info, dataset, context.feature_importance, self.n_top_columns)
-
-        columns_info_to_display = (
-            columns_info
-            if len(columns_info) <= self.n_top_columns
-            else dict(list(columns_info.items())[:self.n_top_columns])
-        )
-
-        df = pd.DataFrame.from_dict(columns_info_to_display, orient='index', columns=['role'])
+        df = pd.DataFrame.from_dict(columns_info, orient='index', columns=['role'])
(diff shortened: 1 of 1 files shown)


deepchecks/tabular/checks/data_integrity/columns_info.py
```diff
@@ -42,7 +42,7 @@ def run_logic(self, context: Context, dataset_kind) -> CheckResult:
         """
         dataset = context.get_data_by_kind(dataset_kind)
         columns_info = dataset.columns_info
-        columns_info = column_importance_sorter_dict(columns_info, dataset, context.feature_importance)
+        columns_info = column_importance_sorter_dict(columns_info, dataset, context.feature_importance, self.n_top_columns)
```
noamzbr, line 45: In that case, we can delete lines 47-51, as they are there only to cap the number of displayed columns at self.n_top_columns

noamzbr (commented):
Good catch, and thanks for the contribution @alphabetagamer! Just to make things more organized, can you please also delete the lines coming after this one that are now redundant? Thanks!

deepchecks/tabular/checks/data_integrity/columns_info.py
alphabetagamer: @noamzbr could you verify the changes?

noamzbr (approved):

deepchecks/tabular/checks/data_integrity/columns_info.py
noamzbr: looks good!

noamzbr:
Hi @alphabetagamer. Sorry for not realizing it earlier, but there was a reason for how the code was built.
The idea is that we always want to return (in the check value) information about all columns, and that the `n_top_columns` is designed only to limit the length of the display. That's why `column_importance_sorter_dict` didn't limit the returned columns, and `n_top_columns` is only used to cut `columns_info_to_display`, which are used solely for the visual display.

That being said, I'll be very happy to receive further contributions from you! You can have a look at our list of [good first issues](https://github.com/deepchecks/deepchecks/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22) for ideas.

alphabetagamer:
@noamzbr The function column_importance_sorter_dict has n_top picking 10 as default parameter value hence it is clipping the columns everytime :) that is why we need to pass the n_top from the caller function to the column_importance_sorter_dict function call else it will always return only top 10 columns. My first pull request was with that change where we passed the value from the caller as parameter.

alphabetagamer:
@noamzbr Yeah passing none would work best and we can make the code leaner by simply writing lines 47-51 from 
```
columns_info_to_display = (
            columns_info
            if len(columns_info) <= self.n_top_columns
            else dict(list(columns_info.items())[:self.n_top_columns])
```
to 
```
columns_info_to_display = dict(list(columns_info.items())[:self.n_top_columns])
```
since python automatically caps list items when slicing beyond the list's length.
Let me know if this is alright so I'll raise a new merge request. Also thank you for being understanding and patient since this is my first time contributing to an open-source project.

noamzbr:
Hi @alphabetagamer sounds good I'd be glad merge your new merge request.

Thank you for finding the bug, and insisting on solving it despite the initial misunderstanding (:

PR #2121, opened 2022-11-01
[Fix: 2110] Pass feature_importance_timeout through to feature importance

### LINKED ISSUE #2110: [BUG] feature_importance_timeout is hard coded and therefore does not work
**Describe the bug**
Feature Importance can be time limited or even skipped, as per documentation:
https://docs.deepchecks.com/stable/user-guide/tabular/feature_importance.html#what-if-the-feature-importance-calculation-takes-too-long

> Permutation feature importance is a complex calculation which can take a lot of time, depending on the number of features and samples in your data. However, except for certain checks, deepchecks does not require feature importance. Therefore, if you want deepchecks to skip the calculation of feature importance, you can use the feature_importance_timeout parameter in the run function, available in all checks and suites. Before running the permutation feature importance, deepchecks predicts the calculation time. If the predicted time is bigger than feature_importance_timeout, the process will be skipped. Configuring this parameter to 0 will ensure the calculation is always skipped.

```
check = MultivariateDrift()
check.run(ds_train, ds_test, model, feature_importance_timeout=0)
```

The variable feature_importance_timeout is correctly passed to the context, however it is not passed in the MultivariateDrift run_logic:

https://github.com/deepchecks/deepchecks/blob/5d5c35cd56afae1094f066729a9c5ac73197913d/deepchecks/tabular/checks/train_test_validation/multivariate_drift.py#L113

```
values_dict, displays = run_multivariable_drift(
            train_dataframe=train_dataset.features_columns,
            test_dataframe=test_dataset.features_columns,
            numerical_features=numerical_features,
            cat_features=cat_features,
            sample_size=sample_size, random_state=self.random_state,
            test_size=self.test_size, n_top_columns=self.n_top_columns,
            min_feature_importance=self.min_feature_importance,
            max_num_categories_for_display=self.max_num_categories_for_display,
            show_categories_by=self.show_categories_by,
            min_meaningful_drift_score=self.min_meaningful_drift_score,
            with_display=context.with_display,
            dataset_names=(train_dataset.name, test_dataset.name)
        )
```

Note that the variable is not passed.
Following the function:

https://github.com/deepchecks/deepchecks/blob/5d5c35cd56afae1094f066729a9c5ac73197913d/deepchecks/core/check_utils/multivariate_drift_utils.py#L73

``` fi, importance_type = calculate_feature_importance_or_none(
        domain_classifier,
        domain_test_dataset,
        force_permutation=True,
        permutation_kwargs={'n_repeats': 10,
                            'random_state': random_state,
                            'timeout': 120,
                            'skip_messages': True}
    )
```

It's hard coded.
Hence, it can never be disabled, nor can the threshold be increased over 120 seconds (e.g. for larger datasets)

**To Reproduce**
Try the example in the documentation. It will not work.

**Expected behavior**
The variable should be passed. It also seems that https://github.com/deepchecks/deepchecks/blob/5d5c35cd56afae1094f066729a9c5ac73197913d/deepchecks/tabular/utils/feature_importance.py#L205
actually does accept the value 0 as a skip command.

**Screenshots**
-

doesn't matter

**Additional context**

I might be able to do a pull request as the fix seems very simple.

IngoStatworx:
Quick fix to pass timeout variable for feature importance.
Raises error if set to 0 (since this is not supported by the algorithm).

Added tests for this case, and for None timeout. 

#### Reference Issues/PRs

Fixes #2110. 

The "feature_importance_timeout" parameter was hard coded

#### Any other comments?

I was unable to install the dev requirements and run a test locally!!!
This is just to get the ball rolling on this issue.

### DIFF
--- a/deepchecks/core/check_utils/multivariate_drift_utils.py
+++ b/deepchecks/core/check_utils/multivariate_drift_utils.py
@@ -42,7 +42,8 @@ def run_multivariable_drift(train_dataframe: pd.DataFrame, test_dataframe: pd.Da
                             max_num_categories_for_display: int, show_categories_by: str,
                             min_meaningful_drift_score: float,
                             with_display: bool,
-                            dataset_names: Tuple[str] = DEFAULT_DATASET_NAMES
+                            dataset_names: Tuple[str] = DEFAULT_DATASET_NAMES,
+                            feature_importance_timeout: int = 120,
                             ):
     """Calculate multivariable drift."""
     train_sample_df = train_dataframe.sample(sample_size, random_state=random_state)[numerical_features + cat_features]
@@ -80,7 +81,7 @@ def run_multivariable_drift(train_dataframe: pd.DataFrame, test_dataframe: pd.Da
         force_permutation=True,
         permutation_kwargs={'n_repeats': 10,
                             'random_state': random_state,
-                            'timeout': 120,
+                            'timeout': feature_importance_timeout,
                             'skip_messages': True}
     )
 
--- a/deepchecks/core/check_utils/multivariate_drift_utils.py
+++ b/deepchecks/core/check_utils/multivariate_drift_utils.py
@@ -16,10 +16,8 @@ import pandas as pd
 import plotly.graph_objects as go
 
 with warnings.catch_warnings():
-    warnings.simplefilter("ignore")
-    from sklearn.experimental import (
-        enable_hist_gradient_boosting,
-    )  # noqa # pylint: disable=unused-import
+    warnings.simplefilter('ignore')
+    from sklearn.experimental import enable_hist_gradient_boosting  # noqa # pylint: disable=unused-import
 
 from sklearn.ensemble import HistGradientBoostingClassifier
 from sklearn.metrics import roc_auc_score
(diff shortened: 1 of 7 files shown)


deepchecks/core/check_utils/multivariate_drift_utils.py
```diff
@@ -78,21 +116,25 @@ def run_multivariable_drift(train_dataframe: pd.DataFrame, test_dataframe: pd.Da
...
         task_type=TaskType.BINARY,
         force_permutation=True,
-        permutation_kwargs={'n_repeats': 10,
-                            'random_state': random_state,
-                            'timeout': 120,
-                            'skip_messages': True}
+        permutation_kwargs={
+            "n_repeats": 10,
+            "random_state": random_state,
+            "timeout": feature_importance_timeout,
```
noamzbr, line 122: This is the second change that should be kept

deepchecks/core/check_utils/multivariate_drift_utils.py
```diff
@@ -16,58 +16,96 @@
...
+    random_state: int,
+    test_size: float,
+    n_top_columns: int,
+    min_feature_importance: float,
+    max_num_categories_for_display: int,
+    show_categories_by: str,
+    min_meaningful_drift_score: float,
+    with_display: bool,
+    dataset_names: Tuple[str] = DEFAULT_DATASET_NAMES,
+    feature_importance_timeout: int = 120,
```
noamzbr, line 62: This is the first change that should be kept

deepchecks/tabular/utils/feature_importance.py
```diff
@@ -21,30 +21,36 @@
 
 from deepchecks import tabular
 from deepchecks.core import errors
-from deepchecks.tabular.metric_utils.scorers import DeepcheckScorer, get_default_scorers, init_validate_scorers
+from deepchecks.tabular.metric_utils.scorers import (
```
noamzbr, line 24: Unless I'm missing something, all changes to this file should be reverted.

tests/tabular/checks/train_test_validation/multivariate_drift_test.py
```diff
@@ -168,4 +220,58 @@ def test_over_255_categories_in_column():
 
     # Assert
     # we only care that it runs
-    assert_that(result.value['domain_classifier_auc'])
+    assert_that(result.value["domain_classifier_auc"])
+
+
+def test_raise_zero_timeout(drifted_data):
```
noamzbr, line 226: These last two tests are the only changes that should be kept here.

noamzbr (changes requested):
Hi @IngoStatworx, really happy you found the time to write this PR.

The changes look good (also - extra points for writing the tests 😉). Only issue is that I think you've been using some kind of IDE auto-indent add-on that changed a lot of the formatting of the code in the files you touched. Could you please go over them and make sure only the relevant changes are kept? Thanks!

About installing the dev env - could you perhaps attach the error message you got? Could be in a new issue or here.

tests/tabular/checks/train_test_validation/multivariate_drift_test.py
IngoStatworx: Right, so I redid the branch, I hope it's still working

IngoStatworx:
> Hi @IngoStatworx, really happy you found the time to write this PR.
> 
> The changes look good (also - extra points for writing the tests 😉). Only issue is that I think you've been using some kind of IDE auto-indent add-on that changed a lot of the formatting of the code in the files you touched. Could you please go over them and make sure only the relevant changes are kept? Thanks!
> 
> About installing the dev env - could you perhaps attach the error message you got? Could be in a new issue or here.

I redid the PR, please check.


As for dev env, first, the make paths are not correct e.g.
```
/bin/sh: /bin/pip: No such file or directory
make: *** [requirements] Error 127
```

Fixing that, it mostly comes down to me being on Apple Arm64. So pip trying to install pinned version of larger packages like pytorch or lightgbm will fail. e.g.
```
####  installing dependencies, it could take some time, please wait! #### 
ERROR: Could not find a version that satisfies the requirement torchvision==0.11.3 (from versions: 0.1.6, 0.1.7, 0.1.8, 0.1.9, 0.2.0, 0.2.1, 0.2.2, 0.2.2.post2, 0.2.2.post3, 0.12.0, 0.13.0, 0.13.1, 0.14.0)
ERROR: No matching distribution found for torchvision==0.11.3
make: *** [requirements] Error 1
```

this is not really surprising, it's an Arm64 thing.

deepchecks/tabular/utils/feature_importance.py
```diff
@@ -309,6 +309,11 @@ def _calc_permutation_importance(
         elif not skip_messages:
             get_logger().info('Calculating permutation feature importance. Expected to finish in %s seconds',
                               predicted_time_to_run)
+    elif timeout == 0 and not skip_messages:
```
noamzbr, line 312: I'm not sure I understood the purpose of this change. User is supposed to set timeout == 0 in case they want to avoid calculating feature importance entirely, which should be possible.  Only thing we should take care of as far as I understand is that a correct error message / warning is shown in case that this happens. 

What is the issue you where trying to avoid by adding this code? What error was raised when timeout is set to 0?

noamzbr (commented):
Looks good, left one question about the timeout == 0 thing.

deepchecks/tabular/utils/feature_importance.py
IngoStatworx: The current logic means that if timeout is 0, then the task will always fail, raising a TimeOutError with the message that feature_importance would take too much time. This is the case since the variable predicted_time_to_run is at least 1.

However, this would confuse the user because the tutorials suggest that feature_importance can be skipped. Indeed, who would set timeout=0 and expect that it is indeed used to compare expected runtime?
Instead, I added the correct error message that timeout=0 is not implemented yet.


The alternative is to pass a "None" as feature_importance if timeout=0. In that case, if the model is not a Pipeline, then the downstream function might successfully extract a built-in feature importance (e.g. if model is a tree).
If that fails, however, the error will be nondescript ValueError with a similar confusion for the model - they request not to compute feature importance, and the error is "Was not able to calculate feature importance".

Bottom line for me: This functionality is not implemented right now, hence my choice.

I can do it either way, just let me know.

IngoStatworx: Somehow I missed the exception handling in calculate_feature_importance_or_none. Now it makes sense :D 

Well please have a look if I missed anything. Now both None, 0, and an integer timeout should hopefully work...

noamzbr (approved):
Looks good!

deepchecks/tabular/utils/feature_importance.py
```diff
@@ -309,6 +310,9 @@ def _calc_permutation_importance(
...
+            raise errors.DeepchecksSkippedFeatureImportance(
+                "feature_importance_timeout set to 0. Skipping any feature importance calculation.")
```
noamzbr, line 315: Change to `'` quotes, and fix indentation.

noamzbr (commented):
Hi @IngoStatworx, can you just fix the issues pylint found?

deepchecks/tabular/utils/feature_importance.py
IngoStatworx: Fixed. Sorry, I was sick over the weekend 👍

noamzbr (changes requested):
Hi @IngoStatworx, see my latest comment on the pylint thread.

deepchecks/tabular/utils/feature_importance.py
IngoStatworx: Wow I did it again.
For context, my dev env auto sets "format on save". For my next contribution, I am gonna set up a separate env.
sry

noamzbr (approved):

noamzbr:
@IngoStatworx nearly there! Alas pylint is a harsh mistress, and it requires you add another blank line above DeepchecksSkippedFeatureImportance (it requires two blank lines between independent definitions in a module)

shir22:
@all-contributors add @IngoStatworx for code, bug

@IngoStatworx many thanks for your contribution and for this important bugfix!

PR #2125, opened 2022-11-03
Make sure we know what the classes are when only proba are passed

### LINKED ISSUE #2090: [TABULAR] Design a way to infer the correct predicted class from probas
Currently in the tabular package predictions must be given / implemented, while predict_proba is optional. Theoretically (and done in practice in Vision) it's possible to get the predicted class from .argmax() on the proba. In tabular though, it's more complicated because classes don't have to be [0, .., n-1].

Two possible solutions:
1. Leave as is, tabular users must give predictions
2. Use the context.classes to get the classes that correspond to the proba columns. Downside is that it practically assumes that all classes are in the label.

noamzbr:
#### Reference Issues/PRs
Closes #2090 

If only proba are passed, make sure we either can see all the classes we need in the labels, or the user has passed model_classes.

### DIFF
--- a/deepchecks/tabular/context.py
+++ b/deepchecks/tabular/context.py
@@ -10,6 +10,7 @@
 #
 """Module for base tabular context."""
 import typing as t
+from collections.abc import Sequence
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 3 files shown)


deepchecks/tabular/context.py
```diff
@@ -203,6 +204,39 @@ def __init__(
...
+        if y_proba_test is not None and not isinstance(y_proba_test, (Sequence, np.ndarray)):
+            raise ValueError('y_proba_test must be a matrix with n_samples rows and n_classes columns')
+        if y_proba_test is not None and np.array(y_proba_test).shape[1] != np.array(y_proba_train).shape[1]:
+            raise DeepchecksValueError(
+                f'y_proba_test with {len(y_proba_test[0])} columns does not match y_proba_train with'
+                f' {len(y_proba_train[0])} columns.'
+            )
+        proba_shape = len(y_proba_train[0]) if y_proba_train is not None else None
+        if (y_proba_test is not None or y_proba_train is not None) and y_pred_train is None and y_pred_test is None:
+            if self._model_classes is None and len(self._observed_classes) != proba_shape:
```
matanper, line 232: A. We are doing this validation when initalizing scorers, not sure if we need to repeat it here
B. Why if inside if instead of end?

deepchecks/tabular/context.py
```diff
@@ -203,6 +204,39 @@ def __init__(
...
+        if y_proba_train is not None and not isinstance(y_proba_train, (Sequence, np.ndarray)):
+            raise ValueError('y_proba_train must be a matrix with n_samples rows and n_classes columns')
+        if y_proba_test is not None and not isinstance(y_proba_test, (Sequence, np.ndarray)):
+            raise ValueError('y_proba_test must be a matrix with n_samples rows and n_classes columns')
+        if y_proba_test is not None and np.array(y_proba_test).shape[1] != np.array(y_proba_train).shape[1]:
+            raise DeepchecksValueError(
+                f'y_proba_test with {len(y_proba_test[0])} columns does not match y_proba_train with'
+                f' {len(y_proba_train[0])} columns.'
+            )
+        proba_shape = len(y_proba_train[0]) if y_proba_train is not None else None
```
matanper, line 230: Why not proba.shape[1]?

deepchecks/tabular/context.py
```diff
@@ -203,6 +204,39 @@ def __init__(
...
+        if (y_proba_test is not None or y_proba_train is not None) and y_pred_train is None and y_pred_test is None:
+            if self._model_classes is None and len(self._observed_classes) != proba_shape:
+                raise DeepchecksValueError(
+                    f'The number of columns in the predicted probability matrix does not match the number of observed'
+                    f' classes in the data, and predictions were not passed so that inferring the existing classes is'
```
matanper, line 235: We do infer also based on labels, so predictions is not mandatory in order to infer

deepchecks/tabular/context.py
noamzbr: I don't know that it's a numpy array / pandas. Could be list of lists and the code will still run.

deepchecks/tabular/context.py
noamzbr: This is exactly what I'm saying here. I've checked len(self._observed_classes) which is the labels and found that it's length doesn't match that of the probas, so I need to resort to the user passing model_classes

deepchecks/tabular/context.py
noamzbr: A. I didn't know about it (assuming you're talking about line 273 in scorer.py). The only difference is that here it's a specific validation in case model_classes wasn't set, and in scorers.py the validation can be triggered either by it not being set or by it being set incorrectly. 
B. Makes it cleaner, rather than a huge if

deepchecks/tabular/context.py
matanper: you are saying "predictions were not passed so that inferring the existing classes is not possible".
That's not true, since we did do the infer using the labels, just it didn't match the probas

deepchecks/tabular/context.py
matanper: A. feels redundant, what happens currently is:
the user doesn't pass classes + the model doesn't have `classes_` -> if a scorer is needed we are issuing a warning "could not find model classes, using observed classes" -> if proba doesn't match classes we raise exception

PR #2126, opened 2022-11-03
Enable getting model predictions using only the proba and the model_classes

### LINKED ISSUE #2090: [TABULAR] Design a way to infer the correct predicted class from probas

noamzbr:
Closes #2090

### DIFF
--- a/deepchecks/tabular/context.py
+++ b/deepchecks/tabular/context.py
@@ -20,6 +20,7 @@ from deepchecks.core.errors import (DatasetValidationError, DeepchecksNotSupport
 from deepchecks.tabular._shared_docs import docstrings
 from deepchecks.tabular.dataset import Dataset
 from deepchecks.tabular.metric_utils import DeepcheckScorer, get_default_scorers, init_validate_scorers
+from deepchecks.tabular.metric_utils.scorers import validate_proba
 from deepchecks.tabular.utils.feature_importance import calculate_feature_importance_or_none
 from deepchecks.tabular.utils.task_inference import infer_task_type_and_classes
 from deepchecks.tabular.utils.task_type import TaskType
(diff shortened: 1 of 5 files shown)


deepchecks/tabular/context.py
```diff
@@ -88,7 +90,9 @@ def __init__(self,
             if dataset is not None:
                 feature_df_list.append(dataset.features_columns)
                 if y_pred is None and y_proba is not None:
+                    validate_proba(y_proba, model_classes)
                     y_pred = np.argmax(y_proba, axis=-1)
+                    y_pred = np.array(model_classes)[y_pred]
```
Nadav-Barak, line 95: Nice

tests/tabular/test_dummy_model.py
```diff
@@ -220,3 +220,12 @@ def test_suite(diabetes_split_dataset_and_model):
     result = suite.run(**args)
     length = get_expected_results_length(suite, args)
     validate_suite_result(result, length)
+
+
+def test_predict_using_proba(iris_binary_string_split_dataset_and_model):
```
Nadav-Barak, line 225: can compare the check result when running on probas vs running on predictions

Nadav-Barak (approved):

deepchecks/tabular/context.py
matanper: also possible `np.argmax(y_proba, axis=-1).map(model_classes.__getitem__)`

matanper (approved):

deepchecks/tabular/context.py
noamzbr: Mine is more elegant

PR #2130, opened 2022-11-06
Add mask dataset

noamzbr:
#### Reference Issues/PRs
Related to https://github.com/deepchecks/mon/issues/341

Add a version of the mask dataset with pre-calculated predictions and "time steps" that are different combinations of the same images.

### DIFF
--- a/deepchecks/vision/datasets/detection/mask.py
+++ b/deepchecks/vision/datasets/detection/mask.py
@@ -1,0 +1,292 @@
+# ----------------------------------------------------------------------------
+# Copyright (C) 2021-2022 Deepchecks (https://www.deepchecks.com)
+#
+# This file is part of Deepchecks.
+# Deepchecks is distributed under the terms of the GNU Affero General
+# Public License (version 3 or later).
+# You should have received a copy of the GNU Affero General Public License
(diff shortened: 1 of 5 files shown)


deepchecks/vision/datasets/detection/mask.py
```diff
@@ -0,0 +1,292 @@
...
+
+def load_dataset(
+        time_index: int = 0,
+        batch_size: int = 32,
+        num_workers: int = 0,
+        shuffle: bool = True,
+        pin_memory: bool = True,
+        object_type: Literal['VisionData', 'DataLoader'] = 'DataLoader'
+) -> t.Union[DataLoader, vision.VisionData]:
+    """Get the mask dataset and return a dataloader.
```
Nadav-Barak, line 107: Move the description of the dataset to top of the page
I think this way docstring compile better

Nadav-Barak (commented):
Better if @nirhutnik will also look

deepchecks/vision/datasets/detection/mask.py
noamzbr: Can do

ItayGabbay (approved):

PR #2136, opened 2022-11-10
change time_index to day_index

noamzbr:


deepchecks/vision/datasets/detection/mask.py
```diff
@@ -138,7 +138,9 @@ def load_dataset(
     root = DATA_DIR
     mask_dir = MaskDataset.download_mask(root)
     time_to_sample_dict = MaskDataset.get_time_to_sample_dict(root)
-    time = list(time_to_sample_dict.keys())[time_index]
+    if day_index > 59:
```
Nadav-Barak, line 141: can also check isinstance int

Nadav-Barak (approved):

PR #2138, opened 2022-11-14
Fix the x zoom in edge cases of drift

### LINKED ISSUE #1880: [BUG] Default Zoom displays in a problematic manner, when there are few close values
**Describe the bug**
See here:
https://docs.deepchecks.com/0.7/checks_gallery/vision/train_test_validation/plot_image_property_drift.html#observe-the-checks-output 
Scrolling a bit down will show:
CleanShot 2022-08-04 at 17 42 54
Instead of something like this (when double clicking):
CleanShot 2022-08-04 at 17 43 03

@matanper - I guess this relates to some of zoom default changes you did for nice categorical plots? (can we find a way in the middle so that both categorical and these types of plots will look reasonable)

noamzbr:
#### Reference Issues/PRs
Fixes #1880 

image


noamzbr:
Updated:
image
image

matanper (approved):

shir22 (approved):

PR #2156, opened 2022-11-20
Remove reduce from TrainTestPerformance

noamzbr:
Current reduce value is the performance on Test, which can be achieved by the SingleDatasetPerformance check. No need for that one only for it's display.

### DIFF
--- a/tests/tabular/checks/model_evaluation/train_test_performance_test.py
+++ b/tests/tabular/checks/model_evaluation/train_test_performance_test.py
@@ -27,6 +27,15 @@ from deepchecks.tabular.metric_utils.scorers import DEFAULT_REGRESSION_SCORERS, 
 from tests.base.utils import equal_condition_result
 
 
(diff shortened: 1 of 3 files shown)


Nadav-Barak (commented):
I think the tests were not there to only test the reduce method. I possible, convert the tests use the result.value instead of deleting them

tests/tabular/checks/model_evaluation/train_test_performance_test.py
```diff
@@ -155,42 +155,6 @@ def test_regression(diabetes_split_dataset_and_model):
             assert_that(metric_col['Value'].iloc[0], instance_of(float))
 
 
-def test_regression_reduced(diabetes_split_dataset_and_model):
```
Nadav-Barak, line 158: dont delete

ItayGabbay (approved):

Nadav-Barak (approved):

PR #2178, opened 2022-11-30
fixing test-windows 

ItayGabbay:


.github/workflows/build.yml
```diff
@@ -67,7 +67,7 @@ jobs:
     runs-on: windows-latest
     strategy:
       matrix:
-        python-version: ["3.8"] #TODO: add 3.10 when scikit-learn have a wheel https://github.com/scikit-learn/scikit-learn/pull/21232
+        python-version: ["3.9"] #TODO: add 3.10 when scikit-learn have a wheel https://github.com/scikit-learn/scikit-learn/pull/21232
```
noamzbr, line 70: Why is that?

ItayGabbay: No concrete reason. Thought it would be faster.

Issue #2160, opened 2022-11-21
[BUG] docstring for aggregation method talks about drift in unrelated checks


noamzbr:
**Describe the bug**
Have a look for example at 
https://docs.deepchecks.com/stable/api/generated/deepchecks.tabular.checks.data_integrity.PercentOfNulls.html#percentofnulls

Here the docstring for aggregation_method refers to drift, but the check isn't drift.
Can be found also in  CategoryMismatchTrainTest

PR #2186, opened 2022-12-06
feature_aggregation_method_argument is shared, so it shouldn't say drift

noamzbr:
#### Reference Issues/PRs

Closes https://github.com/deepchecks/deepchecks/issues/2160

### DIFF
--- a/deepchecks/tabular/_shared_docs.py
+++ b/deepchecks/tabular/_shared_docs.py
@@ -33,18 +33,18 @@ model_classes: Optional[List] , default: None
 """.strip('\n')
 
 _shared_docstrings['feature_aggregation_method_argument'] = """
-argument for the reduce_output functionality, decides how to aggregate the drift scores for a
+argument for the reduce_output functionality, decides how to aggregate the per-feature scores into a
 collective score. The collective score value is between 0 and 1 for all methods other than l2_combination.
 Possible values are:
-'l2_weighted': L2 norm over the combination of drift scores and feature importance, minus the
-L2 norm of feature importance alone, specifically, ||FI + DRIFT|| - ||FI||. This method returns a
+'l2_weighted': L2 norm over the combination of per-feature scores and feature importance, minus the
+L2 norm of feature importance alone, specifically, ||FI + SCORE|| - ||FI||. This method returns a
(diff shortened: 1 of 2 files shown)


deepchecks/tabular/_shared_docs.py
```diff
@@ -33,18 +33,18 @@
...
 
 _shared_docstrings['feature_aggregation_method_argument'] = """
-argument for the reduce_output functionality, decides how to aggregate the drift scores for a
+argument for the reduce_output functionality, decides how to aggregate the per-feature scores into a
 collective score. The collective score value is between 0 and 1 for all methods other than l2_combination.
 Possible values are:
-'l2_weighted': L2 norm over the combination of drift scores and feature importance, minus the
-L2 norm of feature importance alone, specifically, ||FI + DRIFT|| - ||FI||. This method returns a
+'l2_weighted': L2 norm over the combination of per-feature scores and feature importance, minus the
+L2 norm of feature importance alone, specifically, ||FI + SCORE|| - ||FI||. This method returns a
```
nirhutnik, line 40: I think that the "SCORE" here is unclear. If this can't be solved easily then maybe consider splitting the doc into 2 separate paragraphs.

noamzbr: What two paragraphs would you suggest? This section is preamble + description of the individual options.
Will rethink SCORE though

deepchecks/tabular/_shared_docs.py
```diff
@@ -33,18 +33,18 @@
 """.strip('\n')
 
 _shared_docstrings['feature_aggregation_method_argument'] = """
-argument for the reduce_output functionality, decides how to aggregate the drift scores for a
+argument for the reduce_output functionality, decides how to aggregate the per-feature scores into a
```
nirhutnik, line 36: Maybe add italics to the "per-feature scores" and remind that they are a vector of a score per-feature?

nirhutnik (approved):

deepchecks/tabular/_shared_docs.py
nirhutnik: Good enough, I think. See my second suggestion. Anyway approved

PR #2204, opened 2022-12-18
Using also y_pred to infer task type

matanper:


noamzbr (approved):

deepchecks/tabular/context.py
```diff
@@ -221,8 +221,16 @@ def __init__(
...
+
+        observed_labels, self._model_classes = get_labels_and_classes(
+            model, train, test, y_pred_train, y_pred_test, model_classes
+        )
+
+        self._task_type = infer_task_type(train, observed_labels, self._model_classes)
+        if self._task_type in (TaskType.BINARY, TaskType.MULTICLASS):
+            self._observed_classes = sorted(observed_labels.dropna().unique().tolist())
+        else:
+            self._observed_classes = None
```
nirhutnik, line 233: Why repeat this and not just use infer_task_type_and_classes?

nirhutnik (approved):

deepchecks/tabular/context.py
matanper: infer task type and classes is just a shortcut for other places, it gets only model and dataset

matanper: umm never mind, it's possible

PR #2215, opened 2022-12-19
Fixing errors from new numpy version

matanper:


requirements/requirements.txt
```diff
@@ -27,7 +27,7 @@ statsmodels>=0.11.0
 scipy>=1.4.1
 dataclasses>=0.6; python_version < '3.7'
 # jupyterlab-plotly extension that is used to display plotly figures
-plotly>=5.5.0
+plotly>=5.8.0
```
Nadav-Barak, line 30: why?

tests/conftest.py
```diff
@@ -183,7 +183,7 @@ def fillna(data: pd.DataFrame):
             'weird_feature': np.random.choice(np.array([1, 100, 1.0, 'ahh?', 'wee', np.nan, 0],
                                               dtype='object'), size=1000),
             8: pd.array(np.random.choice([0, 1, 5, 6, np.nan], size=1000), dtype='Int64'),
-            'tuples': np.random.choice([(0, 2), (1, 6, 8), (9, 1), (8, 1, 9, 8)], size=1000),
+            'tuples': random.choices([(0, 2), (1, 6, 8), (9, 1), (8, 1, 9, 8)], k=1000),
```
Nadav-Barak, line 186: does this work on older numpy versions as well?
we can also do np.random.choice(np.asarray([(0, 2), (1, 6, 8), (9, 1), (8, 1, 9, 8)], dtype='object'), size=1000)

Nadav-Barak (approved):

tests/conftest.py
matanper: I didn't really understand the suggestion, on new numpy it raises error because the items inside the array are in different lengths from each other

requirements/requirements.txt
matanper: they also used the deprecated numpy things before 5.8

ItayGabbay (approved):

PR #2223, opened 2022-12-22
fix problems with missing classes

JKL98ISR:


deepchecks/tabular/metric_utils/additional_classification_metrics.py
```diff
@@ -50,7 +50,7 @@ def _false_positive_rate_per_class(y_true, y_pred, classes):  # False Positives
     result = []
     for cls in classes:
         y_true_cls, y_pred_cls = np.asarray(y_true) == cls, np.asarray(y_pred) == cls
-        matrix = confusion_matrix(y_true_cls, y_pred_cls)
+        matrix = confusion_matrix(y_true_cls, y_pred_cls, labels=[0, 1])
```
ItayGabbay, line 53: is [0,1] always the case?

deepchecks/tabular/metric_utils/scorers.py
```diff
@@ -122,7 +122,7 @@
     'tnr': make_scorer(true_negative_rate_metric, averaging_method='binary'),
     'jaccard': make_scorer(jaccard_score, zero_division=0),
     'roc_auc': get_scorer('roc_auc'),
-    'neg_log_loss': get_scorer('neg_log_loss')
+    'neg_log_loss': make_scorer(log_loss, greater_is_better=False, needs_proba=True, labels=[0, 1])
```
ItayGabbay, line 125: is it still `neg_log_loss` after the `greater_is_better=False`?

deepchecks/tabular/metric_utils/additional_classification_metrics.py
```diff
@@ -59,7 +59,7 @@ def _micro_false_positive_rate(y_true, y_pred, classes):
     fp, tn = 0, 0
     for cls in classes:
         y_true_cls, y_pred_cls = np.asarray(y_true) == cls, np.asarray(y_pred) == cls
-        matrix = confusion_matrix(y_true_cls, y_pred_cls)
+        matrix = confusion_matrix(y_true_cls, y_pred_cls, labels=[0, 1])
```
Nadav-Barak, line 62: I think is always not true..

Nadav-Barak (changes requested):
Lets talk

deepchecks/tabular/metric_utils/additional_classification_metrics.py
Nadav-Barak: yes

deepchecks/tabular/metric_utils/scorers.py
Nadav-Barak: + I am not sure if log_loss is the same as neg_log_loss, If not - its bad:
I think can create weird behaviour in some of our checks, we require that scorers also follow the convention that greater is better. even written in our docs

matanper: I think we need to wrap the log_loss in lambda:
```
def neg_log_loss(y, y_pred, **kwargs):
  return -log_loss(y, y, y_pred, **kwargs)
```

JKL98ISR: this is how sklearn defines neg_log_loss
you can see it by printing get_scorer('neg_log_loss')
I only added the labels

deepchecks/tabular/metric_utils/scorers.py
```diff
@@ -297,20 +298,27 @@ def _run_score(self, model, data: pd.DataFrame, label_col):
...
-                    raise errors.DeepchecksValueError(
-                        f'Scorer returned {len(scores)} scores, but model contains '
-                        f'{len(self.model_classes)} classes. Can\'t proceed')
-                scores = dict(zip(self.model_classes, scores))
-                # Add classes which been seen in the data but are not known to the model
-                scores.update({cls: np.nan for cls in set(self.observed_classes) - set(self.model_classes)})
-            return scores
-        else:
-            return self.scorer(model, data, label_col)
+        # The scores returned are for the observed classes but we want scores of the observed classes
```
matanper, line 311: Can you fix?
`The scores returned are for the observed classes but we want scores of the observed classes` -> `The scores returned are for the observed classes but we want scores of the model classes`

matanper (approved):

PR #2224, opened 2022-12-25
torch hub coco - skip github validation

matanper:
Attempt to fix rate limit exception


Nadav-Barak (approved):

JKL98ISR:
I fixed it in a previous pr using `torch.hub._validate_not_a_forked_repo = lambda *_: True  # pylint: disable=protected-access`
do we want this one too?

matanper:
which previous PR? currently it wasn't there
Anyway nadav is about to take torch out I believe in his refactor

JKL98ISR:
was in #2223 but yeah if we take torch out this seems good

PR #2244, opened 2023-01-03
common calculation for default properties

Nadav-Barak:
Reduce run time of all checks that use properties 
Also solves small bug when a single check in the suite has custom properties


deepchecks/vision/suites/default_suites.py
```diff
@@ -293,8 +293,23 @@ def data_integrity(image_properties: List[Dict[str, Any]] = None,
...
+        Number of samples to use for the checks in the suite. If None, all samples will be used.
+
+    Returns
+    -------
+    Suite
+        A suite that includes integrity checks.
+    """
+    args = locals()
+    args.pop('kwargs')
+    non_none_args = {k: v for k, v in args.items() if v is not None}
```
noamzbr, line 311: What items can be here except n_samples? Why not write it explicitly?

Nadav-Barak: Youre right. Added the rest of them

noamzbr (approved):

deepchecks/vision/checks/data_integrity/abstract_property_outliers.py
```diff
@@ -94,9 +96,9 @@ def initialize_run(self, context: Context, dataset_kind: DatasetKind):
         self._images_uuid = []
 
         self.properties_list = self.properties_list if self.properties_list else self.get_default_properties(data)
-        if any(p['output_type'] == 'class_id' for p in self.properties_list):
+        if self.properties_list is not None and any(p['output_type'] == 'class_id' for p in self.properties_list):
```
matanper, line 99: when is it none?

matanper (approved):

PR #2280, opened 2023-01-13
[issue-2253][Vision Renewed] vision conftest fix

### LINKED ISSUE #2253: [DEE-19] [FEAT] seperate conftest into core tabular and vision
This is required since only the tabular subpackage should support python 3.6 but since the conftest are united its impossible to run the tabular tests without installing the vision requirements (which is not possible in python 3.6)

**This fix should be implemented in the Vision renewed branch** and DoD should be that the CI passes for the python 3.6

<sub>[DEE-19](https://linear.app/deepchecks/issue/DEE-19/[feat]-seperate-conftest-into-core-tabular-and-vision)</sub>

yromanyshyn:
resolves #2253


yromanyshyn:
@Nadav-Barak 

NOTE:
tests fail with not related to this issue problem
but rather related to the `Vision Renewed` in general
also, tabular tests passed
```
FAILED tests/vision/checks/train_test_validation/property_label_correlation_change_test.py::test_drift_object_detection_tf - AssertionError: 
Expected: a dictionary containing {'test': a dictionary containing {'Aspect Ratio': a numeric value within <0.01> of <0.049>}, 'train': a dictionary containing {'Aspect Ratio': a numeric value within <0.01> of <0.067>}, 'train-test difference': a dictionary containing {'Aspect Ratio': a numeric value within <0.01> of <0.018>}}
     but: value for 'test' value for 'Aspect Ratio' <0.06647253085793685> differed by <0.017472530857936847>
```

Nadav-Barak (approved):
I am pleasantly surprised of how simple the solution is!
If you could also split the remaining tests/conftest and move all that you can to tests/tabular/conftest.py that would be great (as part of future effort to split core and tabular)

yromanyshyn:
@Nadav-Barak 

> I am pleasantly surprised of how simple the solution is! If you could also split the remaining tests/conftest and move all that you can to tests/tabular/conftest.py that would be great (as part of future effort to split core and tabular)

yeap, will do, should not be hard

yromanyshyn:
@Nadav-Barak 

> If you could also split the remaining tests/conftest and move all that you can to tests/tabular/conftest.py that would be great (as part of future effort to split core and tabular)


I moved fixtures that are used exclusively by tabular tests to `tabular/conftest.py`

Nadav-Barak (approved):

PR #2281, opened 2023-01-15
Improve context performance

matanper:
1. prevent calculating observed classes at all costs until needed
2. add an option to pass task_type and observed_classes from outside
3. try to change `convert_into_flat_list` because it's slow

### DIFF
--- a/deepchecks/tabular/_shared_docs.py
+++ b/deepchecks/tabular/_shared_docs.py
@@ -30,6 +30,11 @@ y_proba_test: Optional[np.ndarray] , default: None
     Array of the model prediction probabilities over the test dataset.
 model_classes: Optional[List] , default: None
     For classification: list of classes known to the model
+task_type: Optional[TaskType] , default: None,
+    Current task type
+observed_classes: Optional[t.List] , default: None
+    For classification: list of all observed classes in the data (also ones which are unknown to the model)
+
 """.strip('\n')
 
 _shared_docstrings['feature_aggregation_method_argument'] = """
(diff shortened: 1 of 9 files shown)


deepchecks/tabular/_shared_docs.py
```diff
@@ -30,6 +30,11 @@
     Array of the model prediction probabilities over the test dataset.
 model_classes: Optional[List] , default: None
     For classification: list of classes known to the model
+task_type: Optional[TaskType] , default: None,
+    Current task type
+observed_classes: Optional[t.List] , default: None
```
Nadav-Barak, line 35: Really confusing argument

deepchecks/tabular/_shared_docs.py
```diff
@@ -30,6 +30,11 @@
     Array of the model prediction probabilities over the test dataset.
 model_classes: Optional[List] , default: None
     For classification: list of classes known to the model
+task_type: Optional[TaskType] , default: None,
```
Nadav-Barak, line 33: Currently we ask for task type in Dataset creation, it cannot be in both places

deepchecks/tabular/context.py
```diff
@@ -221,16 +223,35 @@ def __init__(
...
-        if model is None and \
-                not pd.Series([y_pred_train, y_pred_test, y_proba_train, y_proba_test]).isna().all():
+        if model_classes is None:
+            model_classes = infer_classes_from_model(model)
+        labels = None
+        if task_type is None:
+            labels = get_all_labels(model, train, test, y_pred_train, y_pred_test)
+            task_type = infer_task_type(train, labels, model_classes)
+
+        need_observed_classes_for_dummy_model = \
```
Nadav-Barak, line 233: why not place lines 233 up to 241 right before the dummy model creation? could simplify this boolean variable a bit

nirhutnik (approved):

noamzbr (approved):

PR #2289, opened 2023-01-17
Add head() to VisionData

matanper:


deepchecks/vision/vision_data/utils.py
```diff
@@ -39,12 +39,27 @@ def values(cls):
...
+class LabelMap(dict):
+    """Smarter dict for label map."""
+
+    def __init__(self, seq=None, **kwargs):
+        seq = seq or {}
+        super().__init__(seq, **kwargs)
+
+    def __getitem__(self, class_id: int) -> str:
+        """Return the name of the class with the given id."""
+        class_id = int(class_id)
```
Nadav-Barak, line 59: liked the class, would surround the int(class_id) with try catch

deepchecks/vision/vision_data/vision_data.py
```diff
@@ -266,3 +264,91 @@ def __iter__(self):
...
+
+        Parameters
+        ----------
+        num: int, default = 5
+            Number of images to show. Does not show more images than the size of single batch
+        """
+        if not is_notebook():
+            print('head function is supported only inside a notebook', file=sys.stderr)
+            return
+        if num < 1:
```
Nadav-Barak, line 279: should be < 2 + can also validate its an integer

Nadav-Barak (commented):
Really liked many of the changes!
Can you send a pdf of how the head will look like? 
Will it appear when we will use it in our docs?

deepchecks/vision/vision_data/vision_data.py
matanper: Why? the user can't show 1 image?

noamzbr: 1 image is OK, I'm with Matan

Nadav-Barak: 1 is better, so change the text to match it (it says larger 1 is a must)

deepchecks/vision/vision_data/vision_data.py
```diff
@@ -266,3 +264,91 @@ def __iter__(self):
     def __len__(self):
         """Return the number of batches in the batch loader if it is known, otherwise returns None."""
         return len(self._batch_loader) if hasattr(self._batch_loader, '__len__') else None
+
+    def head(self, num: int = 5):
```
Nadav-Barak, line 268: can change variable name to num_images_to_display

PR #2317, opened 2023-02-01
initial commit for imbalanced datasets

noamzbr:
The current drift methods are un-sensitive to highly imbalanced labels. This method is intended to calculate PSI while giving all classes equal weight, looking only at the relative change.

### DIFF
--- a/deepchecks/tabular/checks/model_evaluation/train_test_prediction_drift.py
+++ b/deepchecks/tabular/checks/model_evaluation/train_test_prediction_drift.py
@@ -50,6 +50,9 @@ class TrainTestPredictionDrift(TrainTestCheck, ReduceMixin):
     small number of samples (common practice is categories with less than 5 samples).
     However, in cases of a variable with many categories with few samples, it is still recommended to use Cramer's V.
 
+    **Note:** In case of highly imbalanced classes, it is recommended to use the PSI method, together with setting
+    the ``relative_change`` parameter to ``True``.
+
 
     Parameters
     ----------
@@ -82,6 +85,11 @@ class TrainTestPredictionDrift(TrainTestCheck, ReduceMixin):
     categorical_drift_method: str, default: "cramer_v"
         decides which method to use on categorical variables. Possible values are:
         "cramer_v" for Cramer's V, "PSI" for Population Stability Index (PSI).
+    relative_change: bool, default: False
+        If True, small and large categories will contribute equally to the PSI score, with only the log of the ratio
+        between the expected and actual category size being used. This is useful when we care about the relative
+        difference between the distributions, but not about the absolute difference. Usable only when
+        categorical_drift_method = "psi"
     ignore_na: bool, default True
         For categorical columns only. If True, ignores nones for categorical drift. If False, considers none as a
         separate category. For numerical columns we always ignore nones.
@@ -107,10 +115,11 @@ class TrainTestPredictionDrift(TrainTestCheck, ReduceMixin):
             drift_mode: str = 'auto',
             margin_quantile_filter: float = 0.025,
             max_num_categories_for_drift: int = None,
-            min_category_size_ratio: float = 0.01,
+            min_category_size_ratio: float = 1e-6,
             max_num_categories_for_display: int = 10,
             show_categories_by: str = 'largest_difference',
             categorical_drift_method: str = 'cramer_v',
+            relative_change: bool = False,
             ignore_na: bool = True,
             aggregation_method: t.Optional[str] = 'max',
             max_classes_to_display: int = 3,
@@ -130,6 +139,11 @@ class TrainTestPredictionDrift(TrainTestCheck, ReduceMixin):
         self.max_num_categories_for_display = max_num_categories_for_display
         self.show_categories_by = show_categories_by
         self.categorical_drift_method = categorical_drift_method
+        if categorical_drift_method != 'psi' and relative_change:
+            raise DeepchecksValueError(
+                'relative_change can only be used with categorical_drift_method = "psi"'
+            )
+        self.relative_change = relative_change
         self.ignore_na = ignore_na
         self.max_classes_to_display = max_classes_to_display
         self.aggregation_method = aggregation_method
@@ -188,6 +202,7 @@ class TrainTestPredictionDrift(TrainTestCheck, ReduceMixin):
                 max_num_categories_for_display=self.max_num_categories_for_display,
                 show_categories_by=self.show_categories_by,
                 categorical_drift_method=self.categorical_drift_method,
+                relative_change=self.relative_change,
                 ignore_na=self.ignore_na,
                 with_display=context.with_display,
             )
--- a/deepchecks/tabular/checks/train_test_validation/train_test_feature_drift.py
+++ b/deepchecks/tabular/checks/train_test_validation/train_test_feature_drift.py
@@ -83,6 +83,11 @@ class TrainTestFeatureDrift(TrainTestCheck, ReduceFeatureMixin):
     categorical_drift_method: str, default: "cramer_v"
         decides which method to use on categorical variables. Possible values are:
         "cramer_v" for Cramer's V, "PSI" for Population Stability Index (PSI).
+    relative_change: bool, default: False
+        If True, small and large categories will contribute equally to the PSI score, with only the log of the ratio
+        between the expected and actual category size being used. This is useful when we care about the relative
+        difference between the distributions, but not about the absolute difference. Usable only when
+        categorical_drift_method = "psi"
     ignore_na: bool, default True
         For categorical columns only. If True, ignores nones for categorical drift. If False, considers none as a
         separate category. For numerical columns we always ignore nones.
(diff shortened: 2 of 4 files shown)


deepchecks/tabular/checks/model_evaluation/train_test_prediction_drift.py
```diff
@@ -107,10 +115,11 @@ def __init__(
             drift_mode: str = 'auto',
             margin_quantile_filter: float = 0.025,
             max_num_categories_for_drift: int = None,
-            min_category_size_ratio: float = 0.01,
+            min_category_size_ratio: float = 1e-6,
             max_num_categories_for_display: int = 10,
             show_categories_by: str = 'largest_difference',
             categorical_drift_method: str = 'cramer_v',
+            relative_change: bool = False,
```
nirhutnik, line 122: name suggestions:
``balance_classes``
``weighted_classes``

I think current name doesn't really tell me much :/

deepchecks/tabular/checks/model_evaluation/train_test_prediction_drift.py
```diff
@@ -130,6 +139,11 @@ def __init__(
         self.max_num_categories_for_display = max_num_categories_for_display
         self.show_categories_by = show_categories_by
         self.categorical_drift_method = categorical_drift_method
+        if categorical_drift_method != 'psi' and relative_change:
```
nirhutnik, line 142:
```suggestion
        if categorical_drift_method != 'psi' and relative_change is True:
```

deepchecks/tabular/checks/model_evaluation/train_test_prediction_drift.py
```diff
@@ -82,6 +85,11 @@ class TrainTestPredictionDrift(TrainTestCheck, ReduceMixin):
     categorical_drift_method: str, default: "cramer_v"
         decides which method to use on categorical variables. Possible values are:
         "cramer_v" for Cramer's V, "PSI" for Population Stability Index (PSI).
+    relative_change: bool, default: False
+        If True, small and large categories will contribute equally to the PSI score, with only the log of the ratio
+        between the expected and actual category size being used. This is useful when we care about the relative
+        difference between the distributions, but not about the absolute difference. Usable only when
```
nirhutnik, line 91:
```suggestion
        difference between the distributions, but not about the absolute difference. Must have
```

deepchecks/tabular/checks/model_evaluation/train_test_prediction_drift.py
```diff
@@ -82,6 +85,11 @@ class TrainTestPredictionDrift(TrainTestCheck, ReduceMixin):
     categorical_drift_method: str, default: "cramer_v"
         decides which method to use on categorical variables. Possible values are:
         "cramer_v" for Cramer's V, "PSI" for Population Stability Index (PSI).
+    relative_change: bool, default: False
+        If True, small and large categories will contribute equally to the PSI score, with only the log of the ratio
```
nirhutnik, line 89:
```suggestion
        If True, all categories will have an equal weight in the PSI score, with only the log of the ratio
```

deepchecks/tabular/checks/train_test_validation/train_test_feature_drift.py
```diff
@@ -83,6 +83,11 @@ class TrainTestFeatureDrift(TrainTestCheck, ReduceFeatureMixin):
     categorical_drift_method: str, default: "cramer_v"
         decides which method to use on categorical variables. Possible values are:
         "cramer_v" for Cramer's V, "PSI" for Population Stability Index (PSI).
+    relative_change: bool, default: False
```
nirhutnik, line 86: Are you sure we want to add this to feature drift? Do you think that in *features* we may have a situation where all categories should weigh the same?

deepchecks/tabular/checks/train_test_validation/train_test_feature_drift.py
```diff
@@ -232,9 +244,7 @@ def run_logic(self, context: Context) -> CheckResult:
                 sorted_by = 'drift score'
 
             headnote = [f"""<span>
-                The Drift score is a measure for the difference between two distributions, in this check - the test
-                and train distributions.<br> The check shows the drift score and distributions for the features, sorted
-                by {sorted_by} and showing only the top {self.n_top_columns} features, according to {sorted_by}.
+                TESTTEST
```
nirhutnik, line 247: Well then

deepchecks/tabular/checks/model_evaluation/train_test_prediction_drift.py
```diff
@@ -107,10 +115,11 @@ def __init__(
             drift_mode: str = 'auto',
             margin_quantile_filter: float = 0.025,
             max_num_categories_for_drift: int = None,
-            min_category_size_ratio: float = 0.01,
+            min_category_size_ratio: float = 1e-6,
```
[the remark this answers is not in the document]
nirhutnik: Wait, this actually did not affect PSI so far, and if you're using it for this then you're changing its meaning

deepchecks/utils/distribution/drift.py
```diff
@@ -162,9 +166,12 @@ def psi(dist1: Union[np.ndarray, pd.Series], dist2: Union[np.ndarray, pd.Series]
     psi_value = 0
     for i in range(len(expected_counts)):
         # In order for the value not to diverge, we cap our min percentage value
-        e_perc = max(expected_counts[i] / size_expected, PSI_MIN_PERCENTAGE)
-        a_perc = max(actual_counts[i] / size_actual, PSI_MIN_PERCENTAGE)
-        value = (e_perc - a_perc) * np.log(e_perc / a_perc)
```
nirhutnik, line 167: problematic change here. ``min_category_size_ratio`` is used for the ``preprocess_2_cat_cols_to_same_bins`` thingy. First of all, it means that this parameter is in face meaningless here, as after the preprocessing you only get categories that are > ``min_category_size_ratio``.

But also, you're using the same parameter for 2 different functionalities, even if they are very similar

deepchecks/utils/distribution/drift.py
```diff
@@ -274,6 +282,9 @@ def calc_drift_and_plot(train_column: pd.Series,
     categorical_drift_method: str, default: "cramer_v"
         decides which method to use on categorical variables. Possible values are:
         "cramer_v" for Cramer's V, "PSI" for Population Stability Index (PSI).
+    relative_change: bool, default: False
+        If True, the drift score will be use the relative change between the two distributions, disregarding the
+        absolute values. In this case, the plot will be created with a log scale in the y-axis.
```
nirhutnik, line 287: Why log scale though? Can you show how it looks like?

docs/source/user-guide/general/drift_guide.rst
```diff
@@ -118,6 +118,23 @@ In general, it is recommended to use Cramer's V, unless your variable includes c
 However, in cases of a variable with many categories with few samples, it is still recommended to use Cramer's V, as PSI will not be able to detect change in the smaller categories.
 
 
+Detecting Drift in Unbalanced Classification
```
nirhutnik, line 121:
```suggestion
Detecting Drift in Unbalanced Classification Tasks
```

docs/source/user-guide/general/drift_guide.rst
```diff
@@ -118,6 +118,23 @@ In general, it is recommended to use Cramer's V, unless your variable includes c
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+In classification problems, it is common to have unbalanced data, meaning that the number of samples in each class is
+highly skewed. For example, in a dataset of credit card transactions, the number of fraudulent transactions is usually
+much lower than the number of non-fraudulent transactions, and can be way below 1% of the total number of samples.
+
+In such cases, running the :doc:`TrainTestLabelDrift </checks_gallery/tabular/train_test_validation/plot_train_test_label_drift>`:
+or :doc:`TrainTestPredictionDrift </checks_gallery/tabular/train_test_validation/plot_train_test_prediction_drift>` checks
+with the default parameters will likely lead to a false negative, as for example a change in the percent of fraudulent
```
nirhutnik, line 130: I think "false negative" is wrong here. It's not false, it's just not what it's looking for,.

docs/source/user-guide/general/drift_guide.rst
```diff
@@ -118,6 +118,23 @@ In general, it is recommended to use Cramer's V, unless your variable includes c
...
+transactions from 0.2% to 0.4% will not be detected, but may in fact be very significant for our business.
+
+To detect this kind of drift, set the ``relative_change`` parameter of these checks to True. This will cause the check
+to consider all classes equally, regardless of their size.
+
+**Note**: You must also set the ``categorical_drift_method`` parameter to ``'psi'`` in order to use this method.
```
nirhutnik, line 136:
```suggestion
**Note**: You must also set the ``categorical_drift_method`` parameter to ``'PSI'`` in order to use this method.
```

deepchecks/utils/distribution/drift.py
noamzbr: image

deepchecks/utils/distribution/drift.py
noamzbr: That's what I was thinking... why do we need the PSI_MIN_PERCENTAGE to be different than min_category_size_ratio? I get that we don't want someone to set min_category_size_ratio=0 and get inf values, maybe limit it so that min_category_size_ratio can't be 0?

noamzbr:
Closed by way of #2329

PR #2322, opened 2023-02-05
Improve vision data head display

matanper:
Using `DisplayableResult` from `core` in order to give better options for the head display like `show_in_window`:

```
from deepchecks.vision.datasets.classification.mnist_torch import load_dataset
load_dataset(object_type='VisionData').head().show_in_window()
```


Nadav-Barak (commented):
Can we make it an argument? like .head(show_in_window=True)

matanper:
> Can we make it an argument? like .head(show_in_window=True)

Not relevant to this PR, it's coming from `DisplayableResult` which also affects the check result display

yromanyshyn (approved):

PR #2326, opened 2023-02-12
Noam/mon 1644 enable f1 as a default metric

noamzbr:

1. Add f1 to binary tasks
2. Even in binary, I may want to use multiclass metrics to get all kinds of metrics that consider both classes (e.g., I may want to use f1_macro to consider also a very un-common 0 class).


deepchecks/tabular/metric_utils/scorers.py
```diff
@@ -152,15 +137,35 @@
...
     'tnr_weighted': make_scorer(true_negative_rate_metric, averaging_method='weighted'),
+    'jaccard_macro': make_scorer(jaccard_score, average='macro', zero_division=0),
+    'jaccard_micro': make_scorer(jaccard_score, average='micro', zero_division=0),
+    'jaccard_weighted': make_scorer(jaccard_score, average='weighted', zero_division=0),
+    'jaccard_per_class': make_scorer(jaccard_score, average=None, zero_division=0),
+}
+
+binary_scorers_dict = {
+    'balanced_accuracy': get_scorer('balanced_accuracy'),
+    'average_precision': get_scorer('average_precision'),
```
Nadav-Barak, line 148: average precision is not precision marco?

deepchecks/tabular/metric_utils/scorers.py
```diff
@@ -152,15 +137,35 @@
...
+    'fnr': make_scorer(false_negative_rate_metric, averaging_method='binary'),
+    'tnr': make_scorer(true_negative_rate_metric, averaging_method='binary'),
+    'jaccard': make_scorer(jaccard_score, zero_division=0),
+    'roc_auc': get_scorer('roc_auc'),
+    'neg_log_loss': make_scorer(log_loss, greater_is_better=False, needs_proba=True, labels=[0, 1]),
+    **common_classification_metrics
+}
+
+multiclass_scorers_dict = {
     'roc_auc_per_class': make_scorer(roc_auc_per_class, needs_proba=True),
```
Nadav-Barak, line 163: maybe move all per class methods here?

Nadav-Barak: will look better for monitoring

Nadav-Barak (approved):

deepchecks/tabular/metric_utils/scorers.py
noamzbr: No it's the average over the precision recall curve if I remember correctly

deepchecks/tabular/metric_utils/scorers.py
noamzbr: But it think the user should be able to have per class also for the binary case

Nadav-Barak: I think is a rarer case.
For the case you described - It is most likely representable by other metrics for the positive class
For the case I mentioned - I can be confusing

Anyhow I do not have a strong preference

noamzbr: mm good point, will move them to multiclass only

PR #2331, opened 2023-02-15
added exception or drift_mode!=prediction and balance_classes=True for prediction drift

nirhutnik:


noamzbr (commented):
Update this also in vision

noamzbr (approved):

PR #2338, opened 2023-02-19
Nb/nlp/tweet use case v2

Nadav-Barak:
small additions to support info display per cluster


deepchecks/nlp/text_data.py
```diff
@@ -316,15 +316,15 @@ def task_type(self) -> t.Optional[TaskType]:
         return self._task_type
 
     @property
-    def text(self) -> t.Sequence[str]:
+    def text(self) -> pd.Series:
```
noamzbr, line 319: This will have significant memory overhead...

deepchecks/nlp/utils/embeddings_display.py
```diff
@@ -149,11 +150,11 @@ def create_drift_files(train_text: pd.Series, test_text: pd.Series, train_embedd
     train_dataframe['sample origin'] = 'train'
     test_dataframe['sample origin'] = 'test'
     all_data = pd.concat([train_dataframe, test_dataframe])
+    if additional_data is not None:
+        additional_data = additional_data.loc[all_data.index]
+        all_data = all_data.join(additional_data)
```
noamzbr, line 155: Think you can just join by index (right / left join)

PR #2347, opened 2023-02-20
small updates to vision data head

matanper:


deepchecks/vision/vision_data/vision_data.py
```diff
@@ -265,19 +265,21 @@ def __len__(self):
...
         Parameters
         ----------
         num_images_to_display: int, default = 5
             Number of images to show. Does not show more images than the size of single batch
+        show_in_window: bool, default = False
+            Whether to open the head display in a new python window. requires pyqt5, pyqtwebengine libraries.
         """
-        if not (is_notebook() or is_sphinx()):
-            print('head function is supported only inside a notebook', file=sys.stderr)
+        if not (is_notebook() or is_sphinx()) and show_in_window is False:
```
Nadav-Barak, line 278: does show in window work on Jupiter notebook? 
If not then we should add a print for that case

Nadav-Barak (approved):

PR #2348, opened 2023-02-21
Added validation and normalization of feature importance

nirhutnik:


noamzbr (approved):
Also document this behavior in the aggregation method

PR #2354, opened 2023-02-22
balance distributions in cramers v

Nadav-Barak:
Balance the distribution to make cramers v more consistant with different sample size of the same distributions
Changed link to reference since old link was for code with bug

### DIFF
--- a/deepchecks/core/reduce_classes.py
+++ b/deepchecks/core/reduce_classes.py
@@ -97,7 +97,7 @@ class ReduceFeatureMixin(ReduceMixin):
         return False
 
     @staticmethod
-    def feature_reduce(aggregation_method: str, value_per_feature: pd.Series, feature_importance: Optional[np.array],
+    def feature_reduce(aggregation_method: str, value_per_feature: pd.Series, feature_importance: Optional[pd.Series],
                        score_name: str) -> Dict[str, float]:
         """Return an aggregated drift score based on aggregation method defined."""
         if feature_importance is not None:
@@ -111,18 +111,20 @@ class ReduceFeatureMixin(ReduceMixin):
         elif aggregation_method == 'max':
             return {str('Max ' + score_name): np.max(value_per_feature)}
 
-        if aggregation_method in ['weighted', 'l2_weighted'] and feature_importance is None:
-            get_logger().warning(
-                'Failed to calculate feature importance to all features, using uniform mean instead.')
-            return {str(str.title(aggregation_method.replace('_', ' ')) + ' ' + score_name): np.mean(value_per_feature)}
-
-        elif aggregation_method == 'weighted':
-            return {str('Weighted ' + score_name): np.sum(np.array(value_per_feature) * feature_importance)}
-        elif aggregation_method == 'l2_weighted':
-            sum_drift_fi = np.array(value_per_feature) + feature_importance
-            return {str('L2 Weighted ' + score_name): np.linalg.norm(sum_drift_fi) - np.linalg.norm(feature_importance)}
-        else:
+        if aggregation_method not in ['weighted', 'l2_weighted', 'l3_weighted']:
             raise DeepchecksValueError(f'Unknown aggregation method: {aggregation_method}')
+        elif feature_importance is None:
+            get_logger().warning('Failed to calculate feature importance, using uniform mean instead.')
+            return {str(str.title(aggregation_method.replace('_', ' ')) + ' ' + score_name): np.mean(value_per_feature)}
+        else:
+            value_per_feature, feature_importance = np.asarray(value_per_feature), np.asarray(feature_importance)
+
+        if aggregation_method == 'weighted':
+            return {str('Weighted ' + score_name): np.sum(value_per_feature * feature_importance)}
+        elif aggregation_method == 'l2_weighted':
+            return {str('L2 Weighted ' + score_name): np.sum((value_per_feature ** 2) * feature_importance) ** (1. / 2)}
+        elif aggregation_method == 'l3_weighted':
+            return {str('L3 Weighted ' + score_name): np.sum((value_per_feature ** 3) * feature_importance) ** (1. / 3)}
 
 
 class ReducePropertyMixin(ReduceMixin):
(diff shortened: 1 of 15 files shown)


deepchecks/core/reduce_classes.py
```diff
@@ -101,7 +101,7 @@ def feature_reduce(aggregation_method: str, value_per_feature: pd.Series, featur
                        score_name: str) -> Dict[str, float]:
         """Return an aggregated drift score based on aggregation method defined."""
         if feature_importance is not None:
-            feature_importance = np.array(feature_importance)[value_per_feature.notna().values]
+            feature_importance = feature_importance[value_per_feature.notna().values]
```
nirhutnik, line 104: Did this because for drift, for instance, feature_importance is returned as a list instead of an array, and wanted to enforce this behavior. Are you sure the places in the code where it's a list have been changed?

deepchecks/utils/distribution/drift.py
```diff
@@ -131,23 +131,23 @@ def cramers_v(dist1: Union[np.ndarray, pd.Series], dist2: Union[np.ndarray, pd.S
...
-
-    # If columns have the same single value in both (causing division by 0), return 0 drift score:
-    if contingency_matrix.shape[1] == 1:
-        return 0
+    # Down sample one of the distributions so that they have the same total number of samples
+    dist1_sum, dist2_sum = sum(dist1_counts), sum(dist2_counts)
+    if dist1_sum > dist2_sum:
+        dist1_counts = np.round(dist1_counts * (dist2_sum / dist1_sum))
+    elif dist1_sum < dist2_sum:
+        dist2_counts = np.round(dist2_counts * (dist1_sum / dist2_sum))
```
nirhutnik, line 147: See in the comments of the issue - this is problematic for the imbalanced case and for some reason does not work well there. So consider either taking a deeper look into it or just enforcing this for the balanced case

deepchecks/utils/distribution/drift.py
```diff
@@ -131,23 +131,23 @@ def cramers_v(dist1: Union[np.ndarray, pd.Series], dist2: Union[np.ndarray, pd.S
...
+    contingency_matrix = pd.DataFrame([dist1_counts, dist2_counts], dtype=int)
 
-    # This is based on
-    # https://stackoverflow.com/questions/46498455/categorical-features-correlation/46498792#46498792 # noqa: SC100
-    # and reused in other sources
-    # (https://towardsdatascience.com/the-search-for-categorical-correlation-a1cf7f1888c9) # noqa: SC100
+    # Based on https://en.wikipedia.org/wiki/Cram%C3%A9r%27s_V# bias correction method # noqa: SC100
```
nirhutnik, line 150: Nope. The code itself was a copy (not just the concept), and this is part of the license terms of stackoverflow, to mention that the code was copied from there

nirhutnik (approved):

deepchecks/utils/distribution/drift.py
Nadav-Barak: The code there has a bug and yours does not.. chi2_contingency there takes a confusion matrix

matanper (approved):

deepchecks/tabular/_shared_docs.py
```diff
@@ -34,13 +34,14 @@
...
-L2 norm of feature importance alone, specifically, ||FI + PER_FEATURE_SCORES|| - ||FI||. This method returns a
-value between 0 and sqrt(n_features).
-'weighted': Weighted mean based on feature importance, provides a robust estimation on how
-much the resulting score will affect the model's performance.
+'l2_weighted': Default. L2 norm over the 'per-feature scores' vector weighted by the feature importance, specifically, 
+sqrt( sum(FI * PER_FEATURE_SCORES^2) ). This method takes into account the feature importance yet puts more weight on
+the per-feature scores. This method is recommended for most cases.
+'l3_weighted': Similar to 'l2_weighted', but with L3 norm. Puts even more emphasis on the per-feature scores and 
+specifically on the largest per-feature scores returning a score closer to maximum among the per-feature scores.
+'weighted': Similar to 'l2_weighted', but with L1 norm. Puts equal weight on feature importance and per-feature scores.
```
noamzbr, line 44: Would explain weighted as it is, rather than "see that more complicated thing? That but simpler"

deepchecks/tabular/dataset.py
```diff
@@ -229,6 +228,14 @@ def __init__(
                                index_name if not set_index_from_dataframe_index else None,
                                datetime_name if not set_datetime_from_dataframe_index else None}]
 
+        if len(set(self._data.index)) != len(self._data.index):
+            if set_index_from_dataframe_index:
+                raise DeepchecksValueError('Selected index column has duplicate values.')
+            else:
+                self._data['original_df_index'] = self._data.index
+                self._data.index = range(len(self._data.index))
+                warnings.warn('Dataframe index has duplicate indexes, using sequenced list instead.')
```
noamzbr, line 237:
```suggestion
                warnings.warn('Dataframe index has duplicate indexes, setting index [0,1..,n-1].')
```

deepchecks/utils/distribution/drift.py
```diff
@@ -131,23 +131,23 @@ def cramers_v(dist1: Union[np.ndarray, pd.Series], dist2: Union[np.ndarray, pd.S
...
     if balance_classes is True:
         dist1_counts, dist2_counts = rebalance_distributions(dist1_counts, dist2_counts)
 
-    contingency_matrix = pd.DataFrame([dist1_counts, dist2_counts])
-
-    # If columns have the same single value in both (causing division by 0), return 0 drift score:
-    if contingency_matrix.shape[1] == 1:
-        return 0
+    # Down sample one of the distributions so that they have the same total number of samples
+    dist1_sum, dist2_sum = sum(dist1_counts), sum(dist2_counts)
```
noamzbr, line 143: Do this only if not balance_classes ?

noamzbr (commented):
Update drift guide docs and check example docs.

deepchecks/utils/distribution/drift.py
Nadav-Barak: We will handle unbalanced after, but gut feeling is it is still required

PR #2360, opened 2023-03-01
changed plotly version to the latest

yromanyshyn:
looks like this issue is not actual anymore with latest plotly version, therefore pinning it to `>=5.13.1` 

@ItayGabbay do we have a FAQ docs page?
I think we need to link this plotly docs page there - [Plotly FAQ - JupyterLab problems](https://plotly.com/python/troubleshooting/#jupyterlab-problems)

### DIFF
--- a/docs/source/user-guide/general/showing_results.rst
+++ b/docs/source/user-guide/general/showing_results.rst
@@ -75,6 +75,8 @@ options to resolve them:
 - | Verifying that the ``jupyter-widgets`` extension is installed and enabled.
   | If you're working on jupyterlab, same for ``jupyterlab-plotly`` extension.
   | Restart your jupter server after these changes so that they will have an effect.
+  | For more information and instructions take a look at a plotly `JupeterLab Troubleshooting<https://plotly.com/python/troubleshooting/#jupyterlab-problems>`_ 
+  | documentation section.
 
 - Exporting the results (e.g. :doc:`saving them to html </user-guide/general/export_save_results>`) 
   and viewing the output file.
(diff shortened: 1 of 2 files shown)


matanper:
Which issue exactly do you refer to?

ItayGabbay (approved):

ItayGabbay:
@yromanyshyn can you add it here?
https://docs.deepchecks.com/stable/user-guide/general/showing_results.html#alternative-methods-for-showing-results-troubleshooting

JKL98ISR:
was it tested on kaggle/google collab?
@yromanyshyn

yromanyshyn:
@ItayGabbay 

> @yromanyshyn can you add it here? https://docs.deepchecks.com/stable/user-guide/general/showing_results.html#alternative-methods-for-showing-results-troubleshooting

added

-----

@JKL98ISR 
> was it tested on kaggle/google collab?
> @yromanyshyn

do not we display results in a iframe when it comes to those platforms?

```python
if is_kaggle_env() or is_databricks_env() or is_sagemaker_env():
    self.show_in_iframe(as_widget=as_widget, unique_id=unique_id, **kwargs)
```

as I remember widgets did not work on kaggle even previously/

JKL98ISR:
@yromanyshyn i think it matters for the iframe too

yromanyshyn:
@JKL98ISR 

> @yromanyshyn i think it matters for the iframe too

iframe element is isolated from the rest of the page and we include all needed js dependencies into the iframe, so, it does not matter where you embed it, it should work everywhere

tested on google collab, works there, but I did not test on Kaggle, it has some problems working in my region/

PR #2417, opened 2023-03-23
Fix nlp properties runtime

nirhutnik:


noamzbr (approved):
Approved, but do clarify the relationship between ignore_properties, include_properties and the new parameter

PR #2432, opened 2023-04-02
refactor text data

Nadav-Barak:
Part of effort to robust the nlp base package

Main changes:
Input validation and convert to uniform standard (np arrays)
No outside usage of index (only used internally to support sampling mechanism + display)
Labels and predictions which are not multilabel are converted to str
For token classification - you must supply tokenized text

### DIFF
--- a/deepchecks/nlp/context.py
+++ b/deepchecks/nlp/context.py
@@ -14,6 +14,7 @@ import typing as t
 from operator import itemgetter
 
 import numpy as np
+
 from deepchecks.core.context import BaseContext
 from deepchecks.core.errors import (DatasetValidationError, DeepchecksNotSupportedError, DeepchecksValueError,
                                     ModelValidationError, ValidationError)
--- a/deepchecks/nlp/context.py
+++ b/deepchecks/nlp/context.py
@@ -14,7 +14,6 @@ import typing as t
 from operator import itemgetter
 
 import numpy as np
-
 from deepchecks.core.context import BaseContext
 from deepchecks.core.errors import (DatasetValidationError, DeepchecksNotSupportedError, DeepchecksValueError,
                                     ModelValidationError, ValidationError)
@@ -90,9 +89,9 @@ class _DummyModel(BasicModel):
 
         if train is not None and test is not None:
             # check if datasets have same indexes
-            if set(train.index) & set(test.index):
-                train.reindex(list(map(lambda x: f'train-{x}', list(train.index))))
-                test.reindex(list(map(lambda x: f'test-{x}', list(test.index))))
+            if set(train.get_original_text_indexes()) & set(test.get_original_text_indexes()):
+                train._original_text_index = np.asarray([f'train-{i}' for i in train.get_original_text_indexes()])
+                test._original_text_index = np.asarray([f'test-{i}' for i in test.get_original_text_indexes()])
                 get_logger().warning('train and test datasets have common index - adding "train"/"test"'
                                      ' prefixes. To avoid that provide datasets with no common indexes '
                                      'or pass the model object instead of the predictions.')
@@ -108,26 +107,29 @@ class _DummyModel(BasicModel):
 
                 if dataset.task_type == TaskType.TEXT_CLASSIFICATION:
                     if (y_pred is None) and (y_proba is not None):
-                        if dataset.is_multilabel:
+                        if dataset.is_multi_label_classification():
                             y_pred = (np.array(y_proba) > 0.5)  # TODO: Replace with user-configurable threshold
                             y_pred = [np.array(model_classes)[pred] for pred in y_pred]
                         else:
                             y_pred = np.argmax(np.array(y_proba), axis=-1)
-                            y_pred = np.array(model_classes)[y_pred]
+                            y_pred = np.array(model_classes, dtype='str')[y_pred]
 
                     if y_pred is not None:
-                        y_pred = np.array(y_pred)
+                        if dataset.is_multi_label_classification():
+                            y_pred = np.array(y_pred)
+                        else:
+                            y_pred = np.array(y_pred, dtype='str')
                         if len(y_pred.shape) > 1 and y_pred.shape[1] == 1:
                             y_pred = y_pred[:, 0]
                         ensure_predictions_shape(y_pred, dataset.text)
 
                     if y_proba is not None:
                         ensure_predictions_proba(y_proba, y_pred)
-                        y_proba_dict = dict(zip(dataset.index, y_proba))
+                        y_proba_dict = dict(zip(dataset.get_original_text_indexes(), y_proba))
                         probas.update({dataset.name: y_proba_dict})
 
                 if y_pred is not None:
-                    y_pred_dict = dict(zip(dataset.index, y_pred))
+                    y_pred_dict = dict(zip(dataset.get_original_text_indexes(), y_pred))
                     predictions.update({dataset.name: y_pred_dict})
 
         self.predictions = predictions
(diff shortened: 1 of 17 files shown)


deepchecks/nlp/input_validations.py
```diff
@@ -0,0 +1,110 @@
...
+from deepchecks.core.errors import DeepchecksValueError
+from deepchecks.nlp.task_type import TaskType, TTextLabel
+from deepchecks.utils.logger import get_logger
+from deepchecks.utils.type_inference import infer_categorical_features
+from deepchecks.utils.validation import is_sequence_not_str
+
+
+def validate_tokenized_text(tokenized_text: Optional[Sequence[Sequence[str]]]):
+    """Validate tokenized text format."""
+    error_string = 'tokenized_text must be a Sequence of sequences of strings'
```
nirhutnik, line 26:
```suggestion
    error_string = 'tokenized_text must be a Sequence of Sequences of strings'
```
Or without capitals in both.

deepchecks/nlp/text_data.py
```diff
@@ -65,210 +62,116 @@ class TextData:
...
+                raise DeepchecksValueError('task_type must be set when label is provided')
             self._task_type = TaskType.OTHER
         elif task_type == 'text_classification':
             self._task_type = TaskType.TEXT_CLASSIFICATION
         elif task_type == 'token_classification':
+            if tokenized_text is None:
+                raise DeepchecksValueError('tokenized_text must be provided for token_classification task type')
+            validate_tokenized_text(tokenized_text)
+            modified = [[str(token) for token in tokens_per_sample] for tokens_per_sample in tokenized_text]
+            self._tokenized_text = np.asarray(modified)
```
nirhutnik, line 109: Why do this modification for every tokenized_text?
Also, you already validate that its a sequence of sequence of strings... no?

deepchecks/nlp/text_data.py
```diff
@@ -65,210 +62,116 @@ class TextData:
...
-            self._validate_tokenized_text(tokenized_text)
-            self._text, self._tokenized_text = list(raw_text), list(tokenized_text)
-            if len(raw_text) != len(tokenized_text):
-                raise DeepchecksValueError('raw_text and tokenized_text must have the same length')
-
-        if index is None:
-            self.index = list(range(len(raw_text)))
-        elif len(index) != len(raw_text):
-            raise DeepchecksValueError('index must be the same length as raw_text')
+            self._text = np.asarray([' '.join(tokens) for tokens in tokenized_text])  # Revisit this decision
```
nirhutnik, line 118: Why did you dump the whole elif else block there?

deepchecks/nlp/text_data.py
```diff
@@ -65,210 +62,116 @@ class TextData:
     task_type : str, default: None
         The task type for the text data. Can be either 'text_classification' or 'token_classification'. Must be set if
         label is provided.
-    dataset_name : t.Optional[str] , default: None
+    name : t.Optional[str] , default: None
         The name of the dataset. If None, the dataset name will be defined when running it within a check.
-    index : t.Optional[t.Sequence[int]] , default: None
-        The index of the samples. If None, the index is set to np.arange(len(raw_text)).
```
nirhutnik, line 71: So you're getting rid of the index?
Why?

deepchecks/nlp/text_data.py
```diff
@@ -392,50 +294,24 @@ def calculate_default_properties(self, include_properties: t.List[str] = None,
...
     def properties_types(self) -> t.Dict[str, str]:
         """Return the property types of the dataset."""
+        if self._properties is None:
+            raise DeepchecksValueError('Properties does not exist, please set it first using the set_properties method')
         return self._properties_types
 
-    def __len__(self):
-        """Return number of samples in the dataset."""
-        return self.n_samples
-
```
nirhutnik, line 438: Why?

deepchecks/nlp/text_data.py
```diff
@@ -392,50 +294,24 @@ def calculate_default_properties(self, include_properties: t.List[str] = None,
...
 
     @property
     def properties(self) -> pd.DataFrame:
         """Return the properties of the dataset."""
         if self._properties is None:
-            raise DeepchecksNotSupportedError(
-                'TextData does not contain properties, add them by using calculate_default_properties or '
-                'set_properties functions')
+            get_logger().info('No properties are set, calculating default properties')
+            self.calculate_default_properties()
```
nirhutnik, line 305: Not wanted behavior, we discussed this in the past. As properties can take a long time to calculate, and they are not saved to the TextData object (as we take a copy of the user's original object and not the original one), then calculating them in the Suite/Checks will result in long running times + will have to be run again anyway.

So we decided that properties should be precalculated...

nirhutnik (commented):
Honestly, not sure what or why you did there with the index, as the previous problems with it were solved... Can you explain why you did it?

deepchecks/nlp/text_data.py
Nadav-Barak: I validate its a sequence, here i change it to numpy array of strings

deepchecks/nlp/text_data.py
Nadav-Barak: Rearranged the order of stuff so not sure what you mean specifically but what i did do, is to add validation that if you selected "token classification" you must provide the tokenized text

deepchecks/nlp/text_data.py
Nadav-Barak: It is a cause of confusion and over complication, if you have index you have to handle cases that some of the input parameter come with different index, index not in correct order between inputs etc .. 
This way as a first step we keep things simple - samples are in-order across all inputs (tokenized text, text, label, properties, etc..)

deepchecks/nlp/text_data.py
Nadav-Barak: its somewhere else

deepchecks/nlp/text_data.py
Nadav-Barak: Ok, reverted to previous

deepchecks/nlp/text_data.py
nirhutnik: yeah... but why?

deepchecks/nlp/text_data.py
nirhutnik: @noamzbr approved of this? Because the index was important to him

deepchecks/nlp/context.py
```diff
@@ -90,9 +90,9 @@ def __init__(self,
 
         if train is not None and test is not None:
             # check if datasets have same indexes
-            if set(train.index) & set(test.index):
-                train.reindex(list(map(lambda x: f'train-{x}', list(train.index))))
-                test.reindex(list(map(lambda x: f'test-{x}', list(test.index))))
+            if set(train.get_original_text_indexes()) & set(test.get_original_text_indexes()):
+                train._original_text_index = np.asarray([f'train-{i}' for i in train.get_original_text_indexes()])
+                test._original_text_index = np.asarray([f'test-{i}' for i in test.get_original_text_indexes()])
```
nirhutnik, line 95: So, if index is now an internal thingy, and is ALWAYS range(n), this will happen all the time... 
We can just change the "name" of the dataset instead or something.

deepchecks/nlp/text_data.py
```diff
@@ -65,210 +62,116 @@ class TextData:
...
-        if index is None:
-            self.index = list(range(len(raw_text)))
-        elif len(index) != len(raw_text):
-            raise DeepchecksValueError('index must be the same length as raw_text')
+            self._text = np.asarray([' '.join(tokens) for tokens in tokenized_text])  # Revisit this decision
         else:
-            self.index = list(index)
+            validate_raw_text(raw_text)
+            self._text = np.asarray([str(x) for x in raw_text])
+            if tokenized_text is not None and len(raw_text) != len(tokenized_text):
```
nirhutnik, line 122: "if tokenized_text is not None" not necessary - you're in the "else" block anyway, it must mean raw_text=None and tokenized_text!=None

deepchecks/nlp/text_data.py
nirhutnik: You changed the if-else block of "what if there's only raw_text / tokenized_text". I do like the changes (I agree it's better to say "give me tokenized text" if the task is token classification).
But, if I understand correctly, I see the following caveat:

If the user supplied tokenized_text but not raw_text, and specified (incorrectly) that the task is text classification, your code breaks (tokenized_text is not None, but also not validated).

Maybe there are more, I hated that if-else block but it was necessary at the time (though I think you covered all other cases)

deepchecks/nlp/context.py
nirhutnik: Also also, there was the "reindex" function... Why change it to accessing a private attr of a class?

deepchecks/nlp/text_data.py
Nadav-Barak: So its consistant and will help avoid problems down the line,
ex: labels[[0,1]] return different thing if it is a list or np array of pandas

deepchecks/nlp/text_data.py
Nadav-Barak: If you selected text classification, we only care about the format of the raw_text + that the label is in the relevant format, which is validated.
Specifically, in your case either you will get an exception for bad prediction/label format or we dont care and all checks will work fine with the raw text

deepchecks/nlp/text_data.py
Nadav-Barak: yes

deepchecks/nlp/context.py
Nadav-Barak: To avoid user accesable functions regarding index

deepchecks/nlp/text_data.py
Nadav-Barak: its raw_text!=None and tokenized text can be both

ItayGabbay (approved):

deepchecks/nlp/text_data.py
nirhutnik: 👍

deepchecks/nlp/context.py
nirhutnik: 👍

deepchecks/nlp/text_data.py
nirhutnik: 👍

deepchecks/nlp/text_data.py
nirhutnik: Yeah, sorry, misread that

nirhutnik (approved):

PR #2449, opened 2023-04-16
nlp doc structure + metrics_guide.rst

Nadav-Barak:


docs/source/user-guide/general/metrics_guide.rst
```diff
@@ -309,3 +347,8 @@ ______________
     :language: python
     :lines: 34-64
     :tab-width: 0
+
+NLP Example
+___________
+
+Currently unavailable, will be added in future releases.
```
nirhutnik, line 354: Why add it, though? Shouldn't we remove this and add an issue? (I guess unavailable because you want to show token classification?)

docs/source/user-guide/general/metrics_guide.rst
```diff
@@ -22,13 +22,13 @@ Controlling the metrics helps you shape the checks and suites according to the s
...
-:doc:`SingleDatasetPerformance </checks_gallery/vision/model_evaluation/plot_single_dataset_performance>`
+:doc:`TrainTestPerformance </checks_gallery/tabular/model_evaluation/plot_train_test_performance>`
 come with default metrics.
 
 The default metrics by task type are:
 
-Tabular
-_______
+Classification
+______________
```
nirhutnik, line 31: Now that there are several sections, maybe add a table in the beginning of all content (and to which subpackages it's relevant?)

docs/source/user-guide/general/metrics_guide.rst
```diff
@@ -48,7 +48,29 @@ Multiclass classification per class:
...
+*   Precision ``'token_precision_per_class'``
+*   Recall ``'token_recall_per_class'``
+
+Object Detection (Vision only)
+______________________________
+
+*   Mean average precision ``'average_precision_per_class'``
+*   Mean average recall  ``'average_recall_per_class'``
+
+Regression
```
nirhutnik, line 72:
```suggestion
Regression (Tabular only)
```

nirhutnik (approved):

docs/source/user-guide/general/metrics_guide.rst
Nadav-Barak: For sake of clarity I think its better with this section for users to know it will be added in the future, but dont have strong feeling either way

docs/source/user-guide/general/metrics_guide.rst
shir22: Why limit? (e.g. regression for CV)

shir22 (approved):

docs/source/user-guide/general/metrics_guide.rst
nirhutnik: We don't currently support regression in CV, as far as I know

Nadav-Barak: leaving it as is

PR #2453, opened 2023-04-16
performance docs NLP

Nadav-Barak:
[Metadata Segments Performance — gallery-boilerplate documentation.pdf](https://github.com/deepchecks/deepchecks/files/11242460/Metadata.Segments.Performance.gallery-boilerplate.documentation.pdf)
[Property Segments Performance — gallery-boilerplate documentation.pdf](https://github.com/deepchecks/deepchecks/files/11242461/Property.Segments.Performance.gallery-boilerplate.documentation.pdf)
[Single Dataset Performance — gallery-boilerplate documentation.pdf](https://github.com/deepchecks/deepchecks/files/11242462/Single.Dataset.Performance.gallery-boilerplate.documentation.pdf)

### DIFF
--- a/docs/source/checks/nlp/model_evaluation/plot_confusion_matrix_report.py
+++ b/docs/source/checks/nlp/model_evaluation/plot_confusion_matrix_report.py
@@ -4,7 +4,7 @@
 
 Confusion Matrix Report
 ***********************
-This notebook provides an overview for using and understanding the Confusion Matrix Report check.
+This notebook provides an overview for using and understanding the Confusion Matrix Report check for NLP tasks.
 
 
 **Structure:**
@@ -25,31 +25,23 @@ such as accuracy, precision, recall etc. (`confusion matrix <https://en.wikipedi
 #%%
 # Generate data & model
 # =======================
-import pandas as pd
-from sklearn.datasets import load_iris
-from sklearn.ensemble import AdaBoostClassifier
-from sklearn.model_selection import train_test_split
+from deepchecks.nlp import TextData
+from deepchecks.nlp.checks import ConfusionMatrixReport
+from deepchecks.nlp.datasets.classification.tweet_emotion import load_data, load_precalculated_predictions
 
-from deepchecks.tabular import Dataset
-from deepchecks.tabular.checks import ConfusionMatrixReport
+tweets_data = load_data(data_format='DataFrame', as_train_test=False)
+tweets_dataset = TextData(tweets_data.text, label=tweets_data['label'],
+                          task_type='text_classification')
 
-iris = load_iris(as_frame=True)
-clf = AdaBoostClassifier()
-frame = iris.frame
-X = iris.data
-y = iris.target
-X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.5, random_state=42)
-clf.fit(X_train, y_train)
-ds = Dataset(pd.concat([X_test, y_test], axis=1), 
-            features=iris.feature_names,
-            label='target')
+predictions = load_precalculated_predictions(as_train_test=False)
+
 
 #%%
 # Run the check
 # ===============
 
 check = ConfusionMatrixReport()
-result = check.run(ds, clf)
+result = check.run(tweets_dataset, predictions=predictions)
 result.show()
 
 #%%
--- a/docs/source/checks/nlp/model_evaluation/plot_metadata_segments_performance.py
+++ b/docs/source/checks/nlp/model_evaluation/plot_metadata_segments_performance.py
@@ -1,9 +1,9 @@
 # -*- coding: utf-8 -*-
 """
-Weak Segments Performance
+Metadata Segments Performance
 *************************
 
-This notebook provides an overview for using and understanding the weak segment performance check.
+This notebook provides an overview for using and understanding the metadata segment performance check.
 
 **Structure:**
 
@@ -16,8 +16,9 @@ This notebook provides an overview for using and understanding the weak segment 
 What is the purpose of the check?
 ==================================
 
-The check is designed to help you easily identify the model's weakest segments in the data provided. In addition,
-it enables to provide a sublist of the Dataset's features, thus limiting the check to search in
+The check is designed to help you easily identify the model's weakest segments based on the provided
+:func:`metadata <deepchecks.nlp.text_data.TextData.set_metadata>`. In addition,
+it enables to provide a sublist of the metadata columns, thus limiting the check to search in
 interesting subspaces.
 
 Automatically detecting weak segments
@@ -41,11 +42,12 @@ The check contains several steps:
 # Generate data & model
 # =====================
 
-from deepchecks.tabular.datasets.classification.phishing import (
-    load_data, load_fitted_model)
+from deepchecks.nlp.datasets.classification.tweet_emotion import load_data, load_precalculated_predictions
 
-_, test_ds = load_data()
-model = load_fitted_model()
+_, test_dataset = load_data(data_format='TextData')
+_, test_probas = load_precalculated_predictions(pred_format='probabilities')
+
+test_dataset.metadata.head(3)
 
 #%%
 # Run the check
@@ -55,11 +57,10 @@ model = load_fitted_model()
 # check and especially its output.
 #
 # ``columns / ignore_columns``: Controls which columns should be searched for weak segments. By default,
-# a heuristic is used to determine which columns to use based solely on their feature importance.
+# uses all columns.
 #
 # ``alternative_scorer``: Determines the metric to be used as the performance measurement of the model on different
 # segments. It is important to select a metric that is relevant to the data domain and task you are performing.
-# By default, the check uses Neg RMSE for regression tasks and Accuracy for classification tasks.
 #
 # ``segment_minimum_size_ratio``: Determines the minimum size of segments that are of interest. The check is tuned
 # to find the weakest segment regardless of the segment size and so it is recommended to try different configurations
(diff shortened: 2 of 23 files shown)


docs/source/checks/nlp/model_evaluation/plot_metadata_segments_performance.py
```diff
@@ -0,0 +1,107 @@
...
+The check contains several steps:
+
+#. We calculate loss for each sample in the dataset using the provided model via either log-loss or MSE according
+   to the task type.
+
+#. Select a subset of features for the weak segment search. This is done by selecting the features with the
+   highest feature importance to the model provided (within the features selected for check, if limited).
+
+#. We train multiple simple tree based models, each one is trained using exactly two
+   features (out of the ones selected above) to predict the per sample error calculated before.
```
shir22, line 36: How many features are selected?
How do we choose which feature combination(s) to present / include in the results?

docs/source/checks/nlp/model_evaluation/plot_metadata_segments_performance.py
```diff
@@ -0,0 +1,107 @@
...
+
+test_dataset.metadata.head(3)
+
+#%%
+# Run the check
+# =============
+#
+# The check has several key parameters (that are all optional) that affect the behavior of the
+# check and especially its output.
+#
```
shir22, line 58: Transforming following to bullets would be nicer

docs/source/checks/nlp/model_evaluation/plot_metadata_segments_performance.py
```diff
@@ -0,0 +1,107 @@
...
+# ``columns / ignore_columns``: Controls which columns should be searched for weak segments. By default,
+# uses all columns.
+#
+# ``alternative_scorer``: Determines the metric to be used as the performance measurement of the model on different
```
shir22, line 62: add link that enables to understand how to work with these scorers? or for more details in api reference (if link is there?)

docs/source/checks/nlp/model_evaluation/plot_metadata_segments_performance.py
```diff
@@ -0,0 +1,107 @@
...
+# segments. It is important to select a metric that is relevant to the data domain and task you are performing.
+#
+# ``segment_minimum_size_ratio``: Determines the minimum size of segments that are of interest. The check is tuned
+# to find the weakest segment regardless of the segment size and so it is recommended to try different configurations
+# of this parameter as larger segments can be of interest even the model performance on them is superior.
```
shir22, line 67: Improve clarity?
(What I found unclear: first sentence isn't clear if someone isn't 100% on what we call a "segment" and what do we mean "of interest". and in second part it seems that the default of this is 0, because it's written "regardless of". Would try to make it clearer how this parameter actually affects the displayed results)

docs/source/checks/nlp/model_evaluation/plot_metadata_segments_performance.py
```diff
@@ -0,0 +1,107 @@
...
+* `Automatically detecting weak segments <#automatically-detecting-weak-segments>`__
+* `Generate data & model <#generate-data-model>`__
+* `Run the check <#run-the-check>`__
+* `Define a condition <#define-a-condition>`__
+
+What is the purpose of the check?
+==================================
+
+The check is designed to help you easily identify the model's weakest segments based on the provided
+:func:`metadata <deepchecks.nlp.text_data.TextData.set_metadata>`. In addition,
```
shir22, line 20: does the set_metadata have clear explanation about what is metadata?

and/or - add link to somewhere or explain shortly?

docs/source/checks/nlp/model_evaluation/plot_property_segments_performance.py
```diff
@@ -0,0 +1,107 @@
+# -*- coding: utf-8 -*-
+"""
+Property Segments Performance
+*************************
+
+This notebook provides an overview for using and understanding the metadata segment performance check.
```
shir22, line 6: metadata -> property

docs/source/checks/nlp/model_evaluation/plot_property_segments_performance.py
```diff
@@ -0,0 +1,107 @@
...
+* `Automatically detecting weak segments <#automatically-detecting-weak-segments>`__
+* `Generate data & model <#generate-data-model>`__
+* `Run the check <#run-the-check>`__
+* `Define a condition <#define-a-condition>`__
+
+What is the purpose of the check?
+==================================
+
+The check is designed to help you easily identify the model's weakest segments based on the provided
+:func:`properties <deepchecks.nlp.text_data.TextData.set_properties>`. In addition,
```
shir22, line 20: same here - are properties explained there?

docs/source/checks/nlp/model_evaluation/plot_metadata_segments_performance.py
```diff
@@ -0,0 +1,107 @@
...
+# the result.value attribute.
+
+
+result.value['weak_segments_list'].head(3)
+
+#%%
+# Define a condition
+# ==================
+#
+# We can define on our check a condition that will validate that the model performance on the weakest segment detected
```
shir22, line 99: We can add a condition that will validate the model's performance on the ...

shir22: I think also worth explaining in a sentence why this is a wanted condition (why this is straightforward, I'm not sure it will be clear when reading the current sentence, for someone that isn't familiar with this check)

docs/source/checks/nlp/model_evaluation/plot_metadata_segments_performance.py
```diff
@@ -0,0 +1,107 @@
...
+#. We calculate loss for each sample in the dataset using the provided model via either log-loss or MSE according
+   to the task type.
+
+#. Select a subset of features for the weak segment search. This is done by selecting the features with the
+   highest feature importance to the model provided (within the features selected for check, if limited).
+
+#. We train multiple simple tree based models, each one is trained using exactly two
+   features (out of the ones selected above) to predict the per sample error calculated before.
+
+#. We convert each of the leafs in each of the trees into a segment and calculate the segment's performance. For the
```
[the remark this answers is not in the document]
shir22: "convert a leaf to a segment" -> explain what that means?

docs/source/checks/nlp/model_evaluation/plot_metadata_segments_performance.py
```diff
@@ -0,0 +1,107 @@
...
+
+What is the purpose of the check?
+==================================
+
+The check is designed to help you easily identify the model's weakest segments based on the provided
+:func:`metadata <deepchecks.nlp.text_data.TextData.set_metadata>`. In addition,
+it enables to provide a sublist of the metadata columns, thus limiting the check to search in
+interesting subspaces.
+
+Automatically detecting weak segments
```
shir22, line 24: I wonder if this explanation (and possibly also the "key parameters")
should be in some general concepts and then linked from all weak-segment checks.

I assume that you copy-pasted this from other segments checks, and if we take into account tabular, vision, nlp - there should be 5 "weak segment"  checks

(and btw maybe same for domain classifier if its more than twice in docs but not related directly here)

shir22 (approved):

docs/source/checks/nlp/model_evaluation/plot_metadata_segments_performance.py
```diff
@@ -0,0 +1,107 @@
...
+result.show()
+
+#%%
+# Observe the check's output
+# --------------------------
+#
+# We see in the results that the check indeed found several segments on which the model performance is below average.
+# In the heatmap display we can see model performance on the weakest segments and their environment with respect to the
+# two features that are relevant to the segment. In order to get the full list of weak segments found we will inspect
+# the result.value attribute.
```
shir22, line 90:
```suggestion
# the result.value attribute.
```

shir22, line 90: ``result.value``

docs/source/checks/nlp/model_evaluation/plot_metadata_segments_performance.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+result.value['weak_segments_list'].head(3)
```
shir22, line 93: 1. Would explicitly mention that in the display we show only "top 3" "lowest performers" but actually the list may be longer and they can be observed etc.... (you did mention that just maybe emphasize that in the display we see only top 3). Accordingly wouldn't limit here to top 3 (but other number) cause otherwise one can assume that we don't have more if they don't read.

Specifically - not necessarily the "top segements" are most interesting since we only do a "go-no go" with segment ratio. So other ones may be more interesting if the have a larger percentage of data.
BTW in your example it's actually like that, see the last weak segment:
230418 at 00 29 36 screenshot Acrobat
Would be worth mentioning it in "understanding the results". Assuming we use constant seeds along the way (in example here + in building the trees) so that it will reproduce...

Sidenote - a bit confusing that the output dataframe indices are not ordered. (see screenshot)

2. What isn't clear is - how many do we have? what is the threshold by which we decide if there is a weak segment or not? (corresponding to how many weak segment will we have..? or will we always have?)

docs/source/checks/nlp/model_evaluation/plot_confusion_matrix_report.py
```diff
@@ -0,0 +1,47 @@
+# -*- coding: utf-8 -*-
```
shir22, line 1: Please check also output (you added one of a the Single Dataset Performance and not of the Confusion matrix)

docs/source/checks/nlp/model_evaluation/plot_metadata_segments_performance.py
```diff
@@ -0,0 +1,107 @@
+# -*- coding: utf-8 -*-
+"""
```
shir22, line 2: Can you open issues for all of the warnings so we fix the checks such that they'll work nicely without warnings for current and previous scikit-learn versions?
Seems that there are a few different ones.

docs/source/checks/nlp/model_evaluation/plot_metadata_segments_performance.py
Nadav-Barak: Currently no, later we will have a doc

docs/source/checks/nlp/model_evaluation/plot_metadata_segments_performance.py
Nadav-Barak: maybe, but not now :)

PR #2456, opened 2023-04-18
[DEE-457] "PropertyDrift" and "TrainTestPerformance" checks

yromanyshyn:


deepchecks/utils/abstracts/train_test_performace.py
```diff
@@ -0,0 +1,85 @@
...
+__all__ = ['TrainTestPerformanceAbstract']
+
+
+class TrainTestPerformanceAbstract:
+    """Base functionality for some train-test performance checks."""
+
+    @classmethod
+    @abc.abstractmethod
+    def _default_per_class_scorers(cls) -> t.Mapping[str, str]:
+        raise NotImplementedError()
```
Nadav-Barak, line 31: Why is this needed? 
In general the more prefered flow is to get the scorers list from context.get_scorers(self.scorers, use_avg_defaults=False) - as it was in the tabular sub package

deepchecks/nlp/checks/model_evaluation/train_test_performance.py
```diff
@@ -0,0 +1,212 @@
...
+        self.n_samples = n_samples
+        self.random_state = random_state
+
+    @classmethod
+    def _default_per_class_scorers(cls) -> t.Dict[str, str]:
+        # TODO:
+        return {}
+
+    def run_logic(self, context: Context) -> CheckResult:
+        """Run check."""
```
Nadav-Barak, line 100: Try to move more of this section into the Abstract - similar to what happening in PredictionDriftAbstract - a function the return CheckResult
Optimal - send scorers and the dataset and do everything there
Minimal - create the display in the abstract (send the scorers results)

deepchecks/nlp/checks/train_test_validation/property_drift.py
```diff
@@ -0,0 +1,270 @@
...
+        n_top_properties: int = 5,
+        margin_quantile_filter: float = 0.025,
+        max_num_categories_for_drift: t.Optional[int] = None,
+        min_category_size_ratio: float = 0.01,
+        max_num_categories_for_display: int = 10,
+        show_categories_by: str = 'largest_difference',
+        numerical_drift_method: str = 'KS',
+        categorical_drift_method: str = 'cramers_v',
+        ignore_na: bool = True,
+        aggregation_method: t.Optional[str] = 'l3_weighted',
```
Nadav-Barak, line 114: remove the aggregation_method argument

deepchecks/nlp/checks/train_test_validation/property_drift.py
```diff
@@ -0,0 +1,270 @@
...
+        min_samples: int = 10,
```
Nadav-Barak, line 115: change min samples to 100

deepchecks/nlp/checks/train_test_validation/property_drift.py
```diff
@@ -0,0 +1,270 @@
...
+        self.categorical_drift_method = categorical_drift_method
+        self.ignore_na = ignore_na
+        self.aggregation_method = aggregation_method
+        self.min_samples = min_samples
+        self.n_samples = n_samples
+        self.random_state = random_state
+
+    def run_logic(self, context: Context) -> CheckResult:
+        """Run check."""
+        train = t.cast(TextData, context.train)
```
Nadav-Barak, line 139: why is the cast needed? can you do in the context property call?

deepchecks/nlp/checks/train_test_validation/property_drift.py
```diff
@@ -0,0 +1,270 @@
...
+            columns=self.properties,
+            ignore_columns=self.ignore_properties
+        )
+        test_properties = select_from_dataframe(
+            test.properties,
+            columns=self.properties,
+            ignore_columns=self.ignore_properties
+        )
+
+        columns = set(train_properties.columns).intersection(set(test_properties.columns))
```
Nadav-Barak, line 153: add a function called in creation of context that validate if the train & test have the same properties + properties type and if not display a warning stating only common properties will be used in train-test checks.

Nadav-Barak: best if can be a generic structure like in validate_length_and_calculate_column_types so in the future we will also validate metadata and embeddings in the same manner

deepchecks/nlp/checks/train_test_validation/property_drift.py
```diff
@@ -0,0 +1,270 @@
...
+        cat_columns = set([*train.categorical_properties, *test.categorical_properties])
```
Nadav-Barak, line 154: only use columns that are common to both train and test properties

deepchecks/nlp/checks/train_test_validation/property_drift.py
```diff
@@ -0,0 +1,270 @@
...
+        results = {}
```
Nadav-Barak, line 155: All code below this, can they be put under a abstract class? Similar to what you did in the TrainTestPerformance

I mean here a common abstract class that will be the parent of this class and of FeatureDrift tabular check

tests/nlp/checks/train_test_validation/property_drift_test.py
```diff
@@ -0,0 +1,166 @@
...
+            'Max Word Length': has_entries({'Drift score': 0.18000000000000005, 'Method': 'Kolmogorov-Smirnov'}),
+            'Average Word Length': has_entries({'Drift score': 0.1, 'Method': 'Kolmogorov-Smirnov'}),
+            '% Special Characters': has_entries({'Drift score': 0.16000000000000003, 'Method': 'Kolmogorov-Smirnov'}),
+            'Text Length': has_entries({'Drift score': 0.30000000000000004, 'Method': 'Kolmogorov-Smirnov'}),
+            'Subjectivity': has_entries({'Drift score': 0.14, 'Method': 'Kolmogorov-Smirnov'}),
+            'Sentiment': has_entries({'Drift score': 0.08000000000000007, 'Method': 'Kolmogorov-Smirnov'})
+        }))  # type: ignore
+
+
+class TestLabelClassification:
```
Nadav-Barak, line 132: multilabel

tests/nlp/checks/train_test_validation/property_drift_test.py
```diff
@@ -0,0 +1,166 @@
...
+            raw_text=[
+                random.choice(['I think therefore I am', 'I am therefore I think', 'I am'])
+                for _ in range(20)
+            ],
+            label=[
+                random.choice([[0, 0, 1], [1, 1, 0], [0, 1, 0]])
+                for _ in range(20)
+            ],
+            task_type='text_classification'
+        )
```
Nadav-Barak, line 145: Put this one at context as 'dummy multilabel dataset'

deepchecks/nlp/checks/train_test_validation/property_drift.py
yromanyshyn: fixed

deepchecks/utils/abstracts/train_test_performace.py
yromanyshyn: because `add_condition_class_performance_imbalance_ratio_less_than` needs access to the default scorers

```
def add_condition_class_performance_imbalance_ratio_less_than(
    self: Self,
    threshold: float = 0.3,
    score: t.Optional[str] = None
) -> Self:
    if score is None:
        score = next(iter(self._default_per_class_scorers()))

```

deepchecks/nlp/checks/train_test_validation/property_drift.py
yromanyshyn: to say to the static type analyzer that the `context.train` is of type `TextData`

currently, the static type analyzer infers those properties to be of type `Never`, which means that code cannot and will not be reached, it is because `nlp.context.Context` type inherences those properties from `BaseContext` in which they are defined as `None`

```
class BaseContext(ABC):
    """Base class for contexts."""

    _train = None
    _test = None
    
    ...

    @property
    def train(self):
        """Return train if exists, otherwise raise error."""
        if self._train is None:
            raise DeepchecksNotSupportedError('Check is irrelevant for Datasets without train dataset')
        return self._train

```

Never


> can you do in the context property call?

do you suggest overriding `train` and `test` properties in the nlp `Context` type?

yromanyshyn: >can you do in the context property call?

or we can make `BaseContext` generic?

tests/nlp/checks/train_test_validation/property_drift_test.py
yromanyshyn: fixed

deepchecks/nlp/checks/train_test_validation/property_drift.py
yromanyshyn: fixed, now use only common columns that have the same type, other columns are igored

deepchecks/nlp/checks/train_test_validation/property_drift.py
yromanyshyn: added

tests/nlp/checks/train_test_validation/property_drift_test.py
yromanyshyn: fixed

deepchecks/nlp/checks/train_test_validation/property_drift.py
Nadav-Barak: I think the solution is too complicated so we can just leave it as is.
I thought we can do something simple like adding return type declaration to the property

deepchecks/nlp/checks/model_evaluation/train_test_performance.py
Nadav-Barak: Yea ok

deepchecks/utils/abstracts/train_test_performace.py
Nadav-Barak: wow that is some bad implementation.
Change the condition so it will be as in deepchecks/vision/checks/model_evaluation/class_performance.py
i.e the scorer name is received as input from the user

yromanyshyn: oh, I see why it was done so, to not change the function signature, to not break backward compatibility

anyway, I do not see this as a problematic change/

deepchecks/nlp/context.py
```diff
@@ -326,14 +329,45 @@ def __init__(
...
             if test_dataset.name is None:
                 test_dataset.name = 'Test'
+
         # If both dataset, validate they fit each other
         if train_dataset and test_dataset:
-            if test_dataset.has_label() and train_dataset.has_label() and not \
-                    train_dataset.validate_textdata_compatibility(test_dataset):
+            if (
+                test_dataset.has_label()
+                and train_dataset.has_label()
```
Nadav-Barak, line 337: why is this needed? validate_textdata_compatibility should be able to work regardless is labels are present or not

deepchecks/nlp/context.py
```diff
@@ -326,14 +329,45 @@ def __init__(
...
+                and not train_dataset.validate_textdata_compatibility(test_dataset)
+            ):
                 raise DatasetValidationError('train_dataset and test_dataset must share the same label and task type')
+
+            if (
+                train_dataset._properties is not None
+                and train_dataset._cat_properties is not None
+                and test_dataset._properties is not None
+                and test_dataset._cat_properties is not None
+            ):
```
Nadav-Barak, line 347: The if is redundant, the compare method should be able to menage situations where some of those parameters are none and the rest are not
e.x only test_dataset._cat_properties is none

deepchecks/utils/abstracts/feature_drift.py
```diff
@@ -0,0 +1,185 @@
...
+            sorted_by = 'drift score'
+            key = lambda col: results[col]['Drift score'] or 0
+            columns_order = sorted(results.keys(), key=key, reverse=True)[:self.n_top_columns]
+
+        headnote = [
+            textwrap.dedent(
+                f"""
+                <span>
+                The Drift score is a measure for the difference between two distributions, in this check - the test
+                and train distributions.<br> The check shows the drift score and distributions for the features, sorted
```
Nadav-Barak, line 128: replace features with a argument received externally stating wether it is feature or property - 
same as data_table_name argument in validate_length_and_calculate_column_types

Nadav-Barak (approved):

deepchecks/utils/abstracts/feature_drift.py
```diff
@@ -0,0 +1,185 @@
...
+                by {sorted_by} and showing only the top {self.n_top_columns} features, according to {sorted_by}.
+                </span>
+                """
+            ),
+            get_drift_plot_sidenote(
+                self.max_num_categories_for_display,
+                self.show_categories_by
+            ),
+            'If available, the plot titles also show the feature importance (FI) rank'
```
Nadav-Barak, line 137: remove this text for properties

yromanyshyn: fixed

deepchecks/utils/abstracts/feature_drift.py
yromanyshyn: fixed

deepchecks/nlp/context.py
yromanyshyn: > the compare method should be able to menage situations where some of those parameters are none and the rest are not

hm/ I disagree, IMO, you are suggesting leaking `Context` specific logic into the `compare_dataframes` function, and that is not good

a task of the  `compare_dataframes` function is two compare two data frames, it should not care about anything more, making some of its parameters optional just will overcomplicate this function and makes it more confusing to understand

> e.x only test_dataset._cat_properties is none

sorry, do not understand/
`_properties` also can be none

its definition
```
_properties: t.Optional[pd.DataFrame] = None
```

deepchecks/nlp/context.py
yromanyshyn: hm, I do not know/ I did not add this, I just moved each part of the condition exp on the new line

deepchecks/nlp/context.py
Nadav-Barak: Main requirement - 
if one text data has 5 properties  with 1 of them being categorical and the other has 4 properties with 0 categorical (meaning _cat_properties is None) we still want the functionalities to work

Nadav-Barak: Specifically IMO the logic to handle this case should be inside the compare_dataframes function

deepchecks/nlp/context.py
Nadav-Barak: ok yea i see

Nadav-Barak (approved):

deepchecks/nlp/context.py
yromanyshyn: >if one text data has 5 properties with 1 of them being categorical and the other has 4 properties with 0 categorical (meaning _cat_properties is None) we still want the functionalities to work

oh, yeap, did not think about that, thanks!

>Specifically IMO the logic to handle this case should be inside the compare_dataframes function

moving that `if` statement into the `compare_dataframe` function just means that it will appear just a little bit later in the code

```
# NOW >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>

if (
    train_dataset._properties is not None
    and test_dataset._properties is not None
):
    result = compare_dataframes(
        train=train_dataset._properties,
        test=test_dataset._properties,
        train_categorical_columns=train_dataset._cat_properties,
        test_categorical_columns=test_dataset._cat_properties
    )
    self._common_datasets_properties = result.common
    if result.difference is not None:
        get_logger().warning(
            'Train and Test datasets have different properties, '

# SUGGESTION >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>

# returns none if one of the dataframes is none
result = compare_dataframes(
    train=train_dataset._properties,
    test=test_dataset._properties,
    train_categorical_columns=train_dataset._cat_properties,
    test_categorical_columns=test_dataset._cat_properties
)
if result is not None:
    self._common_datasets_properties = result.common
    if result.difference is not None:
        get_logger().warning(
            'Train and Test datasets have different properties, '
```

PR #2458, opened 2023-04-20
[DEE-486] nlp 'token_*_per_class' scorers fix

yromanyshyn:
also see [DEE-480](https://linear.app/deepchecks/issue/DEE-480#comment-ea1d869e)

### DIFF
--- a/deepchecks/nlp/metric_utils/token_classification.py
+++ b/deepchecks/nlp/metric_utils/token_classification.py
@@ -51,16 +51,16 @@ def get_scorer_dict(
     }
 
     return {
-        'token_accuracy': make_token_scorer(accuracy_score, **common_kwargs),
-        'token_f1_per_class': make_token_scorer(f1_score, **common_kwargs, average=None),
-        'token_f1_macro': make_token_scorer(f1_score, **common_kwargs, average='macro'),
-        'token_f1_micro': make_token_scorer(f1_score, **common_kwargs, average='micro'),
-        'token_precision_per_class': make_token_scorer(precision_score, **common_kwargs, average=None),
-        'token_precision_macro': make_token_scorer(precision_score, **common_kwargs, average='macro'),
-        'token_precision_micro': make_token_scorer(precision_score, **common_kwargs, average='micro'),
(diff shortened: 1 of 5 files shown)


deepchecks/nlp/metric_utils/token_classification.py
```diff
@@ -21,12 +21,7 @@
 __all__ = ['get_default_token_scorers', 'validate_scorers', 'get_scorer_dict']
 
 DEFAULT_AVG_SCORER_NAMES = ('f1_macro', 'recall_macro', 'precision_macro')
-DEFAULT_PER_CLASS_SCORER_NAMES = tuple()
-
-# see issue DEE-473
-# https://linear.app/deepchecks/issue/DEE-473/incorrectly-inferred-model-classes-for-token-classification-task
-#
-# DEFAULT_PER_CLASS_SCORER_NAMES = ('f1_per_class',)
+DEFAULT_PER_CLASS_SCORER_NAMES = ('f1_per_class', 'precision_per_class', 'recall_per_class')
```
Nadav-Barak, line 24: shouldn't they be token_f1_per_class and such? @noamzbr

noamzbr: Yep think Nadav is correct, original value was wrong also ):

noamzbr: Mmm I see @yromanyshyn added `return [f'token_{it}' for it in names]` in line 105 which makes it work. I know it was me that decided to call these scorers "token_*" but I'm not sure why it's necessary. @yromanyshyn any reason why not to just call them with a regular name?

yromanyshyn: @noamzbr 

>@yromanyshyn any reason why not to just call them with a regular name?

hm, no do not see any reason why should they start with `token`
I will rename

yromanyshyn: renamed

Nadav-Barak (approved):

ItayGabbay (approved):

PR #2468, opened 2023-04-23
Added properties and metadata guides

nirhutnik:
Added metadata and properties guides for nlp


deepchecks/nlp/text_data.py
```diff
@@ -66,12 +66,27 @@ class TextData:
...
+        order of the rows must be the same as the order of the samples in the dataset.
+        For more on metadata, see the `NLP Metadata Guide
+        <https://docs.deepchecks.com/en/latest/nlp/nlp-metadata.html>`_.
+    categorical_metadata : t.Optional[t.List[str]] , default: None
+        The names of the categorical metadata columns. If None, categorical metadata columns are automatically inferred.
+        Only relevant if metadata is not None.
+    properties : t.Optional[pd.DataFrame] , default: None
+        The text properties for the samples. Properties must be given as a pandas DataFrame, with the rows representing
+        each sample and columns representing the different properties. If None, no properties are set.
+        The number of rows in the metadata DataFrame must be equal to the number of samples in the dataset, and the
```
noamzbr, line 81:
```suggestion
        The number of rows in the properties DataFrame must be equal to the number of samples in the dataset, and the
```

deepchecks/nlp/text_data.py
```diff
@@ -66,12 +66,27 @@ class TextData:
...
+        order of the rows must be the same as the order of the samples in the dataset.
+        For more on metadata, see the `NLP Metadata Guide
+        <https://docs.deepchecks.com/en/latest/nlp/nlp-metadata.html>`_.
+    categorical_metadata : t.Optional[t.List[str]] , default: None
+        The names of the categorical metadata columns. If None, categorical metadata columns are automatically inferred.
+        Only relevant if metadata is not None.
+    properties : t.Optional[pd.DataFrame] , default: None
+        The text properties for the samples. Properties must be given as a pandas DataFrame, with the rows representing
+        each sample and columns representing the different properties. If None, no properties are set.
```
noamzbr, line 80: What about the 'auto' option?

docs/source/user-guide/nlp/nlp_metadata.rst
```diff
@@ -0,0 +1,80 @@
+.. _nlp_metadata_guide:
+
+=================
+NLP Metadata
+=================
+
+Metadata is any additional information about the texts, that can be used to better understand the data and the model's
+behavior. As opposed to the unstructured text, metadata is structured, and is expected to be a pandas DataFrame.
```
noamzbr, line 8:
```suggestion
behavior. In contrast to unstructured text, metadata is structured, and is expected to be a pandas DataFrame.
```

docs/source/user-guide/nlp/nlp_metadata.rst
```diff
@@ -0,0 +1,80 @@
...
+Metadata can be anything - from the source of the data, the timestamp of its creation, to the age of the author, and
+any other information that is given about the text.
+
+What Is Metadata Used For?
+=============================
+
+Metadata is used by Deepchecks' checks to extract meaningful features from the data, since some computations cannot be
+computed directly on the text (for example, drift). Inspecting the distribution of a metadata column's values (e.g. to
```
noamzbr, line 16: cannot be computed directly on the text - not so clear what is the relation between that fact and the fact we ask for metadata. 
Would perhaps give weak segments as the first example here, as it's clearer and gives a better sense of "why should I care"

docs/source/user-guide/nlp/nlp_metadata.rst
```diff
@@ -0,0 +1,80 @@
...
+   function.
+
+In both options, it's advised to also use the ``categorical_metadata`` parameter to specify which metadata columns are
+categorical. The parameter should be a list of the names of the categorical columns.
+
+
+In the following example, we will pass the metadata to the ``TextData`` object in order to use the
+``MetadataSegmentsPerformance`` check:
+
+.. code-block:: python
```
noamzbr, line 58: Consider showing a printout of a metadata dataframe

docs/source/user-guide/nlp/nlp_properties.rst
```diff
@@ -0,0 +1,167 @@
+.. _nlp_properties_guide:
+
+=================
+NLP Properties
+=================
+
+Properties are one-dimension values that are extracted from either the text, labels or predictions. For example, a
```
noamzbr, line 7: Right now we don't token properties, and addressing it complicates this guide significantly. I would advise for now just talking about text properties.

noamzbr: Also you can't really configure them even when we will implement, so maybe no need to add them here.

docs/source/user-guide/nlp/nlp_properties.rst
```diff
@@ -0,0 +1,167 @@
...
+Language                        Language of the text. Uses the langdetect library
+Sentiment                       Sentiment of the text. Uses the textblob library
+Subjectivity                    Subjectivity of the text. Uses the textblob library
+Toxicity*                       Toxicity of the text. Uses the unitary/toxic-bert model
+Fluency*                        Fluency of the text. Uses the prithivida/parrot_fluency_model model
+Formality*                      Formality of the text. Uses the s-nlp/roberta-base-formality-ranker model
+==============================  ==========
+
+*These properties are not calculated by default, as they may take a long time to calculate. To use them, pass
+``include_long_calculation_properties`` to the :class:`TextData.calculate_properties <deepchecks.nlp.TextData>` method.
```
noamzbr, line 65:
```suggestion
``include_long_calculation_properties=True`` to the :class:`TextData.calculate_properties <deepchecks.nlp.TextData>` method.
```

docs/source/user-guide/nlp/nlp_properties.rst
```diff
@@ -0,0 +1,167 @@
...
+==========================
+
+Whether you are using the built-in properties or your own, the process of using them in the checks is the same.
+In order to use the properties of your text in a check, the properties should already be part of the ``TextData`` object.
+
+
+Calculating The Built-in Properties
+-----------------------------------
+
+In order to use the built-in properties, you must call the ``calculate_default_properties`` method of the ``TextData``
```
noamzbr, line 85: What about the 'auto' option?

docs/source/user-guide/nlp/nlp_properties.rst
```diff
@@ -0,0 +1,167 @@
...
+
+
+Moreover, some properties are not calculated by default, as they may take a long time to calculate. In order to
+use them, pass ``include_long_calculation_properties`` to the ``calculate_default_properties`` method.
+
+In the following example, we will calculate the properties and include only the long calculation property "Toxicity":
+
+.. code-block:: python
+
+  text_data.calculate_default_properties(include_long_calculation_properties=True, include_properties=['Toxicity'])
```
noamzbr, line 127: Will this override the existing properties or just add?

docs/source/user-guide/nlp/nlp_properties.rst
```diff
@@ -0,0 +1,167 @@
...
+
+
+Using Your Own Properties
+-------------------------
+
+If you already have the properties calculated, there are 2 ways to use them:
```
noamzbr, line 133: Missing the why - either you saved the deepchecks properties for this dataset somewhere to save time, or you calculated something smart of your own.

noamzbr (commented):
Meta comment - They guides ignore each other, which I suspect will be confusing as it's already not clear what's the difference between properties and metadata. I think both guides should start with some disambugiuation.

docs/source/user-guide/nlp/nlp_metadata.rst
nirhutnik: This is an rst file.

docs/source/user-guide/nlp/nlp_properties.rst
nirhutnik: Not sure how will be done, but we are planning to do this before the release...

docs/source/user-guide/nlp/nlp_properties.rst
nirhutnik: deja vu

docs/source/user-guide/nlp/nlp_metadata.rst
noamzbr: Can add a table to rst

deepchecks/nlp/text_data.py
```diff
@@ -275,6 +294,9 @@ def set_metadata(
         if self._metadata is not None:
             warnings.warn('Metadata already exist, overwriting it', UserWarning)
 
+        if isinstance(metadata, str):
+            metadata = pd.read_csv(metadata)
```
noamzbr, line 298: Can we add a safeguard here, such that if the first column has the title 0 and has the entries 1..n then we raise warning that it's an index columns saved by mistake and we're ignoring it?

docs/source/user-guide/nlp/nlp_metadata.rst
nirhutnik: Seems weird and unnecessary IMO... Is it important to you?

deepchecks/nlp/text_data.py
nirhutnik: Honestly, I don't think so. I think a user should be responsible for that themselves and they'll see their error.
Also, I don't really think anyone should use this, but added this for symmetry's sake as I added for properties.

docs/source/user-guide/nlp/nlp_metadata.rst
nirhutnik: yeah, it remained from properties accidentally. I changed the sentence a bit, tell me WDYT

docs/source/user-guide/nlp/nlp_properties.rst
nirhutnik: That's a good point. Will change

docs/source/user-guide/nlp/nlp_metadata.rst
noamzbr: don't invest in it more than 5 minutes, but I bet you gpt4 can do that in 1.

deepchecks/nlp/text_data.py
noamzbr: yeah I actually meant that for loading properties but can let this go.

docs/source/user-guide/nlp/nlp_metadata.rst
```diff
@@ -0,0 +1,86 @@
...
+
+Not to be confused with :ref:`text properties <nlp_properties_guide>`, which are features extracted from the text
+itself, such as text length or sentiment.
+
+What Is Metadata Used For?
+=============================
+
+Metadata is used by Deepchecks' checks to examine the dataset, without directly using the text, as some tests are
+difficult to perform on unstructured data.
+Inspecting the distribution of a metadata column's values (e.g. to notice that the source of the texts is different
```
noamzbr, line 20: Still think the weak segment performance example is better as a first sentence because we don't do metadata drift, but can let this go.

noamzbr (approved):

deepchecks/nlp/text_data.py
nirhutnik: So for loading properties, if the user would originally use our load_properties function, then there won't be an issue to begin with :)

docs/source/user-guide/nlp/nlp_metadata.rst
nirhutnik: I think it's more complicated to explain in a short parenthesis, that's why I kept this example.

docs/source/user-guide/nlp/nlp_metadata.rst
nirhutnik: My problem is not that it would take time, it's that I don't think it would look good.

docs/source/user-guide/nlp/nlp_metadata.rst
```diff
@@ -0,0 +1,86 @@
+.. _nlp_metadata_guide:
```
Nadav-Barak, line 1:
```suggestion
.. _nlp__metadata_guide:
```

deepchecks/vision/suites/default_suites.py
```diff
@@ -48,7 +48,7 @@ def train_test_validation(label_properties: List[Dict[str, Any]] = None, image_p
              - :class:`~deepchecks.vision.checks.train_test_validation.ImagePropertyDrift`
            * - :ref:`plot_vision_image_dataset_drift`
              - :class:`~deepchecks.vision.checks.train_test_validation.ImageDatasetDrift`
-           * - :ref:`plot_vision_feature_label_correlation_change`
+           * - :ref:`_nlp__property_label_correlation `
```
shir22, line 51: ?

PR #2477, opened 2023-04-28
[DEE-506] added 'redownload_models' param to the 'calculate_default_properties' function

yromanyshyn:


Nadav-Barak (changes requested):
I would implement it differently, 
instead of the redownload_models parameter add a parameter controlling where the downloaded model should be saved (with some default) and for each model first look for the model file at the provided path - and if not there only then download and save there

yromanyshyn:
@Nadav-Barak I redone it

deepchecks/nlp/utils/text_properties.py
```diff
@@ -263,6 +330,8 @@ def calculate_default_properties(
         ignored, even if they are in the include_properties parameter.
     device : int, default None
         The device to use for the calculation. If None, the default device will be used.
+    models_storage : Union[str, pathlib.Path, None], default None
+        a directory to store the models.
```
Nadav-Barak, line 334: Add desc on where it is saved when None

deepchecks/nlp/utils/text_properties.py
```diff
@@ -55,16 +55,32 @@ def get_transformer_model(
...
+        if not models_storage.is_dir():
+            raise ValueError('"model_storage" expected to be a directory')
+
     if device not in (None, 'cpu'):
         transformers = _import_optional_property_dependency('transformers', property_name=property_name)
         # TODO: quantize if 'quantize_model' is True
-        return transformers.AutoModelForSequenceClassification.from_pretrained(model_name)
+        return transformers.AutoModelForSequenceClassification.from_pretrained(
+            model_name,
+            cache_dir=models_storage
```
Nadav-Barak, line 82: maybe also change the argument name in our function to cache_dir as well

deepchecks/nlp/utils/text_properties.py
Nadav-Barak: Add info that if the folder already contains the relevant resources they are not re-downloaded (+ verify it happens :) )

Nadav-Barak (approved):
Two additional stuff:
1. Can you also try and provide a similar behaviour to the nltk.download functions?
2.  add a test for this functionality

deepchecks/nlp/utils/text_properties.py
yromanyshyn: added

deepchecks/nlp/utils/text_properties.py
yromanyshyn: hm/ do not know, I think `models_storage` represents exactly the purpose of the parameter whereas `cache_dir` is more ambiguous (like, what exactly will be cached)

yromanyshyn:
@Nadav-Barak 

>1. Can you also try and provide a similar behaviour to the nltk.download functions?
>2. add a test for this functionality

**point 2.** added
**point 1.** How exactly, should it look?/ `nltk.download` function has `download_dir` what else is needed?

deepchecks/nlp/utils/text_properties.py
```diff
@@ -25,7 +25,7 @@
 __all__ = ['calculate_default_properties']
 
 
-ONNX_MODELS_STORAGE = pathlib.Path(__file__).absolute().parent / '.onnx-nlp-models'
+MODELS_STORAGE = pathlib.Path(__file__).absolute().parent / '.nlp-models'
```
noamzbr, line 28:
```suggestion
MODELS_STORAGE = pathlib.Path(__file__).absolute().parent / '.deepchecks-nlp-models'
```
Would suggest so it's clear to the user

noamzbr:
> @Nadav-Barak
> 
> > 1. Can you also try and provide a similar behaviour to the nltk.download functions?
> > 2. add a test for this functionality
> 
> **point 2.** added **point 1.** How exactly, should it look?/ `nltk.download` function has `download_dir` what else is needed?

I think you can test that the file is actually created in the dir specified

deepchecks/nlp/utils/text_properties.py
yromanyshyn: IMO, it is not needed because it will be within `DEEPCHECKS_LIB_PATH/nlp/utils/.nlp-models`

image

PR #2483, opened 2023-05-01
Noam/dee 521 unknown token check

noamzbr:
Implements the unknown tokens check.

Display:
image

The less common words / words with 1 appearne go into the "other":
image

TODO:
Write more tests
Write plot example?

### DIFF
--- a/deepchecks/nlp/checks/data_integrity/unknown_tokens.py
+++ b/deepchecks/nlp/checks/data_integrity/unknown_tokens.py
@@ -76,56 +76,75 @@ class UnknownTokens(SingleDatasetCheck):
         if len(samples) == 0:
             raise DeepchecksValueError('Dataset cannot be empty')
 
-        all_unknown_words_counter, total_words = self.find_unknown_words(samples)
+        indices = dataset.get_original_text_indexes()
+
+        all_unknown_words_counter, total_words, unknown_word_indexes = self.find_unknown_words(samples, indices)
         if len(all_unknown_words_counter) == 0:
             display = None
+            value = None
         else:
             fig = self.create_pie_chart(all_unknown_words_counter, total_words)
             display = [fig]
 
-        return CheckResult(None, display=display)
+            # The value contains two fields - unknown_word_percent and unknown_word_details.
+            # The latter contains a dict, in which for each word we have its ratio of the data and the list of indexes
+            # of the samples that contain it.
+            unknown_word_details = {}
+            for word, indexes in unknown_word_indexes.items():
+                unknown_word_details[word] = {'ratio': all_unknown_words_counter[word] / total_words,
+                                              'indexes': indexes}
+            value = {'unknown_word_ratio': sum(all_unknown_words_counter.values()) / total_words,
+                     'unknown_word_details': unknown_word_details}
 
-    def find_unknown_words(self, samples):
+        return CheckResult(value, display=display)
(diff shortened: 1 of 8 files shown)


deepchecks/nlp/checks/data_integrity/unknown_tokens.py
```diff
@@ -0,0 +1,175 @@
...
+__all__ = ['UnknownTokens']
+
+
+@docstrings
+class UnknownTokens(SingleDatasetCheck):
+    """Find samples that contain tokens unsupported by your tokenizer.
+
+    Parameters
+    ----------
+    tokenizer: t.Any , default: None
```
Nadav-Barak, line 36: this needs more details

deepchecks/nlp/checks/data_integrity/unknown_tokens.py
```diff
@@ -0,0 +1,175 @@
...
+            value = {'unknown_word_ratio': sum(all_unknown_words_counter.values()) / total_words,
+                     'unknown_word_details': unknown_word_details}
+
+        return CheckResult(value, display=display)
+
+    def find_unknown_words(self, samples, indices):
+        """Find words with unknown tokens in samples."""
+        # Choose tokenizer based on availability of nltk
+        if nltk.download('punkt'):
+            tokenize = nltk.word_tokenize
```
Nadav-Barak, line 103: This seem a bit weird to me.
If you want to validate the users tokenizer shouldn't we use the same splitting mechanism he uses? we will need to ask for it with explanation

deepchecks/nlp/checks/data_integrity/unknown_tokens.py
```diff
@@ -0,0 +1,175 @@
...
+        # Tokenize samples and count unknown words
+        words_array = [tokenize(sample) for sample in samples]
+        all_unknown_words = []
+        unknown_word_indexes = {}
+        total_words = 0
+        for idx, words in zip(indices, words_array):
+            total_words += len(words)
+            for word in words:
+                tokens = self.tokenizer.tokenize(word)
+                if any(self.tokenizer.convert_tokens_to_ids(token) == self.tokenizer.unk_token_id for token in tokens):
```
Nadav-Barak, line 116: does all tokinizers have the convert_tokens_to_ids function?

deepchecks/nlp/checks/data_integrity/unknown_tokens.py
noamzbr: This is just to split to words. Then we apply the tokenizer, because we want to be able to pinpoint which word was the one to have a problematic tokens. 
We build on the fact that tokenziers are mostly sub-word.
This is done because huggingface tokenizers DONT have the option to know where each token was w relation to the original text.

deepchecks/nlp/checks/data_integrity/unknown_tokens.py
noamzbr: as far as I could see

Nadav-Barak: add this as a requirement for user provided tokenizers + verify it in init

Nadav-Barak (approved):
Need to add somewhere in the display description on what the percent represent

ItayGabbay (approved):

PR #2485, opened 2023-05-01
simplify weak segments

Nadav-Barak:
Mainly - no need to use Dataset and DummyModel in order to run core logic

### DIFF
--- a/deepchecks/nlp/checks/model_evaluation/weak_segments_performance.py
+++ b/deepchecks/nlp/checks/model_evaluation/weak_segments_performance.py
@@ -9,8 +9,7 @@
 # ----------------------------------------------------------------------------
 #
 """Module of weak segments performance check."""
-import warnings
-from typing import Callable, Dict, List, Union
+from typing import Callable, Dict, List, Optional, Union
 
 import numpy as np
 import pandas as pd
@@ -20,10 +19,9 @@ from deepchecks.core import CheckResult
 from deepchecks.core.check_result import DisplayMap
 from deepchecks.core.errors import DeepchecksNotSupportedError, DeepchecksProcessError
 from deepchecks.nlp import Context, SingleDatasetCheck
-from deepchecks.tabular import Dataset
+from deepchecks.nlp.utils.weak_segments import get_relevant_data_table
 from deepchecks.tabular.context import _DummyModel
 from deepchecks.utils.abstracts.weak_segment_abstract import WeakSegmentAbstract
-from deepchecks.utils.dataframes import select_from_dataframe
 from deepchecks.utils.typing import Hashable
 
 __all__ = ['MetadataSegmentsPerformance', 'PropertySegmentsPerformance']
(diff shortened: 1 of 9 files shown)


deepchecks/utils/abstracts/weak_segment_abstract.py
```diff
@@ -96,8 +103,12 @@ def _create_heatmap_display(self, dummy_model, encoded_dataset, weak_segments, a
                         scores[f2_idx, f1_idx] = np.NaN
                         counts[f2_idx, f1_idx] = 0
                     else:
-                        scores[f2_idx, f1_idx] = scorer.run_on_data_and_label(dummy_model, segment_data,
-                                                                              segment_data[encoded_dataset.label_name])
+                        if scorer is not None and dummy_model is not None and label_col is not None:
+                            scores[f2_idx, f1_idx] = scorer.run_on_data_and_label(dummy_model,
```
noamzbr, line 107: simplify further and make a equivalent function to run_on_data_and_label that accepts label and prediction rather than dummy model? Is there the issue with the indexes here that means we must have the dummy model?

deepchecks/utils/performance/partition.py
```diff
@@ -46,11 +46,18 @@ def __init__(self, filter_functions: List[Callable] = None, label: str = ''):
             self.filter_functions = filter_functions
         self.label = label
 
-    def filter(self, dataframe: pd.DataFrame) -> pd.DataFrame:
+    def filter(self, dataframe: pd.DataFrame, label_col: Optional[pd.Series] = None) -> \
```
noamzbr, line 49: In what cases we need to pass the label here?

deepchecks/vision/checks/model_evaluation/weak_segments_performance.py
```diff
@@ -140,52 +135,35 @@ def update(self, context: Context, batch: BatchWrapper, dataset_kind: DatasetKin
...
+        loss_per_sample_col = 'score'
+        results_dict[loss_per_sample_col] = self._sample_scores
         results_df = pd.DataFrame(results_dict)
         properties_used = self.image_properties or default_image_properties
         cat_features = [p['name'] for p in properties_used if p['output_type'] == 'categorical']
-        num_features = [p['name'] for p in properties_used if p['output_type'] == 'numerical']
-        all_features = cat_features + num_features
 
-        dataset = Dataset(results_df, cat_features=cat_features, features=all_features, label='score')
+        #  encoding categorical features based on the loss per sample (not smart but a way to gets the job done)
```
noamzbr, line 144: Why not encode on the label like before?

deepchecks/nlp/checks/model_evaluation/weak_segments_performance.py
```diff
@@ -158,7 +125,7 @@ class PropertySegmentsPerformance(WeakSegmentsAbstractText):
         Properties to check, if none are given checks all properties except ignored ones.
     ignore_properties : Union[Hashable, List[Hashable]] , default: None
         Properties to ignore, if none given checks based on properties variable
-    n_top_properties : int , default: 10
+    n_top_properties : Optional[int] , default: 10
```
noamzbr, line 128: I don't think that you need to write this in the docstring, default: 10 means it's Optional already

deepchecks/nlp/context.py
```diff
@@ -471,23 +471,6 @@ def assert_task_type(self, *expected_types: TaskType):
...
-            raise DeepchecksNotSupportedError(
-                'Check requires metadata, but the the TextData object had none. To use this check, use the '
-                'set_metadata method to set your own metadata with a pandas.DataFrame.')
-
-    @staticmethod
-    def assert_properties(text_data):
-        """Assert that properties exists."""
-        if text_data.properties is None:
-            raise DeepchecksNotSupportedError(
-                'Check requires properties, but the the TextData object had none. To use this check, use the'
```
noamzbr, line 487: We're not using this in any of the property-requiring checks? Why not?

noamzbr (commented):
deepchecks/utils/abstracts/weak_segment_abstract.py is programatically simpler but human-wise still hard to follow.. Can we add some comments?

noamzbr (approved):

PR #2489, opened 2023-05-02
Nb/feat/under annotated segments

Nadav-Barak:


deepchecks/nlp/checks/data_integrity/under_annotated_segments.py
```diff
@@ -0,0 +1,197 @@
...
+        text_data = text_data.sample(self.n_samples, random_state=context.random_state)
+
+        features, cat_features = get_relevant_data_table(text_data, data_type=self.segment_by,
+                                                         columns=self.columns, ignore_columns=self.ignore_columns,
+                                                         n_top_features=self.n_top_features)
+
+        encoded_dataset = self._target_encode_categorical_features_fill_na(features, text_data.label,
+                                                                           cat_features)
+
+        loss_per_sample = pd.Series([1 - pd.isna(x) for x in text_data.label], index=encoded_dataset.data.index)
```
noamzbr, line 60: not that it matters really, but I would say the weak segment is the one with nan, so the loss should be pd.isna(x) and not 1-pd.isna(x)

deepchecks/nlp/checks/data_integrity/under_annotated_segments.py
```diff
@@ -0,0 +1,197 @@
...
+                                                   scorer_name='Annotation Ratio')
+
+        if len(weak_segments) == 0:
+            raise DeepchecksProcessError('Check was unable to find under annotated segments. This is expected if '
+                                         'your data is well annotated. If this is not the case, try increasing '
+                                         f'n_samples or supply more {self.segment_by}.')
+
+        avg_score = round(loss_per_sample.mean(), 3)
+
+        weak_segments = self._generate_check_value_display(weak_segments, cat_features)
```
noamzbr, line 72: P0 here (prior to the scatter plot) can be the existing plot with a different color scheme.

deepchecks/nlp/checks/data_integrity/under_annotated_segments.py
Nadav-Barak: Score is mean of loss per sample, and its much easier if we follow convention that higher score is better
+ that way, the score is actually the annotation ratio in the segment

deepchecks/nlp/checks/data_integrity/under_annotated_segments.py
Nadav-Barak: mm good idea, lets talk

deepchecks/utils/abstracts/weak_segment_abstract.py
```diff
@@ -56,14 +59,18 @@ def _target_encode_categorical_features_fill_na(self, data: pd.DataFrame, label_
         # Target encoding of categorical features based on label col (when unavailable we use loss_per_sample)
         if len(cat_features) > 0:
             t_encoder = TargetEncoder(cols=cat_features)
-            label_as_int = pd.Categorical(data[label_name], categories=sorted(data[label_name].unique())).codes
+            if is_numeric_dtype(label_col):
```
noamzbr, line 62: If task is classification but label is int? then we want to treat it as cat.

deepchecks/utils/dataframes.py
```diff
@@ -20,22 +20,28 @@
...
 
 
+def default_fill_na_series(col: pd.Series, is_categorical: t.Optional[bool] = None) -> pd.Series:
+    """Fill NaN values based on column type."""
+    if is_categorical:
+        return col.astype('object').fillna('None')
+    elif is_numeric_dtype(col):
+        return col.astype('float64').fillna(col.mean())
+    else:
+        return col.fillna(col.mode(), inplace=True)
```
noamzbr, line 42: When do we reach this branch of the if?

tests/nlp/checks/data_integrity/under_annotated_segments_test.py
```diff
@@ -0,0 +1,78 @@
...
+
+def test_tweet_emotion_metadata_fully_annotated(tweet_emotion_train_test_textdata):
+    # Arrange
+    _, test = tweet_emotion_train_test_textdata
+    check = UnderAnnotatedMetaDataSegments().add_condition_segments_relative_performance_greater_than()
+
+    # Act & Assert
+    assert_that(
+        calling(check.run).with_args(test),
+        raises(DeepchecksProcessError, 'Check was unable to find under annotated segments. This is expected if '
```
noamzbr, line 75: Maybe this shouldn't be an error for this check, but rather just no display...

PR #2498, opened 2023-05-05
Recsys base

RayanAAY-ops:
I am submitting a pull request that modifies the existing recommender system metric code to improve its performance, efficiency, and robustness. Specifically, I have focused on optimizing the most commonly used functions to make the computation more efficient and reliable.
To ensure the code is working as expected, I have conducted rigorous testing and validation using the make-validate command, which yielded a score of 9.95/10 on pylint.

Overall, these changes are designed to make the recommender system metrics more effective and user-friendly, while also improving its overall performance ( enhanced performance compared to state of the art libraries rexmex/scikit-learn).

 Please note that while I have focused on optimizing the most commonly used functions, I am open to adding additional functions in the future as needed.

#### Any other comments?

What left:
- Include comprehensive unit tests to ensure all functionalities work as expected.

Thank you for your time and consideration.
RayanAAY
---


deepchecks/recommender/ranking.py
```diff
@@ -7,705 +7,423 @@
...
-    hits = 0.0
-    for i, item in enumerate(recommendation):
-        if item in relevant_items and item not in recommendation[:i]:
-            hits += 1.0
-
-    if len(recommendation) == 0:
+    if relevant_item is None or len(recommendation) == 0:
         return 0.0
-    return hits / len(recommendation)
+    for rank, rec_item in enumerate(recommendation):
```
noamzbr, line 175: So in this metric we assume relevant_items (per a specific session) is only one item, while in the previous ones there could be multiple relevant items per session?

deepchecks/recommender/ranking.py
```diff
@@ -7,705 +7,423 @@
...
-    Computes the cosine distance between two 1D arrays
-    NOTE: fastdist before v1.1.5 returned cosine similarity (which is 1 - distance)
-    :params:
-    u, v   : input arrays, both of shape (n,)
-    w      : weights at each index of u and v. array of shape (n,)
-             if no w is set, it is initialized as an array of ones
-             such that it will have no impact on the output
-    :returns:
-    cosine  : float, the cosine distance between u and v
+def recall_k(relevant_items: Union[List[int], np.array, None],
```
noamzbr, line 20: regarding type - do we assume that items must be ints, or can be any hashable?

deepchecks/recommender/ranking.py
```diff
@@ -7,705 +7,423 @@
...
 
-def kendall_tau(relevant_items: np.array, recommendation: np.array):
-    """
-    Calculate the Kendall's tau, measuring the correspondence between two lists.
+    return np.mean(reciprocal_ranks)
 
-    Args:
-        relevant_items (array-like): An 1 x N array of items.
-        recommendation (array-like):  An 1 x N array of items.
+def dcg(y_true : np.ndarray,
```
noamzbr, line 198: We must have name consistency for the current metric calculation method to work. If previously we called y_true relevant_items and y_pred recommendations, we should keep these names in all metrics.

deepchecks/recommender/ranking.py
```diff
@@ -7,705 +7,423 @@
...
-        recommendation (List): A 1 x N array of items, where M is the number
-                               of predicted lists and N the number of recommended items
-        item_popularity (dict): A dict mapping each item in the recommendations to a popularity value.
-                                  Popular items have higher values.
-        num_users (int): The number of users
-        k (int): The number of items considered in each recommendation.
+    Calculate the popularity-based novelty metric for a list of recommendations.
+    Parameters:
+        recommendations (Union[List[list], np.array]): List of recommendations for each user.
+        items_popularity (Union[dict, None]): Item popularity scores.
```
noamzbr, line 280: Should add a few more words on what popularity is, as in the previous implementation.

deepchecks/recommender/ranking.py
```diff
@@ -7,705 +7,423 @@
...
-    if item_popularity is None:
-        raise ValueError("item_popularity must be provided.")
-
-    epsilon = 1e-10
+        Tuple[float, float]: The popularity-based novelty score and the normalized score.
+    """
+    #epsilon = 1e-10
+    # Calculate the mean surprisal/self-information of the top K popular head items
+    top_information_mean = 0
+    most_popular_item_catalog = round(0.05 * len(items_popularity))
```
noamzbr, line 290: Shouldn't this cutoff (5% most popular items) be a parameter?

deepchecks/recommender/ranking.py
```diff
@@ -7,705 +7,423 @@
...
 
-    return avg_self_information
+    # Calculate novelty and normalized novelty
+    novelty = np.mean(all_self_information)
+    normalized_novelty = novelty / top_information_mean
+    return novelty, normalized_novelty
 
-
-def normalized_distance_based_performance_measure(relevant_items: List, recommendation: List):
+def diversity_score(predictions: List, target_vectors: Union[List, np.array],
```
noamzbr, line 310: same comment about naming of the arguments here

deepchecks/recommender/ranking.py
```diff
@@ -7,705 +7,423 @@
...
-    Calculates the Normalized Distance-based Performance Measure (NPDM) between two
-    ordered lists. Two matching orderings return 0.0 while two unmatched orderings returns 1.0.
-
-    Args:
-        relevant_items (List): List of items
-        recommendation (List): The predicted list of items
+    Compute the diversity score of a set of recommendations.
+    Parameters:
+    predictions (List): List of predicted item indices for each user.
+    target_vectors (Union[List, np.array]): List or array of item feature vectors.
```
noamzbr, line 316: maybe call this item_features in that case?

deepchecks/recommender/ranking.py
```diff
@@ -7,705 +7,423 @@
...
 
-        C_u += C**2
+    # Return the diversity score.
+    return 1 - sim_avg
 
-        if C < 0:
-            C_minus += 1
-        else:
-            C_plus += 1
+def mca(x_categorical: Union[np.array, List[List]],
```
noamzbr, line 335: What is this used for?

deepchecks/recommender/ranking.py
```diff
@@ -1,6 +1,6 @@
 # ----------------------------------------------------------------------------
 # Copyright (C) 2021-2023 Deepchecks (https://www.deepchecks.com)
-#
+# #
```
noamzbr, line 3: Why add this here?

noamzbr (commented):
After addressing these comments, could you please open this PR as a PR into recsys-base rather than main?

deepchecks/recommender/ranking.py
```diff
@@ -7,705 +7,423 @@
...
-# https://github.com/AstraZeneca/rexmex
-## =================================================================================================
+    if relevant_items is None or len(relevant_items) == 0:
+        return 0.0
+    if recommendation is None or len(recommendation) == 0:
+        return 0.0
+    relevant_set = set(relevant_items)
+    rec_set = set(recommendation[:k])
+    hits = list((i in relevant_set) for i in recommendation if i in rec_set)
+    score = np.cumsum(hits) / (1 + np.arange(len(hits)))
```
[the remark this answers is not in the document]
RayanAAY-ops: For recall, using the **set()** function helps to remove any duplicates, allowing us to skip the step of checking whether a recommendation has already been made and speed up computation. However, it is crucial to address and raise awareness of any duplicates that may arise during the recommendation process. This should be considered mandatory and included in the Sanity Check  of the recommender to ensure its reliability.

In general, when using a hybrid model that employs heuristic logic (such as incorporating both popularity and recency), it is recommended to apply the **set()** function to every recommendation to eliminate any potential duplicate candidates.

It's also important to note, that we shouldn't use **set()** for the argument :<recommandations> of precision@K because we care about the order. I already corrected this in my actual fork.

deepchecks/recommender/ranking.py
```diff
@@ -7,705 +7,423 @@
...
-        : The mean rank of the relevant items in a predicted.
+    float: The Precision@k score for the set of recommendations.
     """
-    return np.mean([rank(item, recommendation) for item in relevant_items])
+    if relevant_items is None or len(relevant_items) == 0:
+        return 0.0
+    relevant_set = set(relevant_items)
+    rec_set = set(recommendation[:k])
 
+    hits = sum(1 for i in rec_set if i in relevant_set)
```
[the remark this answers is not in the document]
RayanAAY-ops: Fixed for the second point.
However, for the first point, I don't agree, neither the community / theory. When computing the recall, we do not care about the rank of the predicted item, in other words we must not weight the score.
I'm preparing a file with an exemple for each metric, and comparing this with rexmex/scikit-learn.

deepchecks/recommender/ranking.py
RayanAAY-ops: Reciprocal rank is a way to evaluate how good a recommendation is, but it only looks at the first relevant item in the recommendation list. So, if you're looking for an item on Python Programming and the first book recommended is exactly what you're looking for, reciprocal rank would give a high score. But if the second or third book is also good, reciprocal rank wouldn't take that into account.

This is why there are other evaluation metrics, like DCG and NDCG, that take into account all the relevant items in the recommendation list. These metrics  are the way to go because it gives a more complete picture of how well a recommender system is performing, especially if there are multiple relevant items in the list.

Reciprocal rank is useful when the user is looking for something extremely specific, such as when using search engine.

deepchecks/recommender/ranking.py
RayanAAY-ops: You're right, the arguments should support a List of hashable, especially for INT / STRING.

If the items in the list are categorical in nature, such as movie titles or product names, it may be more natural and easier to use strings to represent them --> making the code more readable and intuitive, since it is easier to interpret the meaning of a string than an integer. 

On the other hand, strings involve more processing and memory allocation. For example, comparing strings requires comparing each character in the string, while comparing integers can be done more efficiently using machine-level operations.

In summary, we can either add more flexiblity by supporting Hashable type in general, or focus on pure performance optimization.

deepchecks/recommender/ranking.py
RayanAAY-ops: Fixed.

deepchecks/recommender/ranking.py
RayanAAY-ops: fixed

deepchecks/recommender/ranking.py
RayanAAY-ops: fixed

deepchecks/recommender/ranking.py
RayanAAY-ops: The MCA (Multiple Correspondence Analysis) is a method used to convert categorical features into numerical features. It uses the chi-square statistic, commonly employed for testing independence between categorical variables in a contingency table (using relative frequencies, and kind of conditional probability).

Alternatively, we could use the "overlap" measure which is awidely used to compare data with categorical features. It is popular due to its simplicity. The measure assigns a value of 1 if two rows have the same category in a specific column, and 0 otherwise.

However, combining the diversity matrix for numerical features and the overlap matrix for categorical features can pose challenges. To avoid this issue, using MCA to convert categorical features into numerical features is a recommended approach. MCA is simple, robust, and eliminates the need to merge different matrices.

deepchecks/recommender/ranking.py
RayanAAY-ops: Good point. Fixed.

deepchecks/recommender/ranking.py
noamzbr: Would prefer the flexible option

deepchecks/recommender/ranking.py
noamzbr: seems legit

PR #2511, opened 2023-05-10
[DEE-456] nlp dummy model refactoring

yromanyshyn:
@Nadav-Barak 

> Currently we have a dummy model for tabular and another one for nlp but they basically do the same stuff.
We would like them to have common abstract containing most of the logic with child class implement datadype specific functionalities.

they do not have much in common, in general, `_predict` and `_predict_proba` methods could be separated into some kind of base class, but in this case, we would need to unify `nlp.TextData` and `tabular.Dataset` APIs, everything else is just a validation logic that is pretty specific IMO. So, I do not see a need in creating some kind of ambiguous and not clear abstractions just to reduce a few lines of code

In general, I just refactored the nlp dummy model and moved the validation logic into the `input_validation.py` module

### DIFF
--- a/deepchecks/nlp/context.py
+++ b/deepchecks/nlp/context.py
@@ -9,7 +9,6 @@
 # ----------------------------------------------------------------------------
 #
 """Module for base nlp context."""
-import collections
 import typing as t
 from operator import itemgetter
 
@@ -17,21 +16,21 @@ import numpy as np
 
 from deepchecks.core.context import BaseContext
 from deepchecks.core.errors import (DatasetValidationError, DeepchecksNotSupportedError, DeepchecksValueError,
-                                    ModelValidationError, ValidationError)
-from deepchecks.nlp.input_validations import compare_dataframes
+                                    ModelValidationError)
+from deepchecks.nlp.input_validations import (_validate_multilabel, _validate_text_classification,
+                                              _validate_token_classification, compare_dataframes)
 from deepchecks.nlp.metric_utils.scorers import init_validate_scorers
 from deepchecks.nlp.metric_utils.token_classification import (get_default_token_scorers, get_scorer_dict,
                                                               validate_scorers)
 from deepchecks.nlp.task_type import TaskType
 from deepchecks.nlp.text_data import TextData
 from deepchecks.nlp.utils.data_inference import infer_observed_and_model_labels
-from deepchecks.nlp.input_validations import _validate_multilabel, _validate_text_classification, _validate_token_classification
 from deepchecks.tabular.metric_utils import DeepcheckScorer, get_default_scorers
 from deepchecks.tabular.utils.task_type import TaskType as TabularTaskType
-from deepchecks.tabular.utils.validation import ensure_predictions_proba, ensure_predictions_shape
 from deepchecks.utils.docref import doclink
 from deepchecks.utils.logger import get_logger
 from deepchecks.utils.typing import BasicModel
+from deepchecks.utils.validation import is_sequence_not_str
 
 __all__ = [
     'Context',
@@ -40,13 +39,14 @@ __all__ = [
     'TTokenPred'
 ]
 
-from deepchecks.utils.validation import is_sequence_not_str
 
 TClassPred = t.Union[t.Sequence[t.Union[str, int]], t.Sequence[t.Sequence[t.Union[str, int]]]]
 TClassProba = t.Sequence[t.Sequence[float]]
+# TODO: is it correct, why tuple?
 TTokenPred = t.Sequence[t.Sequence[t.Tuple[str, int, int, float]]]
 TTextPred = t.Union[TClassPred, TTokenPred]
-TTextProba = t.Union[TClassProba]  # TODO: incorrect, why union have only one type argument?
+# TODO: incorrect, why union have only one type argument?
+TTextProba = t.Union[TClassProba]
 
 
 class _DummyModel(BasicModel):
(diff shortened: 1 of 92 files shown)


deepchecks/nlp/context.py
```diff
@@ -39,13 +39,14 @@
     'TTokenPred'
 ]
 
-from deepchecks.utils.validation import is_sequence_not_str
 
 TClassPred = t.Union[t.Sequence[t.Union[str, int]], t.Sequence[t.Sequence[t.Union[str, int]]]]
 TClassProba = t.Sequence[t.Sequence[float]]
+# TODO: is it correct, why tuple?
 TTokenPred = t.Sequence[t.Sequence[t.Tuple[str, int, int, float]]]
```
noamzbr, line 46: It's old and not correct any more, should be 
```suggestion
TTokenPred = t.Sequence[t.Sequence[t.Union[str, int]]]
```

deepchecks/nlp/context.py
```diff
@@ -72,68 +73,85 @@ class _DummyModel(BasicModel):
...
-                 y_pred_test: TTextPred,
-                 y_proba_test: TTextProba,
-                 train: t.Union[TextData, None] = None,
-                 y_pred_train: TTextPred = None,
-                 y_proba_train: TTextProba = None,
-                 model_classes: list = None,
-                 validate_data_on_predict: bool = True):
+    def __init__(
+        self,
+        *,
```
noamzbr, line 78: why enable any args?

deepchecks/nlp/context.py
```diff
@@ -72,68 +73,85 @@ class _DummyModel(BasicModel):
...
+                    dataset=dataset,
+                    predictions=y_pred,
+                    probabilities=y_proba,
+                    n_of_classes=len(model_classes)
+                )
+                if y_pred is not None:
+                    y_pred = np.array(y_pred, dtype='float')
+                elif y_proba is not None:
+                    y_pred = (np.array(y_proba) > multilabel_proba_threshold)
+                    y_pred = [np.array(model_classes)[pred] for pred in y_pred]
```
noamzbr, line 124: No that line should be deleted for multilabel, the y_pred should be a binary matrix

deepchecks/nlp/context.py
```diff
@@ -72,68 +73,85 @@ class _DummyModel(BasicModel):
...
+                _validate_text_classification(
+                    dataset=dataset,
+                    predictions=y_pred,
+                    probabilities=y_proba,
+                    n_of_classes=len(model_classes)
+                )
                 if y_pred is not None:
-                    y_pred_dict = dict(zip(dataset.get_original_text_indexes(), y_pred))
-                    predictions.update({dataset.name: y_pred_dict})
+                    y_pred = np.array(y_pred, dtype='str')
```
noamzbr, line 134: I'm actually unsure why we do the casting to string. Is the label also cast to str?

deepchecks/nlp/input_validations.py
```diff
@@ -231,3 +235,197 @@ def compare_dataframes(
         difference = None
 
     return DataframesComparison(common, difference)
+
+
+def _validate_text_classification(
+    *,
```
noamzbr, line 241: again why we need arbitrary arg?

deepchecks/nlp/input_validations.py
```diff
@@ -231,3 +235,197 @@ def compare_dataframes(
         difference = None
 
     return DataframesComparison(common, difference)
+
+
+def _validate_text_classification(
```
noamzbr, line 240: I would imagine that the difference between these and tabular _validate_classification would be that tabular has also binary. Is there a way to unify this multiclass logic with the tabular multiclass logic?

deepchecks/nlp/context.py
yromanyshyn: sorry, not sure whether I get your question correctly, 
`*` is used to define keyword-only parameters ([docs](https://peps.python.org/pep-0570/#keyword-only-arguments)),
parameters that can be passed only as keywords 

IMO it will increase readibility

deepchecks/nlp/context.py
yromanyshyn: > I'm actually unsure why we do the casting to string.

I also do not know/

> Is the label also cast to str?

yeap

```
# input_validations.py::validate_modify_label func, line 72
labels = [None if pd.isna(x) else str(x) for x in labels]
```

deepchecks/nlp/input_validations.py
yromanyshyn: > sorry, not sure whether I get your question correctly,
> `*` is used to define keyword-only parameters ([docs](https://peps.python.org/pep-0570/#keyword-only-arguments)),
> parameters that can be passed only as keywords
> 
> IMO it will increase readibility

deepchecks/nlp/context.py
yromanyshyn: fixed

deepchecks/nlp/context.py
```diff
@@ -39,13 +39,14 @@
...
-from deepchecks.utils.validation import is_sequence_not_str
 
 TClassPred = t.Union[t.Sequence[t.Union[str, int]], t.Sequence[t.Sequence[t.Union[str, int]]]]
 TClassProba = t.Sequence[t.Sequence[float]]
+# TODO: is it correct, why tuple?
 TTokenPred = t.Sequence[t.Sequence[t.Tuple[str, int, int, float]]]
 TTextPred = t.Union[TClassPred, TTokenPred]
-TTextProba = t.Union[TClassProba]  # TODO: incorrect, why union have only one type argument?
+# TODO: incorrect, why union have only one type argument?
+TTextProba = t.Union[TClassProba]
```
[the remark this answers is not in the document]
yromanyshyn: fixed

deepchecks/nlp/context.py
yromanyshyn: I replaced it with 
```
y_pred = (np.array(y_proba) > multilabel_proba_threshold)
y_pred = y_pred.astype(int)
```

deepchecks/nlp/input_validations.py
yromanyshyn: > Is there a way to unify this multiclass logic with the tabular multiclass logic?

and Yes and No/

you know, IMO, it is one of those cases when code indeed has some similarities but differ in details and it is  not obvious how to separate those similarities into generic enough function, and if you will try you will end up with a code that is less readable and is harder to understand

for example, I do not find error messages in the `_validate_text_classification` to be suited for the tabular case, so should a unified function accept error messages templates?/

IMO, not worth overcomplicating very simple code just to reduce a few lines

deepchecks/nlp/context.py
noamzbr: 🤷‍♂️

noamzbr (approved):

deepchecks/nlp/input_validations.py
```diff
@@ -236,3 +240,209 @@ def compare_dataframes(
...
```
Nadav-Barak, line 245: _validate_text_classification_predictions
so its clear it does not validate anything else

deepchecks/nlp/input_validations.py
```diff
@@ -236,3 +240,209 @@ def compare_dataframes(
...
+        )
+        if not is_sequence_not_str(predictions):
+            raise ValidationError(format_error_message)
+        if len(predictions) != dataset.n_samples:
+            raise ValidationError(
+                f'Check requires predictions for the "{dataset.name}" dataset '
+                f'to have {dataset.n_samples} rows, same as dataset'
+            )
+        try:
+            predictions = np.array(predictions, dtype='str')
```
Nadav-Barak, line 269: will not work well with None in predictions, better to use 'object'

deepchecks/nlp/input_validations.py
yromanyshyn: it validates not only predictions

yromanyshyn:
@Nadav-Barak 

> To verify, either way predictions and probabilities return as np array?

yeap

deepchecks/nlp/input_validations.py
Nadav-Barak: if you can add test with None in the prediction for regular classification to verify everything works

yromanyshyn: added

PR #2530, opened 2023-05-14
Recsys base

RayanAAY-ops:
#### Reference Issues/PRs

fixed this:
https://github.com/deepchecks/deepchecks/pull/2498

- Arguments definition.
- Logic.
- Add explanations.
- Add arguments cutoff

---

### DIFF
--- a/deepchecks/recommender/ranking.py
+++ b/deepchecks/recommender/ranking.py
@@ -7,6 +7,7 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with Deepchecks.  If not, see <http://www.gnu.org/licenses/>.
 # ----------------------------------------------------------------------------
+#
 """Used for rec metrics."""
 from typing import List, Union,Tuple
 import numpy as np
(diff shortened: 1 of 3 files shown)


deepchecks/recommender/ranking.py
```diff
@@ -7,705 +7,428 @@
...
-    Computes the cosine similarity between the rows of two matrices
-    :params:
-    a, b   : input matrices of shape (m, n) and (k, n)
-             the matrices must share a common dimension at index 1
-    :returns:
-    cosine matrix  : np.array, an (m, k) array of the cosine similarity
-                     between the rows of a and b
-    (returns an array of shape (10, 100))
+def recall_k(relevant_items: Union[List[int], np.array, None],
+             recommendation: Union[List, np.array, None],
```
noamzbr, line 21: regarding types:
1. Can this be None? In what case?
2. Thought we decided that item could be a int or a str, so should be List[Union[str, int]] no?
3. recommendation is a single item no? Otherwise it should have been recommendations

Same comment for all the metrics here

RayanAAY-ops: The purpose of the None value is to handle cases where there are no relevant items or no recommendations available for evaluation. In such cases, the function returns a recall score of 0.0.

By allowing None as a valid input, the function provides flexibility to handle different scenarios. For example, if there are no relevant items for a user, it would not be fair to penalize the recommendation system by returning a low recall score. Similarly, if there are no recommendations generated by the system, it would also result in a recall score of 0.0.

requirements/dev-requirements.txt
```diff
@@ -48,4 +48,6 @@ beautifulsoup4>=4.11.1
 nltk<=3.6.7
 datasets
 langdetect
-textblob
\ No newline at end of file
+textblob
+
+prince==0.7.1
```
noamzbr, line 53: must it be a strict requirement?

requirements/dev-requirements.txt
```diff
@@ -48,4 +49,5 @@ beautifulsoup4>=4.11.1
 nltk<=3.6.7
 datasets
 langdetect
-textblob
\ No newline at end of file
+textblob
+transformers
```
noamzbr, line 53: you can remove this change as it was done in main already

requirements/dev-requirements.txt
```diff
@@ -20,6 +20,7 @@ deepdiff
 scikit-learn==1.0.2; python_version >= '3.7'
 # pandas 1.3.5 is the last version to support Python 3.7
 pandas==1.3.5; python_version >= '3.7'
+prince==0.7.1
```
noamzbr, line 23: Again asking - can we relax this requirement?

RayanAAY-ops: Yes, I will only need to make some changes in the diversity_score for cat features

noamzbr (approved):

PR #2532, opened 2023-05-14
Updated label display in NLP

nirhutnik:


deepchecks/nlp/text_data.py
```diff
@@ -588,13 +595,26 @@ def validate_textdata_compatibility(self, other_text_data: 'TextData') -> bool:
...
-        """Return a copy of the dataset as a pandas Dataframe with the first n_samples samples."""
+    def head(self, n_samples: int = 5, model_classes: list = None) -> pd.DataFrame:
+        """Return a copy of the dataset as a pandas Dataframe with the first n_samples samples.
+
+        Parameters
+        ----------
+        n_samples : int, default 5
+            Number of samples to return.
+        model_classes : list, default None
+            List of classes names to use for multi-label display. Only used if the dataset is multi-label.
```
noamzbr, line 606: if is classification but 0,..,n-1?

tests/nlp/checks/train_test_validation/property_drift_test.py
```diff
@@ -181,7 +181,7 @@ def test_with_drift(self, dummy_multilabel_textdata_train_test):
         condition_results = check.conditions_decision(result)
         assert_that(condition_results, has_items(
             equal_condition_result(is_pass=False,
-                                   details="Failed for 1 out of 6 columns.\nFound 1 "
+                                   details="Failed for 1 out of 7 columns.\nFound 1 "
```
noamzbr, line 184: why did that change?

noamzbr (approved):

deepchecks/nlp/text_data.py
nirhutnik: What? you mean if there are no names? then... you don't have to use it.

tests/nlp/checks/train_test_validation/property_drift_test.py
nirhutnik: There was a funny bug - 
Some time ago we removed the language property from calculations because it wasn't good enough. And then we changed it to use fasttext instead of langdetect and reinserted it. However, the tests were not updated, and still didn't fail. Why? Because CI/CD didn't install fasttext, so it skipped it. So the tests failed locally, but not remotely.
So I re-added fasttext to dev-requirements.txt and that changed the outcomes of some tests.

noamzbr: what why was the result skipping rather than failing... you understand that?

deepchecks/nlp/text_data.py
noamzbr: mm right sry

tests/nlp/checks/train_test_validation/property_drift_test.py
nirhutnik: Because that's how properties are built :) We skip properties that the user doesn't have the package for

nirhutnik: (We print a warning, not raising an error)

ItayGabbay (approved):

tests/nlp/checks/train_test_validation/property_drift_test.py
noamzbr: oh
right
and not test per property

PR #2538, opened 2023-05-15
Extended just dance dataset + small fixes

Nadav-Barak:


deepchecks/nlp/datasets/classification/just_dance_comment_analysis.py
```diff
@@ -28,26 +28,132 @@
...
+    use_full_size : bool, default: False
+        If True, the returned data will be the full dataset, otherwise returns a subset of the data.
+    Returns
+    -------
+    predictions : np.ndarray
+        The prediction of the data elements in the dataset.
+
+    """
+    if use_full_size:
+        raise NotImplementedError('Predictions for the full dataset are not yet available.')
```
yromanyshyn, line 78: why is this parameter then even needed? maybe drop it?

deepchecks/nlp/checks/model_evaluation/weak_segments_performance.py
```diff
@@ -63,6 +62,10 @@ def run_logic(self, context: Context, dataset_kind) -> CheckResult:
         # Decide which scorer and score_per_sample to use in the algorithm run
         is_multilabel = text_data.is_multi_label_classification()
         if is_multilabel:
+            if self.alternative_scorer is None:
+                self.alternative_scorer = {'Precision Macro': 'precision_macro'}
```
noamzbr, line 66: why not f1? precision optimizes for lowering FP while our loss is more "balanced" between FP and FN, which may create a situation in which a weak segment is detected that has a higher / equivalent precision compared to baseline, while it's loss is higher.

deepchecks/nlp/datasets/classification/just_dance_comment_analysis.py
noamzbr: For compatibility with the other loading functions in this file I would imagine. 
@Nadav-Barak I would think though that the reasonable default for use_full_data is true.

deepchecks/nlp/datasets/classification/just_dance_comment_analysis.py
```diff
@@ -28,26 +28,132 @@
...
+        If True, the returned data is split into train and test exactly like the toy model
+        was trained. The first return value is the train data and the second is the test data.
+        Otherwise, returns a single object.
+    use_full_size : bool, default: False
+        If True, the returned data will be the full dataset, otherwise returns a subset of the data.
+
+    Returns
+    -------
+    embeddings : np.ndarray
+        Embeddings for the tweet_emotion dataset.
```
noamzbr, line 111: Not a tweet. Do ctrl+f for "tweet" in this file there are a couple of docstrings that need updating.

deepchecks/nlp/datasets/classification/just_dance_comment_analysis.py
```diff
@@ -75,27 +185,28 @@ def load_data(data_format: str = 'TextData', as_train_test: bool = True, use_ful
...
         return dataset
 
     else:
-        if use_full_size:
-            train = data[data['dateComment'] < '2015-01-01']
-            test = data[data['dateComment'] >= '2015-01-01']
-        else:
-            train = data[data['dateComment'] < '2013-01-01']
-            test = data[data['dateComment'] >= '2021-01-01']
+        train = data[data[_TIME_COL] < _DATE_TO_SPLIT_BY]
```
noamzbr, line 208: can't we use _get_train_test_indexes here as well?

deepchecks/nlp/utils/text_properties.py
```diff
@@ -282,7 +282,7 @@ def _predict(text, classifier, kind):
             return v['score']
         elif kind == 'fluency':
             label_value = 'LABEL_1'
-        elif kind == 'fluency':
+        elif kind == 'formality':
```
noamzbr, line 285: I know this is not related to this PR but I regret this code and think the treatment of prediction for each property should be within each property, as in this implementation the logic is split between two functions and it's confusing...

deepchecks/nlp/datasets/classification/just_dance_comment_analysis.py
Nadav-Barak: The partial size is good enough, 15K samples

deepchecks/nlp/datasets/classification/just_dance_comment_analysis.py
Nadav-Barak: we can, will update

deepchecks/nlp/utils/text_properties.py
Nadav-Barak: i can change it none the less, any objections @yromanyshyn ?

deepchecks/nlp/checks/model_evaluation/weak_segments_performance.py
Nadav-Barak: Not a strong preference, changed

deepchecks/nlp/datasets/classification/just_dance_comment_analysis.py
Nadav-Barak: you right, done

noamzbr (approved):

PR #2558, opened 2023-05-24
Add min_samples and n_top_classes to TrainTestPerformance in NLP

nirhutnik:
image

### DIFF
--- a/deepchecks/nlp/checks/model_evaluation/train_test_performance.py
+++ b/deepchecks/nlp/checks/model_evaluation/train_test_performance.py
@@ -177,11 +177,11 @@ class TrainTestPerformance(TrainTestPerformanceAbstract, TrainTestCheck):
 
         # Nullify rows with less than min_samples:
         results_df.loc[results_df['Number of samples'] < self.min_samples, 'Value'] = None
-        classes_without_enough_samples = results_df[results_df['Value'].isna()]['Class'].unique().tolist()
+        classes_without_enough_samples = results_df[results_df['Class'].notna()][results_df['Value'].isna()]['Class'].unique().tolist()
 
         # Show only top n classes:
         if self.n_top_classes:
-            samples_per_class = results_df[['Class', 'Dataset', 'Number of samples']].drop_duplicates()
+            samples_per_class = results_df[results_df['Class'].notna()][['Class', 'Dataset', 'Number of samples']].drop_duplicates()
             samples_per_class = samples_per_class[~samples_per_class['Class'].isin(classes_without_enough_samples)]
 
             if self.show_classes_by == 'train_largest':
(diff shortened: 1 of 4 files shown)


deepchecks/nlp/checks/model_evaluation/train_test_performance.py
```diff
@@ -33,6 +33,17 @@ class TrainTestPerformance(TrainTestPerformanceAbstract, TrainTestCheck):
     scorers: Union[Mapping[str, Union[str, Callable]], List[str]], default: None
         Scorers to override the default scorers, find more about the supported formats at
         https://docs.deepchecks.com/stable/user-guide/general/metrics_guide.html
+    min_samples: int, default: 30
+        Minimum number of samples required to calculate model performance. If scorer is per class, then
+        this is the minimum number of samples per class.
+    n_top_classes: int, default: 10
+        Number of top classes to show in the label graphs. The top classes are determined by the show_classes_by
+        parameter. if None, then all classes are shown.
+    show_classes_by: str, default: 'test_largest'
```
Nadav-Barak, line 42: Maybe also option of largest diff?

deepchecks/nlp/checks/model_evaluation/train_test_performance.py
```diff
@@ -33,6 +33,17 @@ class TrainTestPerformance(TrainTestPerformanceAbstract, TrainTestCheck):
     scorers: Union[Mapping[str, Union[str, Callable]], List[str]], default: None
         Scorers to override the default scorers, find more about the supported formats at
         https://docs.deepchecks.com/stable/user-guide/general/metrics_guide.html
+    min_samples: int, default: 30
```
Nadav-Barak, line 36: hmm, maybe 10? for score per class I think its a bit harsh (sometime we will be evaluated on a small dataset)

deepchecks/utils/abstracts/train_test_performace.py
```diff
@@ -137,3 +163,35 @@ def add_condition_class_performance_imbalance_ratio_less_than(
         name = f"Relative ratio difference between labels '{score}' score is less than {format_percent(threshold)}"
         condition = get_condition_class_performance_imbalance_ratio_less_than(threshold=threshold, score=score)
         return self.add_condition(name=name, condition_func=condition)
+
+# def _filter_classes_without_enough_samples(results_df, class_list, train_dataset_name, test_dataset_name) \
```
Nadav-Barak, line 167: comment

deepchecks/nlp/checks/model_evaluation/train_test_performance.py
nirhutnik: It's about significance as well. you'll have a drop of 10% for a change in one sample out of 10. I think 10 is too low, and 30 is a common min sample size.

nirhutnik: I mean, in general, if my metrics reduced on such a small class, I also wouldn't care...

deepchecks/nlp/checks/model_evaluation/train_test_performance.py
nirhutnik: Also, the issue is - by which metric? And for each graph separately? It really makes it complicatd :/

PR #2561, opened 2023-05-28
data duplicates takes a long time for long docs

noamzbr:
Solution (idea by @nirhutnik) is to further inspect only samples identified as duplicates when looking at their first 200 characters.

### DIFF
--- a/deepchecks/nlp/checks/data_integrity/conflicting_labels.py
+++ b/deepchecks/nlp/checks/data_integrity/conflicting_labels.py
@@ -117,7 +117,7 @@ class ConflictingLabels(SingleDatasetCheck, ConflictingLabelsAbstract):
             'Text': dataset.text,
         })
 
-        return df, labels
+        return df
 
     def run_logic(self, context: Context, dataset_kind) -> CheckResult:
         """Run check."""
@@ -132,7 +132,7 @@ class ConflictingLabels(SingleDatasetCheck, ConflictingLabelsAbstract):
         # Reduce dataset by first checking on truncated strings
         truncated_samples = [cut_string(x) for x in dataset.text]
 
-        df, labels = self._get_df_and_labels(dataset, truncated_samples)
+        df = self._get_df_and_labels(dataset, truncated_samples)
 
         ambiguous_samples_hashes = self._get_conflicting_indices(df)
         indices_to_reinspect = df[df['hash'].isin(ambiguous_samples_hashes)].index.to_list()
@@ -146,7 +146,7 @@ class ConflictingLabels(SingleDatasetCheck, ConflictingLabelsAbstract):
             return CheckResult(value=result_value)
 
         # Now that we have narrowed down the dataset, we can check on full strings
-        df, labels = self._get_df_and_labels(dataset, dataset.text)
+        df = self._get_df_and_labels(dataset, dataset.text)
 
         ambiguous_samples_hashes = self._get_conflicting_indices(df)
 
--- a/deepchecks/nlp/checks/data_integrity/conflicting_labels.py
+++ b/deepchecks/nlp/checks/data_integrity/conflicting_labels.py
@@ -102,6 +102,23 @@ class ConflictingLabels(SingleDatasetCheck, ConflictingLabelsAbstract):
         ambiguous_samples_hashes = n_of_labels_per_sample[n_of_labels_per_sample > 1]
         return frozenset(ambiguous_samples_hashes.index.to_list())
 
+    def _get_df_and_labels(self, dataset, samples):
+        samples_hashes = hash_samples(normalize_samples(
+            samples,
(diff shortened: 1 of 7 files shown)


deepchecks/nlp/checks/data_integrity/conflicting_labels.py
```diff
@@ -82,30 +82,69 @@ def _text_normalization_kwargs(self):
...
+        ))
+
+        labels = self._get_labels(dataset)
+
+        df = pd.DataFrame({
+            'hash': truncated_samples_hashes,
+            'Sample ID': dataset.get_original_text_indexes(),
+            'Label': labels,
+            'Text': dataset.text,
+        })
```
nirhutnik, line 129: Why not take this out into an outer function and just run it twice? seems like a lot of code duplication

deepchecks/nlp/checks/data_integrity/conflicting_labels.py
```diff
@@ -82,30 +82,69 @@ def _text_normalization_kwargs(self):
...
+
+        ambiguous_samples_hashes = self._get_conflicting_indices(df)
+        indices_to_reinspect = df[df['hash'].isin(ambiguous_samples_hashes)].index.to_list()
+        dataset = dataset.copy(rows_to_use=indices_to_reinspect)
+
+        if len(dataset) == 0:
+            result_value = dict(percent_of_conflicting_samples=0, conflicting_samples=pd.DataFrame(
+                index=pd.MultiIndex(levels=[[], [], []], codes=[[], [], []], names=['Duplicate', 'Sample ID', 'Label']),
+                columns=['Text']))
+            return CheckResult(value=result_value)
```
nirhutnik, line 139: Not sure what happens here... But wouldn't make more sense to switch it, put "if len(dataset) !=0", run the thingy again. and then continue with the code that works anyway if there are duplications or if there aren't?

deepchecks/nlp/checks/data_integrity/conflicting_labels.py
```diff
@@ -46,17 +46,17 @@ class ConflictingLabels(SingleDatasetCheck, ConflictingLabelsAbstract):
...
-        remove_punctuation: bool = True,
-        normalize_unicode: bool = True,
-        remove_stopwords: bool = True,
-        ignore_whitespace: bool = False,
-        n_to_show: int = 5,
-        n_samples: int = 10_000_000,
-        random_state: int = 42,
-        max_text_length_for_display: int = 30,
-        **kwargs
+            self,
```
nirhutnik, line 49: In general - missing tests for these. For instance, a check that makes sure that samples like 'a'*1000 aren't considered the same as 'a'*2000

deepchecks/nlp/utils/text.py
```diff
@@ -117,6 +117,15 @@ def normalize_text(
     return text_sample
 
 
+def cut_string(input_str: str, cut_length: int = 200) -> str:
+    """Cut a string to 200 characters, but cut only at whitespaces."""
+    if len(input_str) > cut_length:
```
nirhutnik, line 122: Why does it matter if it's in whitespace or not?
and out of curiousity - why 200?

deepchecks/nlp/checks/data_integrity/conflicting_labels.py
nirhutnik: I mean, I know you did that. But even the hashing and creation of the df

deepchecks/nlp/checks/data_integrity/text_duplicates.py
```diff
@@ -83,7 +83,29 @@ def run_logic(self, context: Context, dataset_kind):
...
+        grouped_samples_truncated = df_truncated.groupby(by=['hash'], dropna=False, group_keys=True)
+        reinspect_idx = df_truncated[grouped_samples_truncated['Text'].transform('count') > 1].index.to_list()
+        # At this stage, what was detected as unique is really unique
+        n_of_unique += sum(grouped_samples_truncated['Text'].transform('count') == 1)
+
+        # Reinspect samples that are truncated
+        dataset = dataset.copy(rows_to_use=reinspect_idx)
+        if len(dataset) == 0:
+            return CheckResult(value={'percent_of_duplicates': 0,
+                                      'duplicates': pd.DataFrame()})
```
nirhutnik, line 105: All comments as in conflicting labels

nirhutnik (approved):

deepchecks/nlp/utils/text.py
noamzbr: whitspace - so not to cut mid-word
200 - just so?

nirhutnik: yeah, but what does it matter if you cut mid-word? you're just doing this to compare duplicates...
And was curious, seems like an odd number

PR #2565, opened 2023-05-28
Nir/dee 638 long docs sentimentsubjectivity take a long time

nirhutnik:
Solved in 2 ways:
1. Common cache for subjectivity and sentiment
2. sampling of text into single words. Also enabled for other purposes as this will be done for other properties soon.

Here are the differences in results when sampling 10,000 words out of long texts (sample size = 162)

You can see the mean, mean error, and std:

image


deepchecks/nlp/utils/text_properties.py
```diff
@@ -35,6 +37,41 @@
...
+        The mode to sample in. Can be either 'characters', 'words' or 'sentences'.
+    limit : int, default 1000
+        The maximum number of words or sentences to sample.
+    """
+    np.random.seed(random_seed)
+    if mode == 'characters':
+        if len(text) <= limit:
+            return text
+        else:
+            all_units = np.random.choice(list(text), size=limit, replace=False)
```
noamzbr, line 60: Dunno in what case you'll use that, but I would imagine that you'd want to sample in order

deepchecks/nlp/utils/text_properties.py
```diff
@@ -35,6 +37,41 @@
...
+            return ''.join(all_units)
+    elif mode == 'words':
+        all_units = re.split(r'\W+', text)
+        if len(all_units) > limit:
+            all_units = np.random.choice(all_units, size=limit, replace=False)
```
noamzbr, line 65: 2 things:
1. What will this regex do with punctuation? 
2. Again think this should be while keeping the original order.

deepchecks/nlp/utils/text_properties.py
```diff
@@ -35,6 +37,41 @@
...
+        return ' '.join(all_units)
+    elif mode == 'sentences':
+        all_units = sent_tokenize(text)
```
noamzbr, line 68: need to do the "try download" thing like other places where downloadable stuff is used
Is downloadable stuff is used?

deepchecks/nlp/utils/text_properties.py
```diff
@@ -260,14 +297,24 @@ def language(
     return language_codes
 
 
-def sentiment(raw_text: Sequence[str]) -> List[str]:
+def sentiment(raw_text: Sequence[str]) -> List[float]:
     """Return list of floats of sentiment."""
-    return [textblob.TextBlob(text).sentiment.polarity for text in raw_text]
+    if text_blob_cache.get('textblob') is None:
+        # TextBlob uses only the words and not the relations between them, so we can sample the text
```
noamzbr, line 303: So sampling is a good idea, but should really replace textblob...

deepchecks/nlp/utils/text_properties.py
```diff
@@ -260,14 +297,24 @@ def language(
...
+        # to speed up the process:
+        raw_text = [_sample_for_property(text=text, mode='words') for text in raw_text]
+        text_blob_cache['textblob'] = [textblob.TextBlob(text).sentiment for text in raw_text]
```
noamzbr, line 306: what is the typical RAM size of this cache for a largish dataset?

deepchecks/nlp/utils/text_properties.py
nirhutnik: Thought in Special Characters, actually just removed later as it didn't work for it

deepchecks/nlp/utils/text_properties.py
nirhutnik: Technically already done in the property itself. But can move to there.

deepchecks/nlp/utils/text_properties.py
nirhutnik: Why? And do you have an alternative?

deepchecks/nlp/utils/text_properties.py
nirhutnik: N * size(float). Though doesn't really matter, think that you're keeping it in memory anyway as 2 properties a few miliseconds later

deepchecks/nlp/utils/text_properties.py
noamzbr: future usage

deepchecks/nlp/utils/text_properties.py
noamzbr: In case this code is later called by other properties

deepchecks/nlp/utils/text_properties.py
noamzbr: oh right thought it's a cache also for the strings

deepchecks/nlp/utils/text_properties.py
nirhutnik: I don't see a relevant use-case for now. And if we'll need, we can change, and make sure it doesn't affect the scores. You don't agree? No need for this. Also, complicates the code IMO

deepchecks/nlp/utils/text_properties.py
nirhutnik: Done in the other PR

deepchecks/nlp/utils/text_properties.py
nirhutnik: Well there's one for that in the next PR, you're in for a treat. But it's technically limited.

PR #2568, opened 2023-05-29
reorder checks in suites NLP

Nadav-Barak:
Make our coolest checks appear first


shir22:
Looks good, except that I'd move the prediction drift to top (commented also on original message)

PR #2573, opened 2023-06-01
Embed long documents with openai

noamzbr:
Embed long documents using the recipe from https://github.com/openai/openai-cookbook/blob/main/examples/Embedding_long_inputs.ipynb

### DIFF
--- a/deepchecks/nlp/text_data.py
+++ b/deepchecks/nlp/text_data.py
@@ -302,7 +302,8 @@ class TextData:
             )
         return self._embeddings
 
-    def calculate_builtin_embeddings(self, model: str = 'miniLM', file_path: str = 'embeddings.npy'):
+    def calculate_builtin_embeddings(self, model: str = 'miniLM', file_path: str = 'embeddings.npy',
+                                     device: t.Optional[str] = None):
         """Calculate the built-in embeddings of the dataset.
 
         Parameters
@@ -313,11 +314,13 @@ class TextData:
             'open_ai': using the ADA model in the open_ai library. Requires an API key.
         file_path : str, default: 'embeddings.npy'
             The path to save the embeddings to.
+        device : str, default: None
+            The device to use for calculating the embeddings. If None, the default device will be used.
         """
         if self._embeddings is not None:
             warnings.warn('Embeddings already exist, overwriting them', UserWarning)
 
-        self._embeddings = calculate_builtin_embeddings(text=self.text, model=model, file_path=file_path)
+        self._embeddings = calculate_builtin_embeddings(text=self.text, model=model, file_path=file_path, device=device)
 
     def set_embeddings(self, embeddings: np.ndarray, verbose: bool = True):
         """Set the embeddings of the dataset.
--- a/deepchecks/nlp/text_data.py
+++ b/deepchecks/nlp/text_data.py
@@ -303,7 +303,7 @@ class TextData:
         return self._embeddings
 
     def calculate_builtin_embeddings(self, model: str = 'miniLM', file_path: str = 'embeddings.npy',
-                                     device: t.Optional[str] = None, long_sample_averaging: str = 'average+warn',
+                                     device: t.Optional[str] = None, long_sample_behaviour: str = 'average+warn',
                                      open_ai_batch_size: int = 500):
         """Calculate the built-in embeddings of the dataset.
 
@@ -317,7 +317,7 @@ class TextData:
             The path to save the embeddings to.
         device : str, default: None
             The device to use for calculating the embeddings. If None, the default device will be used.
-        long_sample_averaging : str, default 'average+warn'
+        long_sample_behaviour : str, default 'average+warn'
             How to handle long samples. Averaging is done as described in
             https://github.com/openai/openai-cookbook/blob/main/examples/Embedding_long_inputs.ipynb
             Currently, applies only to the 'open_ai' model, as the 'miniLM' model can handle long samples.
@@ -335,7 +335,7 @@ class TextData:
             warnings.warn('Embeddings already exist, overwriting them', UserWarning)
 
         self._embeddings = calculate_builtin_embeddings(text=self.text, model=model, file_path=file_path, device=device,
-                                                        long_sample_averaging=long_sample_averaging,
+                                                        long_sample_behaviour=long_sample_behaviour,
                                                         open_ai_batch_size=open_ai_batch_size)
 
     def set_embeddings(self, embeddings: np.ndarray, verbose: bool = True):
--- a/deepchecks/nlp/text_data.py
+++ b/deepchecks/nlp/text_data.py
@@ -316,7 +316,7 @@ class TextData:
             The path to save the embeddings to.
         device : str, default: None
             The device to use for calculating the embeddings. If None, the default device will be used.
-        long_doc_averaging : str, default 'warn'
+        long_doc_averaging : str, default 'average+warn'
             How to handle long documents (docments that are longer than the model context window).
             Can be either 'average+warn', 'average', 'truncate' or 'raise'. Averaging is done as described in
             https://github.com/openai/openai-cookbook/blob/main/examples/Embedding_long_inputs.ipynb
@@ -326,6 +326,7 @@ class TextData:
                 - 'average': average the embeddings of the chunks.
                 - 'truncate': truncate the document to the maximum length.
                 - 'raise': raise an error if the document is too long.
+                - 'nan': return an embedding vector of nans for each document that is too long.
         """
         if self._embeddings is not None:
             warnings.warn('Embeddings already exist, overwriting them', UserWarning)
--- a/deepchecks/nlp/text_data.py
+++ b/deepchecks/nlp/text_data.py
@@ -303,7 +303,7 @@ class TextData:
         return self._embeddings
 
     def calculate_builtin_embeddings(self, model: str = 'miniLM', file_path: str = 'embeddings.npy',
-                                     device: t.Optional[str] = None):
+                                     device: t.Optional[str] = None, long_doc_averaging: str = 'average+warn'):
         """Calculate the built-in embeddings of the dataset.
 
         Parameters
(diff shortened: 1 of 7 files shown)


deepchecks/nlp/utils/text_embeddings.py
```diff
@@ -9,14 +9,52 @@
...
+def iterate_batched(tokenized_text, chunk_length):
+    """Chunk text into tokens of length chunk_length."""
+    chunks_iterator = batched(tokenized_text, chunk_length)
+    yield from chunks_iterator
+
 
 def calculate_builtin_embeddings(text: np.array, model: str = 'miniLM',
-                                 file_path: Optional[str] = 'embeddings.npy') -> np.array:
+                                 file_path: Optional[str] = 'embeddings.npy',
+                                 device: Optional[str] = None) -> np.array:
```
Nadav-Barak, line 57: we discussed an argument to control what to do with texts longer then limit and that by default will be average with warning/exception

Nadav-Barak: the warnings / exception should state with samples were longer then limit + what is the limit (which is based on selected model i guess)

deepchecks/nlp/utils/text_embeddings.py
```diff
@@ -30,6 +68,8 @@ def calculate_builtin_embeddings(text: np.array, model: str = 'miniLM',
         by using the command openai.api_key = YOUR_API_KEY
```
Nadav-Barak, line 68: Need to add info on that process we are doing + ref link to open ai cook book

deepchecks/nlp/utils/text_embeddings.py
noamzbr: Can't state which samples, can be too many. Will length of maximal one

deepchecks/nlp/utils/text_embeddings.py
noamzbr: Added

deepchecks/nlp/text_data.py
```diff
@@ -313,11 +314,24 @@ def calculate_builtin_embeddings(self, model: str = 'miniLM', file_path: str = '
             'open_ai': using the ADA model in the open_ai library. Requires an API key.
         file_path : str, default: 'embeddings.npy'
             The path to save the embeddings to.
+        device : str, default: None
+            The device to use for calculating the embeddings. If None, the default device will be used.
+        long_doc_averaging : str, default 'warn'
```
Nadav-Barak, line 319:
```suggestion
        long_doc_averaging : str, default ''average+warn''
```

deepchecks/nlp/text_data.py
```diff
@@ -313,11 +314,24 @@ def calculate_builtin_embeddings(self, model: str = 'miniLM', file_path: str = '
...
+            How to handle long documents (docments that are longer than the model context window).
+            Can be either 'average+warn', 'average', 'truncate' or 'raise'. Averaging is done as described in
+            https://github.com/openai/openai-cookbook/blob/main/examples/Embedding_long_inputs.ipynb
+
+            Currently, applies only to the 'open_ai' model, as the 'miniLM' model can handle long documents.
```
Nadav-Barak, line 324: Add a todo, i guess for really long documents it cant

deepchecks/nlp/text_data.py
Nadav-Barak, line 319: Add a option of return None for long samples

deepchecks/nlp/text_data.py
```diff
@@ -313,11 +314,24 @@ def calculate_builtin_embeddings(self, model: str = 'miniLM', file_path: str = '
             'open_ai': using the ADA model in the open_ai library. Requires an API key.
         file_path : str, default: 'embeddings.npy'
             The path to save the embeddings to.
+        device : str, default: None
+            The device to use for calculating the embeddings. If None, the default device will be used.
+        long_doc_averaging : str, default 'warn'
+            How to handle long documents (docments that are longer than the model context window).
+            Can be either 'average+warn', 'average', 'truncate' or 'raise'. Averaging is done as described in
```
Nadav-Barak, line 321: alternatives to consider: 
1. always warn
2. default is raise and if average is selected dont warn

deepchecks/nlp/utils/text_embeddings.py
```diff
@@ -30,6 +70,17 @@ def calculate_builtin_embeddings(text: np.array, model: str = 'miniLM',
...
     file_path : Optional[str], default 'embeddings.csv'
         If given, the embeddings will be saved to the given file path.
+    device : str, default None
+        The device to use for the embeddings. If None, the default device will be used.
+    long_doc_averaging : str, default 'warn'
+        How to handle long documents (docments that are longer than the model context window).
+         Can be either 'average+warn', 'average', 'truncate' or 'raise'.
+         Currently, applies only to the 'open_ai' model, as the 'miniLM' model can handle long documents. Averaging is
+         done as described in https://github.com/openai/openai-cookbook/blob/main/examples/Embedding_long_inputs.ipynb
```
Nadav-Barak, line 79: same here but lets discuss in above thread

requirements/nlp-requirements.txt
```diff
@@ -1,4 +1,6 @@
 seqeval>=1.0.0
 nltk>=3.4.0,<=3.6.7
 textblob>=0.17.1
-umap-learn
\ No newline at end of file
+umap-learn
+transformers>=4.0.0
+tiktoken; python_version >= '3.8'
```
Nadav-Barak, line 6: must?

deepchecks/nlp/text_data.py
noamzbr: it can, the problem is just with open_ai

deepchecks/nlp/text_data.py
noamzbr: 1. Against - has to be an option to average without warning
2. I think you want things to "work" out of the box.

requirements/nlp-requirements.txt
noamzbr: yes

deepchecks/nlp/utils/text_embeddings.py
noamzbr: Not sure what is same here

deepchecks/nlp/text_data.py
nirhutnik: (but without saying doc, see above comment)

deepchecks/nlp/utils/text_embeddings.py
```diff
@@ -30,6 +70,18 @@ def calculate_builtin_embeddings(text: np.array, model: str = 'miniLM',
         by using the command openai.api_key = YOUR_API_KEY
     file_path : Optional[str], default 'embeddings.csv'
         If given, the embeddings will be saved to the given file path.
+    device : str, default None
+        The device to use for the embeddings. If None, the default device will be used.
+    long_doc_averaging : str, default 'warn'
+        How to handle long documents (docments that are longer than the model context window).
+         Can be either 'average+warn', 'average', 'truncate' or 'raise'.
```
nirhutnik, line 77: If you're describing the options later, no need to say them here

deepchecks/nlp/utils/text_embeddings.py
```diff
@@ -53,21 +105,74 @@ def calculate_builtin_embeddings(text: np.array, model: str = 'miniLM',
...
+                                  f'found is {max_doc_length} tokens. The document will be split into chunks and the '
+                                  f'embeddings will be averaged. To avoid this warning, set '
+                                  f'long_doc_averaging="average" or long_doc_averaging="truncate".')
+                elif long_doc_averaging == 'raise':
+                    raise ValueError(f'At least one document is longer than {max_tokens} tokens, which is the maximum '
+                                     f'context window handled by {model}. Maximal document '
+                                     f'length found is {max_doc_length} tokens. To avoid this error, set '
+                                     f'long_doc_averaging="average" or long_doc_averaging="truncate".')
+
+            batch_size = 500
```
nirhutnik, line 148: should be configurable parameter

deepchecks/nlp/utils/text_embeddings.py
```diff
@@ -53,21 +105,74 @@ def calculate_builtin_embeddings(text: np.array, model: str = 'miniLM',
...
+
+            result_embeddings = []
+            idx = 0
+            for tokens_in_sample in encoded_texts:
+                text_embeddings = []
+                text_lens = []
+                while idx < len(chunk_lens) and sum(text_lens) < len(tokens_in_sample):
+                    text_embeddings.append(chunk_embeddings[idx])
+                    text_lens.append(chunk_lens[idx])
+                    idx += 1
```
nirhutnik, line 163: Why the while loop? Can be just replaced with 2 list comprehensions... no?

deepchecks/nlp/utils/text_embeddings.py
```diff
@@ -53,21 +105,74 @@ def calculate_builtin_embeddings(text: np.array, model: str = 'miniLM',
...
+
+                if long_doc_averaging == 'nan' and len(text_embeddings) > 1:
+                    text_embedding = np.ones_like(text_embeddings[0]) * np.nan
```
nirhutnik, line 166: If I get this correctly, you still calculate the text embeddings for all the chunks even if you truncate, you just don't use it in the end? I don't think this is good, this wastes a lot of resources + the user pays for each API call.

deepchecks/nlp/utils/text_properties.py
```diff
@@ -39,6 +39,7 @@
 MODELS_STORAGE = pathlib.Path(__file__).absolute().parent / '.nlp-models'
 FASTTEXT_LANG_MODEL = 'https://dl.fbaipublicfiles.com/fasttext/supervised-models/lid.176.bin'
 DEFAULT_SENTENCE_SAMPLE_SIZE = 300
+MAX_CHARS = 720  # Bert accepts max of 512 tokens, and experimentally 720 chars seems to work while 1024 doesn't
```
nirhutnik, line 42: I mean.
Technically, a user can have a string where each character is a token.

deepchecks/nlp/utils/text_properties.py
```diff
@@ -344,6 +345,21 @@ def subjectivity(text: str) -> float:
 def _predict(text: str, classifier, kind: str) -> float:
     """Return prediction of huggingface Pipeline classifier."""
     try:
+        # TODO: make this way smarter, and not just a hack. Count tokens, for a start. Then not just sample sentences.
+        # If text is longer than classifier context window, sample it:
+        if len(text) > MAX_CHARS:
```
nirhutnik, line 350: Why not do the tokenization hack you have for open ai?

deepchecks/nlp/utils/text_embeddings.py
noamzbr: because of the sum len condition

deepchecks/nlp/utils/text_properties.py
noamzbr: yeah this is more a patch then a complete solution

deepchecks/nlp/text_data.py
noamzbr: I think that though seperating is more "complete", it's simpler and easier to have just one parameter controlling everything about the behaviour for long docs. Also, the logic is that if you explicitly set a method, you don't need a warning for it ever.

nirhutnik: I really think this is too complicated.
And I get the warning, but Keep It Simple. I think this is a complicated solution that's not user friendly. Never seen a parameter that does 2 things like that, and that they are completely unrelated.

deepchecks/nlp/utils/text_embeddings.py
nirhutnik: But the text_lens can't be over the number of tokens in sample, can it? you run on batches anyway...

deepchecks/nlp/utils/text_properties.py
noamzbr: I'd have to use the specific tokenizer of each model here, decided to avoid for now...

nirhutnik (approved):

deepchecks/nlp/utils/text_properties.py
noamzbr: will do

PR #2578, opened 2023-06-03
Use second column level as type indicator in properties

noamzbr:
The property dataframe has two column level - first is the propety name and second is the data type
In csv this means the two first rows are these two column headers


noamzbr:
Closed, will do https://github.com/deepchecks/deepchecks/pull/2579 instead

PR #2597, opened 2023-06-13
Setting any non-numeric value to nan and tests

harsh-deepchecks:
#### Reference Issues/PRs

Fixes DEE - 619 Text Property Outliers fails when only one property fails

Seems to fails when there are non-numeric values in properties. So, if there are such values, we can replace those with np.nan and then have a error message associated since plotting the graph for that property would be incorrect as there are non-numeric values for numeric properties.

---


harsh-deepchecks:
Below is the plot for two properties where I added non-numeric data such as string "hello" as one of the values for "Sentences Count" property and at the end we display the error message. 

A suggestion -> If we add try-catch, then I think we can capture all other exceptions as well and display them in a similar manner? 

image

deepchecks/nlp/checks/data_integrity/text_property_outliers.py
```diff
@@ -81,10 +81,17 @@ def run_logic(self, context: Context, dataset_kind: DatasetKind) -> CheckResult:
         for name, values in properties.items():
             # If the property is single value per sample, then wrap the values in list in order to work on fixed
```
Nadav-Barak, line 82: make sure this comment is next to the relevant place + add comment explaining your newly added code

deepchecks/nlp/checks/data_integrity/text_property_outliers.py
```diff
@@ -129,7 +136,8 @@ def run_logic(self, context: Context, dataset_kind: DatasetKind) -> CheckResult:
                 # we have in the data
                 'lower_limit': max(lower_limit, min(values_arr)),
                 'upper_limit': min(upper_limit, max(values_arr)) if is_numeric else None,
-                'outlier_ratio': len(text_outliers) / len(values_arr)
+                'outlier_ratio': len(text_outliers) / len(values_arr),
+                'property_error': property_error
```
Nadav-Barak, line 140: would change this, 
if a property outlier calculation fails - it shouldn't have the same structure in the result dict
specifically after you add the try - catch
under the catch would set result[name] = 'exception error msg'

deepchecks/nlp/checks/data_integrity/text_property_outliers.py
```diff
@@ -81,10 +81,17 @@ def run_logic(self, context: Context, dataset_kind: DatasetKind) -> CheckResult:
...
+            is_numeric = name not in cat_properties
+            property_error = None
             if not isinstance(values[0], list):
-                values = [[x] for x in values]
+                if is_numeric:
+                    curr_nan_count = pd.isnull(values).sum()
+                    values = pd.to_numeric(values, errors='coerce')
+                    updated_nan_count = pd.isnull(values).sum()
+                    if updated_nan_count > curr_nan_count:
+                        property_error = 'Numeric property contains non-numeric values.'
```
Nadav-Barak, line 92: would here raise exception - no need to run outlier calculation for properties like this

deepchecks/nlp/checks/data_integrity/text_property_outliers.py
harsh-deepchecks: Done

deepchecks/nlp/checks/data_integrity/text_property_outliers.py
harsh-deepchecks: Yes, added try-except block

deepchecks/nlp/checks/data_integrity/text_property_outliers.py
harsh-deepchecks: Done

deepchecks/nlp/checks/data_integrity/text_property_outliers.py
```diff
@@ -79,99 +79,118 @@ def run_logic(self, context: Context, dataset_kind: DatasetKind) -> CheckResult:
...
-
-                            display.append(fig)
-                        else:
-                            no_outliers = pd.concat(
-                                [no_outliers, pd.Series(property_name, index=[
-                                    f'Not enough non-null samples to compute'
-                                    f' properties (min_samples={self.min_samples}).'
-                                ])]
-                            )
+                try:
```
Nadav-Barak, line 159: why we need try here?

deepchecks/nlp/checks/data_integrity/text_property_outliers.py
```diff
@@ -79,99 +79,118 @@ def run_logic(self, context: Context, dataset_kind: DatasetKind) -> CheckResult:
...
+                    if isinstance(info, str):
+                        no_outliers = pd.concat([no_outliers, pd.Series(property_name, index=[info])])
+                    elif len(info['indices']) == 0:
+                        no_outliers = pd.concat([no_outliers, pd.Series(property_name, index=['No outliers found.'])])
                     else:
-                        no_outliers = pd.concat([no_outliers, pd.Series(property_name, index=[
-                            f'Outliers found but not shown in graphs (n_show_top={self.n_show_top}).'])])
+                        if len(display) < self.n_show_top:
+                            dist = df_properties[property_name]
+                            if len(dist[~pd.isnull(dist)]) >= self.min_samples:
```
Nadav-Barak, line 168: maybe move this if into the first loop - if not enough samples raise relevant exception

deepchecks/nlp/checks/data_integrity/text_property_outliers.py
```diff
@@ -79,99 +79,118 @@ def run_logic(self, context: Context, dataset_kind: DatasetKind) -> CheckResult:
...
+                                lower_limit = info['lower_limit']
+                                upper_limit = info['upper_limit']
+
+                                fig = get_text_outliers_graph(
```
Nadav-Barak, line 172: maybe just wrap this part in try

tests/nlp/checks/data_integrity/text_property_outliers_test.py
```diff
@@ -84,3 +85,20 @@ def test_not_enough_samples(tweet_emotion_train_test_textdata):
...
+    text_data.calculate_builtin_properties(include_properties=['Sentences Count', 'Average Word Length'])
+    text_data.properties['Sentences Count'].iloc[9] = 'as'
+    text_data.properties['Average Word Length'].iloc[9] = 100
+
+    # Act
+    check = TextPropertyOutliers()
+    result = check.run(text_data)
+
+    # Assert
+    assert_that(result.value['Sentences Count'], equal_to('Numeric property contains non-numeric values.'))
```
Nadav-Barak, line 104: also assert on the Average Word Length value

tests/nlp/checks/data_integrity/text_property_outliers_test.py
```diff
@@ -84,3 +85,20 @@ def test_not_enough_samples(tweet_emotion_train_test_textdata):
 
     assert_that(calling(check.run).with_args(test),
                 raises(NotEnoughSamplesError, 'Need at least 6000 non-null samples to calculate outliers.'))
+    
+def test_non_numeric_values_in_text_properties_outliers(tweet_emotion_train_test_textdata):
```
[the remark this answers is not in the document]
harsh-deepchecks: I thought the string numeric values was a valid error. But after fixing that error no longer exist.

deepchecks/nlp/checks/data_integrity/text_property_outliers.py
```diff
@@ -73,85 +73,102 @@ def run_logic(self, context: Context, dataset_kind: DatasetKind) -> CheckResult:
         cat_properties = dataset.categorical_properties
         properties = df_properties.to_dict(orient='list')
 
-        if all(len(np.hstack(v).squeeze()) < self.min_samples for v in properties.values()):
-            raise NotEnoughSamplesError(f'Need at least {self.min_samples} non-null samples to calculate outliers.')
-
         # The values are in the same order as the batch order, so always keeps the same order in order to access
```
Nadav-Barak, line 76: Not related to you but remove this comment

Nadav-Barak (approved):
Looks great!

deepchecks/nlp/checks/data_integrity/text_property_outliers.py
```diff
@@ -73,85 +73,102 @@ def run_logic(self, context: Context, dataset_kind: DatasetKind) -> CheckResult:
...
+                result[name] = {
+                    'indices': [dataset.get_original_text_indexes()[i] for i in text_outliers],
+                    # For the upper and lower limits doesn't show values that are smaller/larger than the actual values
+                    # we have in the data
+                    'lower_limit': max(lower_limit, min(values_arr)),
+                    'upper_limit': min(upper_limit, max(values_arr)) if is_numeric else None,
+                    'outlier_ratio': len(text_outliers) / len(values_arr),
+                }
+            except Exception as exp:  # pylint: disable=broad-except
+                result[name] = f'{exp}'
```
Nadav-Barak, line 142: make sure exp is printed out nicely

harsh-deepchecks: Verified and exp will just be an error message and resolving this comment.

PR #2600, opened 2023-06-14
airbnb fixes and improvements

JKL98ISR:


deepchecks/tabular/datasets/regression/airbnb.py
```diff
@@ -113,76 +111,46 @@ def load_data(data_format: str = 'Dataset', as_train_test: bool = True, modify_t
         Represent the format of the returned value. Can be 'Dataset'|'Dataframe'
         'Dataset' will return the data as a Dataset object
         'Dataframe' will return the data as a pandas Dataframe object
-    as_train_test : bool , default: True
-        If True, the returned data is split into train and test exactly like the toy model
-        was trained. The first return value is the train data and the second is the test data.
-        In order to get this model, call the load_fitted_model() function.
-        Otherwise, returns a single object.
+   load_train : bool , default: True
```
Nadav-Barak, line 114: why? is it not consistant with the other dataset load function + not consistant with the quickstarts

Nadav-Barak (commented):
I dont think this what Bressler meant when he created the issue.
@noamzbr if it is, lets discuss.

deepchecks/tabular/datasets/regression/airbnb.py
JKL98ISR: i chatted with @noamzbr specifically on that and airbnb isn't used in the quickstart.
we wanted to not download uneeded times

noamzbr (commented):
@Nadav-Barak actually airbnb is used in the latter way in the mon quickstarts - first train data and preds are loaded, then production data and pred. It makes sense to use it like that for the quickstarts, though I agree it's not the format used in other datasets. Agree with @JKL98ISR that download times are critical here and so we want to avoid redundant downloads. 

@JKL98ISR - did you run this in the monitoring onboarding flow? How did it look like?

JKL98ISR:
looks good @noamzbr. tested all possibilities

noamzbr (approved):

deepchecks/tabular/datasets/regression/airbnb.py
```diff
@@ -113,76 +112,46 @@ def load_data(data_format: str = 'Dataset', as_train_test: bool = True, modify_t
...
```
noamzbr, line 115: @JKL98ISR missing one space here, that's the indentation issue

PR #2603, opened 2023-06-15
multivariate drift fix

JKL98ISR:


deepchecks/tabular/checks/train_test_validation/multivariate_drift.py
```diff
@@ -101,7 +101,9 @@ def run_logic(self, context: Context) -> CheckResult:
         cat_features = train_dataset.cat_features
         numerical_features = train_dataset.numerical_features
 
-        sample_size = min(self.n_samples, train_dataset.n_samples, test_dataset.n_samples)
+        sample_size = min(train_dataset.n_samples, test_dataset.n_samples)
+        if self.n_samples is not None:
+            sample_size = min(sample_size, self.n_samples)
```
Nadav-Barak, line 106:
```suggestion
        sample_size = min(train_dataset.n_samples, test_dataset.n_samples) self.n_samples is None else self.n_samples
```

Nadav-Barak: sorry for being annoying, you can ignore if you wish

Nadav-Barak (approved):

deepchecks/tabular/checks/train_test_validation/multivariate_drift.py
JKL98ISR: was on the edge if to do like that or not, compromised on another change, you can take a look

PR #2608, opened 2023-06-18
Deepchecks fix

nirhutnik:
Added deepchecks Fix!

### DIFF
--- a/deepchecks/core/fix_classes.py
+++ b/deepchecks/core/fix_classes.py
@@ -1,0 +1,152 @@
+# ----------------------------------------------------------------------------
+# Copyright (C) 2021-2023 Deepchecks (https://www.deepchecks.com)
+#
+# This file is part of Deepchecks.
+# Deepchecks is distributed under the terms of the GNU Affero General
+# Public License (version 3 or later).
+# You should have received a copy of the GNU Affero General Public License
+# along with Deepchecks.  If not, see <http://www.gnu.org/licenses/>.
+# ----------------------------------------------------------------------------
+#
+# pylint: disable=import-outside-toplevel
+"""Module containing the fix classes and methods."""
+import abc
+from typing import List, Optional, Tuple, Union
+
+import numpy as np
+import pandas as pd
+
+from deepchecks.core.checks import DatasetKind
+from deepchecks.tabular import deprecation_warnings  # pylint: disable=unused-import # noqa: F401
+from deepchecks.tabular._shared_docs import docstrings
+from deepchecks.tabular.context import Context
+from deepchecks.tabular.dataset import Dataset
+from deepchecks.utils.typing import BasicModel
+
+__all__ = [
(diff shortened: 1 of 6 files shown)


deepchecks/tabular/checks/data_integrity/data_duplicates.py
```diff
@@ -118,3 +122,30 @@ def run_logic(self, context: Context, dataset_kind):
             display = None
 
         return CheckResult(value=percent_duplicate, display=display)
+
+    @docstrings
+    @merge_args(SingleDatasetCheck.run)
```
Nadav-Barak, line 127: following our discussion in algo slack, Would make it possible to supply check result to the fix function and use it without supplying all of the run mandatory arguments. 
Also would add more text explaining that if no check result is not supplied the fix will rerun the check and if check result is supplied it does not rerun

deepchecks/tabular/checks/data_integrity/string_mismatch.py
```diff
@@ -178,6 +180,39 @@ def condition(result, max_ratio: float):
...
+    @docstrings
+    @merge_args(SingleDatasetCheck.run)
+    def fix(self, *args, check_result: CheckResult = None, **kwargs) -> FixResult:
+        """Run fix.
+
+        Parameters
+        ----------
+        {additional_context_params:2*indent}
+        check_result : CheckResult, default: None
+            CheckResult object to use for fixing the dataset.
```
Nadav-Barak, line 192: following the above comment - will extend the discussion here

deepchecks/core/fix_classes.py
```diff
@@ -0,0 +1,139 @@
...
+    'TrainTestCheckFixMixin'
+]
+
+
+class FixResult:
+    """Class to store the result of a fix."""
+
+    def __init__(
+            self,
+            fixed_train: Optional[Dataset] = None,
```
Nadav-Barak, line 35: if its a single dataset - will it be returned as fixed train?

Nadav-Barak (commented):
Something additional that came to mind, does all our fixes make it so that rerunning the check the fixed dataset will not find issues? not sure it mandetory

deepchecks/tabular/checks/data_integrity/data_duplicates.py
```diff
@@ -118,3 +122,30 @@ def run_logic(self, context: Context, dataset_kind):
...
+        keep : Literal['first', 'last', False], default: 'first'
+            Whether to keep the first or last duplicate row.
+            If False, all duplicates will be removed.
+
+        Returns
+        -------
+        Dataset
+            Dataset with fixed duplicates.
+        """
+        context = self.get_context(*args, **kwargs)
```
Nadav-Barak, line 144: if i created a single check object then ran it multiple times the context will be overrided no?

Maybe a safer option is to make this method static

PR #2611, opened 2023-06-24
Not raise DeepchecksProcessError for UnderAnnotated + WeakSegments

harsh-deepchecks:
#### Reference Issues/PRs

Fixes [DEE-621 - WeakSegments + Under annotated checks return error if no segments found](https://linear.app/deepchecks/issue/DEE-621/weaksegments-under-annotated-checks-return-error-if-no-segments-found)

The WeakSegments and UnderAnnotated checks earlier returns DeepchecksProcessError. But now, after the change, it will return a empty CheckResult in the following cases:
- If annotation ratio of the data is > 95%. (Only for UnderAnnotated check). This is a configurable parameter accepted by property and metadata checks.
- If weak segments is empty, it will return a message to try increasing the segment_by (either properties or metadata).

Also, both checks return a NotEnoughSamplesError if there are less than 10 samples. 

Below is the output for data integrity suite when Under annotated checks are skipped based on the above mentioned conditions: 
image

Running stand-alone under annotated check: 
image

### DIFF
--- a/deepchecks/nlp/checks/data_integrity/under_annotated_segments.py
+++ b/deepchecks/nlp/checks/data_integrity/under_annotated_segments.py
@@ -42,8 +42,7 @@ class UnderAnnotatedSegments(SingleDatasetCheck, WeakSegmentAbstract):
     def __init__(self, segment_by: str, columns: Union[Hashable, List[Hashable], None],
                  ignore_columns: Union[Hashable, List[Hashable], None], n_top_features: int,
                  segment_minimum_size_ratio: float, n_samples: int,  n_to_show: int,
-                 categorical_aggregation_threshold: float, annotation_ratio_threshold: float,
-                 **kwargs):
+                 categorical_aggregation_threshold: float, **kwargs):
         super().__init__(**kwargs)
         self.segment_by = segment_by
         self.columns = columns
@@ -53,7 +52,7 @@ class UnderAnnotatedSegments(SingleDatasetCheck, WeakSegmentAbstract):
         self.n_samples = n_samples
         self.n_to_show = n_to_show
         self.categorical_aggregation_threshold = categorical_aggregation_threshold
-        self.annotation_ratio_threshold = annotation_ratio_threshold
+        self.annotation_ratio_threshold = ANNOTATION_RATIO_THRESHOLD
 
     def run_logic(self, context: Context, dataset_kind) -> CheckResult:
         """Run check."""
@@ -249,8 +248,6 @@ class UnderAnnotatedPropertySegments(UnderAnnotatedSegments):
         Maximum number of samples to use for this check.
     n_to_show : int , default: 3
         number of segments with the weakest performance to show.
-    annotation_ratio_threshold : float , default: 95.0
-        The threshold ranging from 1 to 100 for the under annotated property segement check to run.
     categorical_aggregation_threshold : float , default: 0.05
         In each categorical column, categories with frequency below threshold will be merged into "Other" category.
     """
@@ -263,7 +260,6 @@ class UnderAnnotatedPropertySegments(UnderAnnotatedSegments):
                  n_samples: int = 10_000,
                  categorical_aggregation_threshold: float = 0.05,
                  n_to_show: int = 3,
-                 annotation_ratio_threshold: float = ANNOTATION_RATIO_THRESHOLD,
                  **kwargs):
         super().__init__(segment_by='properties',
                          columns=properties,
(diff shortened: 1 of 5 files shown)


deepchecks/nlp/checks/model_evaluation/weak_segments_performance.py
```diff
@@ -105,8 +110,9 @@ def run_logic(self, context: Context, dataset_kind) -> CheckResult:
                                                    dummy_model=dummy_model, scorer=scorer)
 
         if len(weak_segments) == 0:
-            raise DeepchecksProcessError('WeakSegmentsPerformance was unable to train an error model to find weak '
-                                         f'segments. Try increasing n_samples or supply more {self.segment_by}.')
+            display_msg = 'WeakSegmentsPerformance was unable to train an error model to find weak segments.'\
+                          f'Try supplying more {self.segment_by}.'
```
Nadav-Barak, line 114:
```suggestion
                          f'Try supplying additional {self.segment_by}.'
```

deepchecks/nlp/checks/model_evaluation/weak_segments_performance.py
```diff
@@ -105,8 +110,9 @@ def run_logic(self, context: Context, dataset_kind) -> CheckResult:
...
+            return CheckResult(value={'message': display_msg}, display=[display_msg])
```
Nadav-Barak, line 115: can have the value directly a str I think

deepchecks/nlp/checks/data_integrity/under_annotated_segments.py
```diff
@@ -30,15 +30,20 @@
...
 
 class UnderAnnotatedSegments(SingleDatasetCheck, WeakSegmentAbstract):
     """Check for under annotated data segments."""
 
     def __init__(self, segment_by: str, columns: Union[Hashable, List[Hashable], None],
                  ignore_columns: Union[Hashable, List[Hashable], None], n_top_features: int,
-                 segment_minimum_size_ratio: float, n_samples: int,
-                 categorical_aggregation_threshold: float, n_to_show: int, **kwargs):
+                 segment_minimum_size_ratio: float, n_samples: int,  n_to_show: int,
+                 categorical_aggregation_threshold: float, annotation_ratio_threshold: float,
```
Nadav-Barak, line 45: annotation_ratio_threshold should be a float

deepchecks/nlp/checks/data_integrity/under_annotated_segments.py
```diff
@@ -232,6 +249,8 @@ class UnderAnnotatedPropertySegments(UnderAnnotatedSegments):
         Maximum number of samples to use for this check.
     n_to_show : int , default: 3
         number of segments with the weakest performance to show.
+    annotation_ratio_threshold : float , default: 95.0
```
Nadav-Barak, line 252: I think we should not make it a argument but a const in the class

deepchecks/nlp/checks/model_evaluation/weak_segments_performance.py
```diff
@@ -55,6 +57,9 @@ def run_logic(self, context: Context, dataset_kind) -> CheckResult:
         text_data = context.get_data_by_kind(dataset_kind)
         text_data = text_data.sample(self.n_samples, random_state=context.random_state, drop_na_label=True)
 
+        if text_data.n_samples < MIN_TEXT_SAMPLES:
+            raise NotEnoughSamplesError(f'Not enough samples to find weak {self.segment_by} segments.'
+                                        ' Minimum 10 samples required.')
```
Nadav-Barak, line 62:
```suggestion
                                        f' Minimum {MIN_TEXT_SAMPLES} samples required.')
```

Nadav-Barak (approved):
Approved with small comments

deepchecks/nlp/checks/model_evaluation/weak_segments_performance.py
harsh-deepchecks: Actually the conditions expect the value to be a Dict.

deepchecks/nlp/checks/data_integrity/under_annotated_segments.py
harsh-deepchecks: Done

deepchecks/nlp/checks/data_integrity/under_annotated_segments.py
harsh-deepchecks: Made it a const instead of parameter

PR #2621, opened 2023-07-04
Calculation of NLP properties with batches

ItayGabbay:
Before/after time comparison for 3000 samples:

| Batch size | Time(s) |
|--------|--------|
| 1 (no batches) | 124 |
| 16 | **12** |
| 32 | **5** | 

In addition, I added handling of exceptions in case the batch_size is too large - we're gradually reducing the batch size until there is sufficient memory.

### DIFF
--- a/deepchecks/nlp/utils/text_properties.py
+++ b/deepchecks/nlp/utils/text_properties.py
@@ -367,33 +367,36 @@ def _predict(text_batch: Sequence[str], classifier, kind: str, batch_size: int) 
         # TODO: make this way smarter, and not just a hack. Count tokens, for a start. Then not just sample sentences.
         # If text is longer than classifier context window, sample it:
         text_list_to_predict = []
-        reduced_batch_size = batch_size  # Initialize the reduced batch size
 
-        while reduced_batch_size >= 1:
+        for text in text_batch:
+            if len(text) > MAX_CHARS:
+                sentences = _sample_for_property(text, mode='sentences', limit=10, return_as_list=True)
+                text_to_use = ''
+                for sentence in sentences:
+                    if len(text_to_use) + len(sentence) > MAX_CHARS:
+                        break
+                    text_to_use += sentence + '. '
+
+                # if even one sentence is too long, use part of the first one:
+                if len(text_to_use) == 0:
+                    text_to_use = cut_string(sentences[0], MAX_CHARS)
+                text_list_to_predict.append(text_to_use)
+            else:
+                text_list_to_predict.append(text)
+
+        results = []
+        remaining_texts = text_list_to_predict[:]
+        reduced_batch_size = batch_size
+
+        while remaining_texts:
             try:
-                for text in text_batch:
-                    if len(text) > MAX_CHARS:
-                        sentences = _sample_for_property(text, mode='sentences', limit=10, return_as_list=True)
-                        text_to_use = ''
-                        for sentence in sentences:
-                            if len(text_to_use) + len(sentence) > MAX_CHARS:
-                                break
-                            text_to_use += sentence + '. '
-
-                        # if even one sentence is too long, use part of the first one:
-                        if len(text_to_use) == 0:
-                            text_to_use = cut_string(sentences[0], MAX_CHARS)
-                        text_list_to_predict.append(text_to_use)
-                    else:
-                        text_list_to_predict.append(text)
-
-                v_list = classifier(text_list_to_predict, batch_size=reduced_batch_size)
-                results = []
+                text_list_to_predict_batch = remaining_texts[:reduced_batch_size]
+                v_list = classifier(text_list_to_predict_batch, batch_size=reduced_batch_size)
 
                 for v in v_list:
                     if not v:
                         results.append(np.nan)
-                    if kind == 'toxicity':
+                    elif kind == 'toxicity':
                         results.append(v['score'])
                     elif kind == 'fluency':
                         results.append(v['score'] if v['label'] == 'LABEL_1' else 1 - v['score'])
@@ -402,11 +405,14 @@ def _predict(text_batch: Sequence[str], classifier, kind: str, batch_size: int) 
                     else:
                         raise ValueError('Unsupported value for "kind" parameter')
 
+                remaining_texts = remaining_texts[reduced_batch_size:]
+                reduced_batch_size = min(reduced_batch_size, len(remaining_texts))
+
                 return results  # Return the results if prediction is successful
 
             except Exception:  # pylint: disable=broad-except
-                reduced_batch_size = max(reduced_batch_size // 2, 1)  # Reduce the batch size by half
-                text_list_to_predict = []  # Clear the list of texts to predict for retry
+                reduced_batch_size = max(reduced_batch_size // 2, 1)
+                results.extend([np.nan] * reduced_batch_size)
 
         return [np.nan] * batch_size  # Prediction failed, return NaN values for the original batch size
 
@@ -634,7 +640,7 @@ def average_syllable_length(text: str, cmudict_dict: dict = None) -> float:
 
 
 def _batch_wrapper(text_batch: Sequence[str], func: Callable, **kwargs) -> List[Any]:
-    """A utility function to wrap the non-batched properties with batches."""
+    """Wrap the non-batched properties execution with batches API."""
     results = []
     for text in text_batch:
         results.append(run_available_kwargs(func, text=text, **kwargs))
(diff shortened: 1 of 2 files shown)


deepchecks/nlp/utils/text_properties.py
```diff
@@ -870,44 +900,47 @@ def calculate_builtin_properties(
...
-        unit='Text Sample'
-    )
-    for text in progress_bar:
-        progress_bar.set_postfix(
-            {'Sample': truncate_string(text, max_length=20) if text else 'EMPTY STRING'},
-            refresh=False
-        )
-        if pd.isna(text):
+    with tqdm(total=int(len(raw_text)//batch_size)) as pbar:
+        for i in range(0, len(raw_text), batch_size):
```
Nadav-Barak, line 906: you can do tqdm(range(0, len(raw_text), batch_size)) than you dont need to manually update pbar

deepchecks/nlp/utils/text_properties.py
```diff
@@ -870,44 +900,47 @@ def calculate_builtin_properties(
...
+            batch = raw_text[i:i + batch_size]
+            batch_properties = defaultdict(list)
+            pbar.update(1)
+
+            # filtering out empty sequences
+            filtered_sequences = [seq for seq in batch if pd.isna(seq) is False]
```
Nadav-Barak, line 912: what do you do with them? return none in the relevant places?

yromanyshyn (approved):

deepchecks/nlp/utils/text_properties.py
ItayGabbay: np.nan. as it was before

deepchecks/nlp/utils/text_properties.py
```diff
@@ -374,28 +380,53 @@ def _predict(text: str, classifier, kind: str) -> float:
...
+                elif kind == 'fluency':
+                    results.append(v['score'] if v['label'] == 'LABEL_1' else 1 - v['score'])
+                elif kind == 'formality':
+                    results.append(v['score'] if v['label'] == 'formal' else 1 - v['score'])
+                else:
+                    raise ValueError('Unsupported value for "kind" parameter')
+
+            return results  # Return the results if prediction is successful
+
+        except Exception:  # pylint: disable=broad-except
```
Nadav-Barak, line 425: seems that for a batch of size 32 that has one bad example you will always come the at the end to running everything sample by sample but after you run many unnecessary calculations

deepchecks/nlp/utils/text_properties.py
```diff
@@ -870,40 +917,41 @@ def calculate_builtin_properties(
...
-            for prop in text_properties:
-                calculated_properties[prop['name']].append(np.nan)
-            continue
-        sample_language = run_available_kwargs(language, text=text, **kwargs)
+    for i in tqdm(range(0, len(raw_text), batch_size)):
+        batch = raw_text[i:i + batch_size]
+        batch_properties = defaultdict(list)
+
+        # filtering out empty sequences
+        filtered_sequences = [seq for seq in batch if pd.isna(seq) is False]
```
Nadav-Barak, line 925: (if i understood currently) You are filtering out stuff but dont return none in the relevant places. if i send you 100 samples and you give me back a dataframe with 93 values, how can i work with this?

Nadav-Barak (changes requested):
This is a big functionality, we should have several tests for it

deepchecks/nlp/utils/text_properties.py
```diff
@@ -870,40 +917,41 @@ def calculate_builtin_properties(
...
+                    if prop['name'] in BATCH_PROPERTIES:
+                        value = run_available_kwargs(func=prop['method'], text_batch=filtered_sequences, **kwargs)
+                    else:
+                        value = _batch_wrapper(text_batch=filtered_sequences, func=prop['method'], **kwargs)
+                    batch_properties[prop['name']].extend(value)
                 except ImportError as e:
                     warnings.warn(warning_message.format(prop['name'], str(e)))
-                    calculated_properties[prop['name']].append(np.nan)
+                    batch_properties[prop['name']].extend([np.nan] * len(batch))
                     import_warnings.add(prop['name'])
```
nirhutnik, line 950: Not sure I followed here. I don't see how you keep the order here. what if I have 32 samples, and sample number 7 is not in English? Where do you make sure that number 7 gets the "none" value? I only see that you filter by language but don't fill the others with None, and make sure to keep the order.

ItayGabbay: Fixed.

deepchecks/nlp/utils/text_properties.py
ItayGabbay: Fixed.

deepchecks/nlp/utils/text_properties.py
ItayGabbay: Added a retry mechanism to avoid that

deepchecks/nlp/utils/text_properties.py
```diff
@@ -374,28 +381,54 @@ def _predict(text: str, classifier, kind: str) -> float:
...
+                if not v:
+                    results.append(np.nan)
+                elif kind == 'toxicity':
+                    results.append(v['score'])
+                elif kind == 'fluency':
+                    results.append(v['score'] if v['label'] == 'LABEL_1' else 1 - v['score'])
+                elif kind == 'formality':
+                    results.append(v['score'] if v['label'] == 'formal' else 1 - v['score'])
+                else:
+                    raise ValueError('Unsupported value for "kind" parameter')
```
Nadav-Barak, line 422: Duplicate code can copy to function

tests/nlp/utils/test_properties.py
```diff
@@ -182,6 +182,25 @@ def test_calculate_average_sentence_length_property(tweet_emotion_train_test_tex
...
+
+    # Act
+    for batch in batch_sizes:
+        result = calculate_builtin_properties(test_text, include_properties=['Average Words Per Sentence'],
+                                              batch_size=batch)[0]
+        result_none_text = calculate_builtin_properties([None], include_properties=['Average Words Per Sentence'],
+                                                        batch_size=batch)[0]
+
+        # Assert
+        assert_that(result['Average Words Per Sentence'][0: 10], equal_to([5.667, 7.0, 11.0, 12.0, 8.0, 19.0, 3.0, 9.0,
```
Nadav-Barak, line 199: add assertion also on property that is english only to test if works well also there

Nadav-Barak (approved):

deepchecks/nlp/utils/text_properties.py
```diff
@@ -374,28 +381,54 @@ def _predict(text: str, classifier, kind: str) -> float:
...
+                            raise ValueError('Unsupported value for "kind" parameter')
+                    except Exception:  # pylint: disable=broad-except
+                        results.append(np.nan)
+                return results  # Return the results if prediction is successful
+
+            v_list = classifier(text_list_to_predict, batch_size=reduced_batch_size)
+            results = []
+
+            for v in v_list:
+                if not v:
```
noamzbr, line 413: seems like there's easily avoidable code duplication. Maybe take this whole switch clause to an external function?

deepchecks/nlp/utils/text_properties.py
```diff
@@ -614,6 +650,19 @@ def average_syllable_length(text: str, cmudict_dict: dict = None) -> float:
...
 
 
+def _batch_wrapper(text_batch: Sequence[str], func: Callable, **kwargs) -> List[Any]:
+    """Wrap the non-batched properties execution with batches API."""
+    results = []
+    language_property_result = []
+    if 'language_property_result' in kwargs:
+        language_property_result = kwargs.pop('language_property_result')
+    for i, text in enumerate(text_batch):
+        kwargs['language_property_result'] = language_property_result[i] if i < len(language_property_result) else None
```
noamzbr, line 660: didn't understand this line

deepchecks/nlp/utils/text_properties.py
```diff
@@ -870,40 +925,54 @@ def calculate_builtin_properties(
...
+                        and ignore_non_english_samples_for_english_properties is True:
+                    filtered_sequences = [e for i, e in enumerate(filtered_sequences) if i not in non_english_indices]
+                kwargs['batch_size'] = batch_size
                 try:
-                    value = run_available_kwargs(prop['method'], text=text, **kwargs)
-                    calculated_properties[prop['name']].append(value)
+                    if prop['name'] in BATCH_PROPERTIES:
+                        value = run_available_kwargs(func=prop['method'], text_batch=filtered_sequences, **kwargs)
+                    else:
+                        value = _batch_wrapper(text_batch=filtered_sequences, func=prop['method'], **kwargs)
```
noamzbr, line 958: I kinda hate this, can we perhaps add _batch_wrapper inside the original property dict? It's ugly duplication but I don't like that we have these gazillion lists of property types...

ItayGabbay: I think it makes sense to have batched & non-batched properties

PR #2634, opened 2023-07-17
Add examples to new and existing properties

noamzbr:


docs/source/nlp/usage_guides/nlp_properties_extended.rst
```diff
@@ -86,3 +130,14 @@ The property is calculated by identifying key entities and quantities in the LLM
...
+
+Examples
+~~~~~~~~
+
+======================================================================================================================  =====================================================  ===================
+LLM Input                                                                                                               LLM Output                                             Grounded in Context
+======================================================================================================================  =====================================================  ===================
+Michael Jordan (1963) is an American former professional basketball player and businessman. In what year was he born?   He was born in 1963.                                   1.0
+Michael Jordan (1963) is an American former professional basketball player and businessman. In what year was he born?   Michael Jeffrey Jordan was born in 1963                0.5
+Michael Jordan (1963) is an American former professional basketball player and businessman. In what year was he born?   He won many NBA championships with the Chicago Bulls   0.0
```
shir22, line 142: This example is misleading IMO... shouldn't we demonstrate with a different question everytime?
e.g. "In which group did he play?"
Cause the from the example it seems that it's not answering the question, more than it's just using other information not from the context

docs/source/nlp/usage_guides/nlp_properties_extended.rst
```diff
@@ -56,6 +78,17 @@ The model uses the roberta-base architecture, and was trained on
 `Rao and Tetreault, 2018 <https://aclanthology.org/N18-1012>`__ and online formality corpus from
 `Pavlick and Tetreault, 2016 <https://aclanthology.org/Q16-1005>`__.
 
+Examples
+~~~~~~~~
+
+================================================================  ========
```
shir22, line 84: I assume the underline and line above should be longer than / equal to the title length. For all examples

shir22 (approved):

docs/source/nlp/usage_guides/nlp_properties_extended.rst
noamzbr: Not needed for correct rendering of the output

docs/source/nlp/usage_guides/nlp_properties_extended.rst
noamzbr: right, changed the last question

PR #2635, opened 2023-07-18
Differentiate special chars from punctuations

noamzbr:
Special chars are non alphanumeric unicode chars that are not whitespaces or (most) punctuations.

Added a % Punctuation instead


JKL98ISR (approved):

deepchecks/nlp/utils/text_properties.py
```diff
@@ -265,6 +269,11 @@ def average_word_length(text: str) -> float:
 
 def percentage_special_characters(text: str) -> float:
     """Return percentage of special characters (as float between 0 and 1)."""
+    return len([c for c in text if c in NON_PUNCTUATION_SPECIAL_CHARS]) / len(text) if len(text) != 0 else 0
```
nirhutnik, line 272: What about what we said about non-letter/digits?

noamzbr: Can you remind me? What did we say?

shir22 (approved):

PR #2641, opened 2023-07-23
Add ability to cache text prop models

shayts7:
#### Reference Issues/PRs
https://linear.app/deepchecks/issue/LLM-214/make-load-openai-data-log-interaction-async

model loading (even after saved to disk in the first time) takes few seconds, in the current implementation of the LLM server - that blocks outer calls.
This change enables caching for those models (be default False), after cache - operation takes ~0.2sec

#### Any other comments?
memory footprint was tests - the cached models takes ~1.7G 
Today we have 4 workers in production - means we are participating to capture ~7.2G (out of 32G of RAM)

In the future we will extract props calculation outside of the application...

---


deepchecks/nlp/utils/text_properties_models.py
```diff
@@ -0,0 +1,221 @@
...
+        device: Optional[str] = None,
+        models_storage: Union[pathlib.Path, str, None] = None,
+        use_cache=False
+):
+    if use_cache:
+        return _get_transformer_pipeline(property_name, model_name, device, models_storage)
+    return _get_transformer_pipeline.__wrapped__(property_name, model_name, device, models_storage)
+
+
+@lru_cache(maxsize=5)
```
Nadav-Barak, line 148: why 5?

deepchecks/nlp/utils/text_properties_models.py
```diff
@@ -0,0 +1,221 @@
...
+        )
+        raise ImportError(error_template.format(
+            property_name=property_name,
+            package_name=package_name
+        )) from error
+    else:
+        return lib
+
+
+def get_transformer_pipeline(
```
Nadav-Barak, line 136: would add docstring

deepchecks/nlp/utils/text_properties.py
```diff
@@ -875,6 +696,8 @@ def calculate_builtin_properties(
         Also, if a folder already contains relevant resources they are not re-downloaded.
     batch_size : int, default 8
         The batch size.
+    cache_models : bool, default False
+        cache the models being used in this function, to save load time in next execution
```
Nadav-Barak, line 700: would add more details, what are the memory requirements

deepchecks/nlp/utils/text_properties_models.py
```diff
@@ -0,0 +1,221 @@
...
+from functools import lru_cache
+from typing import Optional, Union
+
+import requests
+from nltk import corpus
+
+MODELS_STORAGE = pathlib.Path(__file__).absolute().parent / '.nlp-models'
+
+
+def get_create_model_storage(models_storage: Union[pathlib.Path, str, None] = None):
```
Nadav-Barak, line 12: what happens here if you use cache=true?

deepchecks/nlp/utils/text_properties_models.py
```diff
@@ -0,0 +1,221 @@
...
+
+def get_transformer_pipeline(
+        property_name: str,
+        model_name: str,
+        device: Optional[str] = None,
+        models_storage: Union[pathlib.Path, str, None] = None,
+        use_cache=False
+):
+    if use_cache:
+        return _get_transformer_pipeline(property_name, model_name, device, models_storage)
```
yromanyshyn, line 144: could you say, how much memory each model occupies?

deepchecks/nlp/utils/text_properties_models.py
shayts7: I case that two more will come in the future...

deepchecks/nlp/utils/text_properties.py
shayts7: The memory is vary between model and model...

deepchecks/nlp/utils/text_properties_models.py
shayts7: In that case `get_create_model_storage()` will be called once and store the model to disk, from that time one - the model will be fetched from the cache, but after restart - the `model storage` benefits will come to hand as we won't need to download the model for the net, but only load it from disk

deepchecks/nlp/utils/text_properties_models.py
shayts7: In total the models captures ~1.7G of RAM, you can see a break down in this file:
[memray-flamegraph-tests.py.50912.html.gz](https://github.com/deepchecks/deepchecks/files/12143377/memray-flamegraph-tests.py.50912.html.gz)

deepchecks/nlp/utils/text_properties_models.py
Nadav-Barak: not critical but would describe this flow in some docstring

deepchecks/nlp/utils/text_properties_models.py
shayts7: added a comment that explains `__wrapped__`

deepchecks/nlp/utils/text_properties_models.py
shayts7: I prefer not to write it in this method, as this can be changed in the future and then we will left with unwanted comment :-)

yanivshani100 (approved):

PR #2670, opened 2023-09-12
cache models only on cpu + quantization

Nadav-Barak:
1. Changed quantization to huggingface default instead of onnx
2. Cached models are always stored on cpu and moved to selected device before inference

image


Nadav-Barak:
Resolves https://linear.app/deepchecks/issue/LLM-398/quantize-our-models

noamzbr (approved):

PR #2671, opened 2023-09-18
cache models on cpu + hugging face based quantization

Nadav-Barak:
This pr contain two main contributions:

1. Models can be cached on cpu and then moved to relevant device before inference for nlp properties calculation - this to skip the time required to load the models from disc
2. Deepchecks now support huggingface based quantization for all relevant models

This pr relevant test were tested on both a GPU and a CPU only machines. Resolves https://linear.app/deepchecks/issue/LLM-533/release-memory-from-gpu-when-using-deepchecks-models

### DIFF
--- a/deepchecks/nlp/utils/text_properties.py
+++ b/deepchecks/nlp/utils/text_properties.py
@@ -613,16 +613,18 @@ def _select_properties(
     ignore_properties = [prop.lower() for prop in ignore_properties] if ignore_properties else None
 
     if include_properties is not None:
-        properties = [prop for prop in ALL_PROPERTIES if prop['name'].lower() in include_properties]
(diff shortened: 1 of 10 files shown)


deepchecks/nlp/utils/text_properties.py
```diff
@@ -847,6 +850,15 @@ def calculate_builtin_properties(
         if k in calculated_properties
     }
 
+    if cache_models:
+        # Move the transformers models to CPU RAM memory
+        for model_name in ['toxicity_classifier', 'formality_classifier', 'fluency_classifier']:
```
noamzbr, line 855: Maybe this should be in the same place as the other property constant lists.

noamzbr (approved):

PR #2681, opened 2023-10-16
onnx models

Nadav-Barak:
Ability to use onnx gpu optimized models for property calculations


deepchecks/nlp/utils/text_properties.py
```diff
@@ -758,20 +749,23 @@ def calculate_builtin_properties(
                 calculated_properties[prop] = [np.nan] * len(raw_text)
         kwargs['cmudict_dict'] = get_cmudict_dict(use_cache=cache_models)
 
-    if 'Toxicity' in properties_types:
+    if 'Toxicity' in properties_types and 'toxicity_classifier' not in kwargs:
```
noamzbr, line 752: Can you help me understand that?

noamzbr (commented):
There is a big danger that someone will pass device='gpu' in some usage, e.g. how do we call calculate_builtin_properties? Would consider forcing device=None for onnx models

noamzbr (approved):

PR #2695, opened 2023-11-05
Added support for filterting check results by name

harsh-deepchecks:
- Added implementation for fetching the check results by header name of the check. I have added this functionality in the same function and added checks in order to accept either indexes or names (one of them required always).
- Will be working on the other part of the task to figure out a way to know the row in the DF which causes any checks to fail so that user can know which data point is problematic.

### DIFF
--- a/deepchecks/core/suite.py
+++ b/deepchecks/core/suite.py
@@ -130,12 +130,12 @@ class SuiteResult(DisplayableResult):
             raise DeepchecksNotSupportedError('Either idx or names should be passed')
         if idx and names:
             raise DeepchecksNotSupportedError('Only one of idx or names should be passed')
-        output = []
-        for index, result in enumerate(self.results):
-            if idx and index in idx:
-                output.append(result)
-            elif names and result.get_header() in names:
-                output.append(result)
+
+        if names:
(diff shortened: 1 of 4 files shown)


deepchecks/core/suite.py
```diff
@@ -106,14 +106,20 @@ def __init__(
             else:
                 raise TypeError(f'Unknown type of result - {type(result).__name__}')
 
-    def select_results(self, idx: Set[int]) -> List[Union[
+    def select_results(self, idx: Set[int] = None, names: Set[str] = None) -> List[Union[
         'check_types.CheckResult',
         'check_types.CheckFailure'
     ]]:
-        """Select results by indexes."""
+        """Select results either by indexes or result header names."""
```
Nadav-Barak, line 113: Add docstring for the arguments

Nadav-Barak (commented):
Add 2-3 tests for that functionality with different combinations between names and indexes

deepchecks/core/suite.py
harsh-deepchecks: Done

Nadav-Barak (approved):

PR #2705, opened 2023-11-22
weak segment should have maximum of 1 category

Nadav-Barak:


deepchecks/utils/abstracts/weak_segment_abstract.py
```diff
@@ -213,24 +213,48 @@ def _weak_segments_search(self, data: pd.DataFrame, score_per_sample: pd.Series,
...
+        # Filter and adapt the weak segments results
+        result = pd.DataFrame(columns=weak_segments.columns)
+        used_features = set()
+        for _, row in weak_segments.sort_values(score_title).iterrows():
+            new_row = row.copy()
+            if not multiple_segments_per_feature and \
+                    (row['Feature1'] in used_features or row['Feature2'] in used_features):
+                continue
+
+            # Make sure segments based on categorical features are based only on a single category
```
noamzbr, line 225: I think this should not be the default behavior for deepchecks

deepchecks/utils/abstracts/weak_segment_abstract.py
```diff
@@ -213,24 +213,48 @@ def _weak_segments_search(self, data: pd.DataFrame, score_per_sample: pd.Series,
...
+            if row['Feature1'] in self.encoder_mapping:
+                unique_values_in_range = [x for x in self.encoder_mapping[row['Feature1']]['encoded_value'].values if
```
noamzbr, line 227: issue with this logic is that it can leave us with weak segments that are not actually weak - you are checking that they are big enough (which is good), but there is no guarantee that the segment has a score than is lower than the rest of the data by more than a small margin.

noamzbr (approved):
Please add `max_categories_in_weak_segment` and consider if the default should be 1 or None

